PR #10, opened 2022-05-04
feat(): new endpoint to delete the cache

midgerate:


src/api.ts
```diff
@@ -1,11 +1,37 @@
...
-import { set, get, streamToBuffer } from './aws';
+import { set, get, streamToBuffer, remove } from './aws';
 import resolvers from './resolvers';
 import constants from './constants.json';
 
 const router = express.Router();
 
+router.post('/:type/:id/delete', async (req, res) => {
+  const { type, id } = req.params;
+  const { sizes } = req.body;
```
bonustrack, line 11: The issue with this is that we wont be able to really clear an image from the cache, we would need to populate the different size from the UI, but some might be unknown. Instead of this I recommend make it in a way that when we store image under a folder with a name that use the key of the base image, and when we delete an image cache we just delete everything on the folder with that key. This route can just be a GET request (make it easy to clear cache manually).

midgerate: Okay good idea. Will do the change like this.

src/aws.ts
```diff
@@ -29,11 +29,40 @@ export async function set(key, value) {
...
-export async function get(key) {
+export async function remove(folder) {
+  try {
+    const Bucket = process.env.AWS_BUCKET_NAME;
+    const listedObjects = await client.listObjectsV2({
+      Bucket,
+      Prefix: `public/stamp/${folder}`
+    });
+
+    if (listedObjects.Contents.length === 0) return;
```
bonustrack, line 40: Server crash when trying this on empty folder
image

src/api.ts
```diff
@@ -1,68 +1,80 @@
 import express from 'express';
 import { parseQuery, resize, sha256 } from './utils';
-import { set, get, streamToBuffer } from './aws';
+import { set, get, streamToBuffer, remove } from './aws';
 import resolvers from './resolvers';
 import constants from './constants.json';
 
 const router = express.Router();
 
+router.delete('/:type/:id', async (req, res) => {
```
bonustrack, line 9: This should be a get route not delete

src/aws.ts
```diff
@@ -15,11 +15,11 @@ export async function streamToBuffer(stream: Readable) {
   });
 }
 
-export async function set(key, value) {
+export async function set(key, value, folder) {
```
bonustrack, line 18: We should not change the arguments in this function and get function

bonustrack (commented):
I'll take over this PR

src/api.ts
```diff
@@ -1,68 +1,80 @@
...
+  const { type, id } = req.params;
+
+  try {
+    const { address, network, w, h } = await parseQuery(id, { s: constants.max });
+    const baseImageFolder = sha256(JSON.stringify({ type, network, address, w, h }));
+    remove(baseImageFolder);
```
bonustrack, line 15: I think it doesn't fail for you because you don't await here

bonustrack:
Should be good now!

Issue #6, opened 2022-03-16
Load ENS avatar directly from onchain data


bonustrack:
Currently ENS avatar are loaded from ENS metadata server, it would resolve faster if we implement our own resolver that get avatar bases on onchain data

ryanli-me:
what's the current perf of ENS metadata server vs our own resolver?

bonustrack:
It will improve speed because we will load avatar directly without going through ENS metadata server, I'm not sure how much it will get improved I expect 30% faster, but even if its a little bit there is lot of optimization we could do later with that, we could detect when someone have updated his image or not have any image set. Here is some example code to resolve ENS image directly:
https://github.com/ensdomains/ens-metadata-service/blob/master/src/avatar.ts
https://github.com/metaphor-xyz/davatar-helpers/blob/master/packages/react/src/Image.tsx

bonustrack:
@ryanli-me

FlacoJones:
Relevant docs on fetching avatar uri from on-chain source: https://gist.github.com/Arachnid/9db60bd75277969ee1689c8742b75182

ryanli-me:
Sorry this got lost. here's an on-chain resolution logic I did adapted from ether.js.

https://github.com/cyberconnecthq/cyberprofile/blob/main/utils/resolver.ts

bonustrack:
Fixed in #12

PR #12, opened 2022-05-09
Fetches ENS Avatar URL from on-chain using ethers.js ENS methods

FlacoJones:
**DESCRIPTION**

Basically this gets ens name -> gets resolver for that name -> gets avatarUrl from resolver all using ethers.js rather than the previous ABI methods.

**DISCUSSION**

Thoughts on adopting the [built-in ENS methods from ethers.js](https://docs.ethers.io/v5/api/providers/provider/#Provider--ens-methods) to fetch avatar url's from on-chain?

There are some reverse resolution *gotchas* that using a library will save us from needing to account for, e.g. [this](https://docs.ens.domains/dapp-developer-guide/resolving-names#reverse-resolution).


FlacoJones:
Can test with vitalik.eth

`http://localhost:3000/avatar/0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045`

FlacoJones:
Closes #6

src/resolvers/ens.ts
```diff
@@ -1,23 +1,20 @@
...
-export default async function resolve(address, network) {
+export default async function resolve(address) {
   try {
-    const name = await resolveName(network, address);
-    const url = `https://metadata.ens.domains/mainnet/avatar/${name}`;
-    const input = (await axios({ url, responseType: 'arraybuffer' })).data as Buffer;
-    return await resize(input, max, max);
+    const provider = new StaticJsonRpcProvider('https://rpc.ankr.com/eth');
+    const name = await provider.lookupAddress(address);
+    if (name) {
```
bonustrack, line 11: Seem like if there isn't any name this function wont get resolved, I would just do like this here:
```suggestion
    if (!name) return false;
```

bonustrack:
Thanks for your PR! Added some suggestions, didn't knew about the Ethers.js built-in solution.

src/resolvers/ens.ts
```diff
@@ -1,23 +1,20 @@
...
+      const resolver = await provider.getResolver(name);
```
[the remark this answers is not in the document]
FlacoJones: Answer here: https://docs.ethers.io/v5/api/providers/provider/#Provider-lookupAddress

```
Performs a reverse lookup of the address in ENS using the Reverse Registrar. If the name does not exist, or the forward lookup does not match, null is returned.

An ENS name requries additional configuration to setup a reverse record, they are not automatically set up.
```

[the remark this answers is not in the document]
FlacoJones: I'm guessing that ethers.js implements the spec in the gist above ^ under the hood

FlacoJones:
After talking with @mktcode i may switch this to use the @ensdomains/ens-avatar package.

It even natively supports in-memory caching and prevents XSS attacks in SVGs.

bonustrack:
@FlacoJones All images are resized and turned to WebP, I don't think injection would work. Most important here would be speed, would be great if we can compare @ensdomains/ens-avatar vs Ethers.js

FlacoJones:
Not great - about 4 seconds slower than original. I guess 3 calls via a provider isn't great either compared to a nicely cached centralized metadata server.

bonustrack:
Interesting thats much slower than previously. I wonder how much of the time was due to resolve the name vs resolve the image. The code for metadata server is here: https://github.com/ensdomains/ens-metadata-service maybe the way its implemented make it much faster?

bonustrack:
And resolving the name also could be benchmark with getNames vs lookupAddress

FlacoJones:
Under the hood the `ens-metadata-service` uses `@ensdomains/ens-avatar` for fetching the metadata.

As we see here it uses the same `ENS Name -> resolver -> text record containing NFT uri` [here](https://github.com/ensdomains/ens-avatar/blob/main/src/index.ts#L58).

It could be something to do with caching on their hosted metadata service. As seen here, it happened that every once in a while there would be a 5 second call in the original

```
Call to ens.resolve (ORIGINAL) took 1333.342137992382 milliseconds.
Call to ens.resolve (ORIGINAL) took 1248.363413989544 milliseconds.

Call to ens.resolve (ORIGINAL) took 5220.557393014431 milliseconds.

Call to ens.resolve (ORIGINAL) took 1269.0398569703102 milliseconds.
Call to ens.resolve (ORIGINAL) took 1272.6246370077133 milliseconds.
```

FlacoJones:
These benchmarks are kind of meaningless actually since the two for the new method have no cache, while the ENS metadata hosted service likely does.

FlacoJones:
Confirmed that `getNames` is a full second faster than provider.lookupAddress:

Over 10 calls:

`Average response time for resolveName (contract) after 10 calls is 447.33214750001207 milliseconds.`

`Average response time for resolveName_ethers after 10 calls is 1256.3636573001743 milliseconds.`

Will revert to that.

FlacoJones:
> And resolving the name also could be benchmark with getNames vs lookupAddress

Confirmed that getNames is a full second faster than provider.lookupAddress:

Over 10 calls:

Average response time for resolveName (contract) after 10 calls is 447.33214750001207 milliseconds.

Average response time for resolveName_ethers after 10 calls is 1256.3636573001743 milliseconds.

Will revert to that.

FlacoJones:
New results in.

This compares the CURRENT approach to an approach where we first get the ENS name directly from the contract as before, and then use `provider.getAvatar(name)` provided by the ethers.js library you can see [here](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/base-provider.ts#L2036).

```
Average response time for ORIGINAL (likely with caching on metadata server) after 4 calls is 3471.11 milliseconds.

Average response time for "getName from contract, then provider.getAvatar(name)" after 4 calls is 2560.87 milliseconds.
```

So a full second faster, uncached.

It probably takes about the same time as it does on the ENS servers, with the extra second accounting for network latency.

FlacoJones:
I'm assuming that if even @ethersproject follows the "get name from address, then get resolver, then get avatarUrl" flow then that might be the way to go, relying on them rather than us to maintain against many edge cases.

FlacoJones:
The code for `resolver.getAvatar()` is quite a bit more involved than just 'get the text'. It looks like they account for quite a lot of edge cases in the avatar scheme. Code here for [resolver.getAddress()](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/base-provider.ts#L512)

FlacoJones:
PS: Benchmarking times were for "time to get url," not including any image processing

FlacoJones:
I benchmarked the new method again at home on better WiFi and now its down to just 1.5 seconds including resizing...

package.json
```diff
@@ -6,11 +6,13 @@
...
     "build": "tsc",
     "dev": "nodemon src/index.ts",
-    "start": "node build/src/index.js"
+    "start": "node build/src/index.js",
+    "benchmark": "ts-node __tests__/benchmarking.ts"
   },
   "dependencies": {
     "@aws-sdk/client-s3": "^3.51.0",
     "@download/blockies": "^1.0.3",
+    "@ensdomains/ens-avatar": "^0.1.4",
```
bonustrack, line 15: This should be removed

src/resolvers/ens.ts
```diff
@@ -1,21 +1,24 @@
 import axios from 'axios';
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
-import { Contract } from '@ethersproject/contracts';
 import { resize } from '../utils';
 import { max } from '../constants.json';
+import { Contract } from 'ethers';
 
-async function resolveName(network, address) {
+export async function resolveName(address) {
```
bonustrack, line 7: any reason to export the function?

src/resolvers/ens.ts
```diff
@@ -1,21 +1,24 @@
 import axios from 'axios';
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
-import { Contract } from '@ethersproject/contracts';
 import { resize } from '../utils';
 import { max } from '../constants.json';
+import { Contract } from 'ethers';
```
bonustrack, line 5: Can you revert to @ethersproject/contracts? I'd prefer to install just what's needed

package.json
FlacoJones: removed

src/resolvers/ens.ts
FlacoJones: nope. removed

src/resolvers/ens.ts
FlacoJones: reverted

FlacoJones:
Do you want me to leave the benchmarking utility in there under the new `__tests__` folder or remove that too? 

Benchmark.js is thousands of lines of overkill for what we needed so I just whipped it up. Obviously not related to ENS avatars though

bonustrack:
Thanks @FlacoJones sorry missed you PR but yes I think it's better to remove the benchmark utils from the code. If you can make this change we can merge this PR

FlacoJones:
Removing

bonustrack (approved):
Looks good! Thanks @FlacoJones

Issue #15, opened 2022-05-27
Add tests for resolvers


bonustrack:
We should add tests for each resolvers with Jest, here is an example repo with Jest test https://github.com/snapshot-labs/checkpoint we can use same format

PR #17, opened 2022-06-04
test: add resolver tests

Sekhmet:
Closes https://github.com/snapshot-labs/stamp/issues/15

`space` is not tested because I couldn't find any addresses that would resolve for it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,15 @@
+name: Test
+on: [push]
+
+jobs:
+  build-test:
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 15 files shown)


bonustrack:
@Sekhmet for space it's not using addresses actually but ens name instead like "balancer.eth" or "ens.eth", this correspond to the spaces ids that you see in the urls like https://snapshot.org/#/ens.eth

Sekhmet:
@bonustrack thanks, I tried it before and it didn't work - I think I was testing against wrong resolver then.

bonustrack (approved):
Works well for me! Feel free to merge

Issue #1, opened 2022-02-19
Add Ceramic IDX avatar resolver


bonustrack:
We should have a resolver for avatars from Ceramic IDX solution.

bonustrack:
Seem like Ceramic avatar solution is moving to this website https://self.id

PR #18, opened 2022-06-11
feat: add SelfID resolver

Sekhmet:
Closes: https://github.com/snapshot-labs/stamp/issues/1

Adds SelfID resolver.

Needs to use `https://ceramic-private.3boxlabs.com` as ceramic API, otherwise it will give you different DID and won't resolve the same avatar as self.id.

## Test plan
- Sign up at https://self.id/ and add your avatar.
- Go to http://localhost:3000/avatar/YOUR_ADDRESS (http://localhost:3000/avatar/0x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70)
- It gives you back your avatar (I've seen some issues that Snapshot's pinata would take long time to resolve this image (few minutes after first attempt it will start working).

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@aws-sdk/client-s3": "^3.51.0",
     "@download/blockies": "^1.0.3",
+    "@self.id/core": "^0.3.0",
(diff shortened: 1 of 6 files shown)


bonustrack (approved):
Looks good!

Issue #14, opened 2022-05-19
Add Lens protocol avatar image as resolver


bonustrack:
We should add Lens protocol profile within Stamp here is some docs: https://docs.lens.xyz/docs/get-default-profile

PR #19, opened 2022-06-11
feat: add Lens resolver

Sekhmet:
Closes: https://github.com/snapshot-labs/stamp/issues/14

Adds [Lens.xyz](https://lens.xyz/) resolver. Couldn't test with my own account as it looks like it's not open yet for everyone.

## Test plan
- Disable `ens` resolver if you are testing with addresses that have the same image on `ens` as on `lens` so you can tell if it works or not.
- Try going to http://localhost:3000/avatar/fabien.lens
- It should return an image.
- Try going to http://localhost:3000/avatar/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7
- It should return an image.

### DIFF
--- a/src/resolvers/lens.ts
+++ b/src/resolvers/lens.ts
@@ -13,12 +13,9 @@ function getDefaultImage(picture) {
 }
 
 export default async function resolve(address) {
-  const path = isAddress(address) ? 'defaultProfile' : 'profile';
-
-  const request =
-    path === 'defaultProfile'
-      ? `${path}(request: { ethereumAddress: "${getAddress(address)}"})`
-      : `${path}(request: { handle: "${address}"})`;
+  const request = isAddress(address)
+    ? `{ ownedBy: "${getAddress(address)}", limit: 1 }`
+    : `{ handles: ["${address}"], limit: 1 }`;
 
   try {
     const { data } = await axios({
@@ -27,19 +24,21 @@ export default async function resolve(address) {
       data: {
         query: `
             query Profile {
-              ${request} {
-                picture {
-                  ... on NftImage {
-                    contractAddress
-                    tokenId
-                    uri
-                    chainId
-                    verified
-                  }
-                  ... on MediaSet {
-                    original {
-                      url
-                      mimeType
+              profiles(request: ${request}) {
+                items {
+                  picture {
+                    ... on NftImage {
+                      contractAddress
+                      tokenId
+                      uri
+                      chainId
+                      verified
+                    }
+                    ... on MediaSet {
+                      original {
+                        url
+                        mimeType
+                      }
                     }
                   }
                 }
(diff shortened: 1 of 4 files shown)


bonustrack (commented):
Why it need to disable ENS resolver? Currently in the defined order Lens would be the primary resolver, ENS would be visible only if there is no Lens Snapshot or avatar

When I test with http://localhost:3000/avatar/0x3A5bd1E37b099aE3386D13947b6a90d97675e5e3 it give me the same image than on https://stamp.fyi/avatar/0x3A5bd1E37b099aE3386D13947b6a90d97675e5e3
I tried locally and it seem like the response is empty from GraphQL, you can test with my account:
https://lenster.xyz/u/fabien.lens
0xeF8305E140ac520225DAf050e2f71d5fBcC543e7

Sekhmet:
> Why it need to disable ENS resolver? Currently in the defined order Lens would be the primary resolver, ENS would be visible only if there is no Lens Snapshot or avatar

I mentioned disabling ENS just for testing, because this address has the same avatar on Lens and ENS so you won't know if it works. That's why stamp.fyi works now (it gets it from ENS).


> I tried locally and it seem like the response is empty from GraphQL, you can test with my account:
https://lenster.xyz/u/fabien.lens
0xeF8305E140ac520225DAf050e2f71d5fBcC543e7

Thanks. I don't have access to Lens ("SORRY YOUR ADDRESS IS NOT ELIGIBLE") so all I could test with was that one sample address from docs.
This address seems to work with Lens: http://localhost:3000/avatar/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7
I will look into making handles work, didn't know those are a thing

Sekhmet:
@bonustrack updated, now it should work with bot addresses and Lens handles.

It won't work for `0xeF8305E140ac520225DAf050e2f71d5fBcC543e7` because you would need to set it as your default portfolio ([docs](https://docs.lens.xyz/docs/create-set-default-profile-typed-data)). There might be way to do it on the UI as well.

src/resolvers/lens.ts
```diff
@@ -1,19 +1,33 @@
...
+
+  return null;
+}
+
 export default async function resolve(address) {
+  const path = isAddress(address) ? 'defaultProfile' : 'profile';
+
+  const request =
+    path === 'defaultProfile'
+      ? `${path}(request: { ethereumAddress: "${getAddress(address)}"})`
```
bonustrack, line 20: I've noticed that on Lenster UI https://lenster.xyz/u/fabien.lens the request is actually doing something like: 
```
profile(request: { ownedBy: "${address}"})
```
Using the Ethereum address, maybe this help getting avatar even if the image is not set as default

bonustrack (approved):
Works well now

Issue #2, opened 2022-02-19
Add Jazzicon resolver


bonustrack:
We should have a resolver for Jazzicon https://github.com/danfinlay/jazzicon This is a popular alternative to Blockie, used in MetaMask for example. Jazzicon should be used as fallback resolver, we can introduce a param "fb" by default the fallback would be set to blockie but if someone set fb=jazzicon it would fallback to Jazzicon resolver

PR #20, opened 2022-06-11
feat: add Jazzicon resolver

Sekhmet:
Closes https://github.com/snapshot-labs/stamp/issues/2

## Test plan
- http://localhost:3000/avatar/0x290ADCcA6253aCe88b10A6bb34C07a5Ad10fC6B0 returns blockie fallback
- http://localhost:3000/avatar/0x290ADCcA6253aCe88b10A6bb34C07a5Ad10fC6B0?fb=jazzicon returns jazzicon fallback
- Test with your own address, it should look identical to your address in metamask

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,5 +1,5 @@
 import express from 'express';
-import { parseQuery, resize, setHeader, sha256 } from './utils';
+import { parseQuery, resize, setHeader, sha256, getCacheKey } from './utils';
 import { set, get, streamToBuffer, clear } from './aws';
 import resolvers from './resolvers';
 import constants from './constants.json';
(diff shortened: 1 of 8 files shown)


src/api.ts
```diff
@@ -20,6 +20,8 @@ router.get('/clear/:type/:id', async (req, res) => {
 
 router.get('/:type/:id', async (req, res) => {
   const { type, id } = req.params;
+  const { fb } = req.query;
+
   const { address, network, w, h } = await parseQuery(id, type, req.query);
   const key1 = sha256(
     JSON.stringify({ type, network, address, w: constants.max, h: constants.max })
```
bonustrack, line 27: Awesome! It seem to work well. Just here we should add the fb param with 'blockie' or 'jazzicon' to create the key1 and also key2, these keys are used to get and store caches, if we dont include them here we may render an image with jazzicon as fallback even if the request doesnt include fb param.

src/resolvers/jazzicon.ts
```diff
@@ -0,0 +1,19 @@
...
+import sharp from 'sharp';
+import jazzicon from '@metamask/jazzicon';
+import { resize } from '../utils';
+import { max } from '../constants.json';
+
+const dom = new JSDOM('');
+global.document = dom.window.document;
+
+export default async function resolve(address) {
+  const { innerHTML, style } = jazzicon(64, parseInt(address.slice(2, 10), 16));
```
bonustrack, line 11: Just changing this otherwise it make low quality on large size image
```suggestion
  const { innerHTML, style } = jazzicon(500, parseInt(address.slice(2, 10), 16));
```

bonustrack (approved):
Awesome!

PR #31, opened 2022-09-26
feat: add crossbell resolver

AtlasQuan:


bonustrack:
@AtlasQuan Thanks for the PR! I wasn't aware about Crossbell, do you know what is the coverage, or how many addresses have their avatar set on Crossbell?

bonustrack:
@AtlasQuan Ok interesting, so Crossbell is also a chain, do you know how many of these 31k addresses have activity in mainnet? When adding a new resolver we need to make sure there is enough coverage, cuz it will increase loading time of the avatar on first load.

AtlasQuan:
I just wrote a script to calculate how many Crossbell users are active (having at least 1 transaction) on mainnet. The result is 25149 out of 31034. Plus, Crossbell is still at a very early dev stage, meaning there might be more users joining in this ecosystem in the future. Let me know what you think.

The snippet is below.

```
import ethers from "ethers";

const provider = new ethers.providers.WebSocketProvider(/* ws uri */);

const fetchAddresses = async (
	pageToken = "/token/0xa6f969045641cf486a747a2688f3a5a6d43cd0d8/token-holders?"
) => {
	return fetch(`[https://scan.crossbell.io/${pageToken}&type=JSON](https://scan.crossbell.io/$%7BpageToken%7D&type=JSON)`)
		.then((res) => res.json())
		.then((res) => {
			res.items = res.items.map((item) => item.match(/0x\w{40}/)[0]);
			return res;
		});
};

let result = 0;
let pageToken;
const main = async () => {
	while (true) {
		const { items, next_page_path } = await fetchAddresses(pageToken);

		await Promise.all(
			items.map(async (address) => {
				const count = await provider.getTransactionCount(address);

				console.log({ count });
				if (count > 0) {
					result++;
				}
			})
		);

		console.log({ result, next_page_path });

		if (!next_page_path) {
			break;
		}

		pageToken = next_page_path;
	}

	console.log({ result });
};

main();
```

image

DIYgod:
Hi, is there any progress on this pr? We are using stamp on [RSS3](https://rss3.io/), but have been missing this resolver that we want.

PR #33, opened 2022-09-28
doc: add RSS3 into the integration list

Songkeys:
image

They are using this service as well!

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ cdn.stamp.fyi/avatar/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7?**cb=1**
 #### [Snapshot](http://snapshot.org)
 #### [Parcel](https://parcel.money)
 #### [Lenster](https://lenster.xyz)
+#### [RSS3](https://rss.3)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,7 @@ cdn.stamp.fyi/avatar/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7?**cb=1**
 #### [Snapshot](http://snapshot.org)
 #### [Parcel](https://parcel.money)
 #### [Lenster](https://lenster.xyz)
+#### [RSS3](https://rss.3)
```
bonustrack, line 56: This is not a correct domain, I assume this is the right one? 
```suggestion
#### [RSS3](https://rss3.io)
```
Could you tell me where RSS3 use Stamp? I tried to login on the app but I can't see a Stamp avatar.

Songkeys: Oh yes! Sorry about this. I must be so drunk to write things like that 😓. Thank you for correcting!

bonustrack:
Perfect thanks it's up!

PR #39, opened 2023-01-10
fix: write resized images to correct location

Sekhmet:
We read it from `key1/key2` and we wrote it to `key2/key1`.


bonustrack (approved):
utACK

bonustrack:
Good catch, this mean we were not using caching for resized images for all this time 😮

Issue #40, opened 2023-01-11
Avatar cache issue


ChaituVR:
https://discord.com/channels/955773041898573854/1030772384308932608/1062305484771688540

Sekhmet:
@ChaituVR have we seen this issue recently?

ChaituVR:
Didn't see any reports recently :) Will close this then.

PR #41, opened 2023-01-11
fix: lift cache on fallback images

Sekhmet:
Closes: https://github.com/snapshot-labs/stamp/issues/40

Currently we have couple of issues with how cache works:
- If fallback image is served it is still written to S3 so it will be stuck to fallback image forever (or as long as our S3 expiry policy is configured for) as cache takes priority. I don't think we should cache fallback images in S3 at all, instead just cache it at edge and just lower TTL for cache.
- Fallback gets TTL of 1 day which seems overly pessimistic, we can lower it to 1h, so if avatar is set (or some issue that caused it to fail to load no longer exists) it can update itself quicker.

## Test plan

- Setup local repo with AWS S3.
- Return false in space.ts
- Visit: http://localhost:3000/space/0xbattleground.eth?s=160 -> you get fallback image, Cache-Control is set to 3600
- Remove false in space.ts
- Visit that URL again -> you get proper image, Cache-Control is set to 86400

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -63,7 +63,7 @@ router.get('/:type/:id', async (req, res) => {
       const fallbackImage = await resolvers[fallback](address, network);
       const resizedImage = await resize(fallbackImage, w, h);
 
-      setHeader(res, 'SHORT');
+      setHeader(res, 'SHORT_CACHE');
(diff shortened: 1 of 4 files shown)


bonustrack (approved):
tACK

PR #51, opened 2023-04-11
feat: add space-sx resolver

j0hnfl0w:
Related https://github.com/snapshot-labs/sx-ui/issues/419

### DIFF
--- a/src/constants.json
+++ b/src/constants.json
@@ -5,6 +5,6 @@
   "resolvers": {
     "avatar": ["lens", "ens", "snapshot"],
     "token": ["trustwallet", "zapper"],
-    "space": ["space"]
+    "space": ["space", "space-sx"]
   }
 }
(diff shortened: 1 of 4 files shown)


Sekhmet (changes requested):
nACK

It needs to be exported in resolvers/index.ts, otherwise it will crash.

src/constants.json
```diff
@@ -5,6 +5,6 @@
   "resolvers": {
     "avatar": ["lens", "ens", "snapshot"],
     "token": ["trustwallet", "zapper"],
-    "space": ["space"]
+    "space": ["space", "spacesx"]
```
Sekhmet, line 8: I'm not sure if it makes sense to have them as same entity, if we want to keep it the same I'd suggest reversing the order, because this impacts priority (last element overrides previous ones).

I think it would make sense to create separate resolver for `space-sx` (will need to update UI then).

What do you think @bonustrack ?

bonustrack: Yes agree it should be a separate resolver here, we don't want load image on both S and SX for the same request.

j0hnfl0w: https://github.com/snapshot-labs/sx-ui/pull/576

Sekhmet (approved):
tACK

PR #54, opened 2023-05-01
chore: Improve linter

wa0x6e:
Current lint task on github action is using `--fix`, muting all errors. 

## Changes

- Add a new `yarn:nofix` task, which will show warning/error when run, instead of fixing errors. This task will raise error properly on github actions
- Add a new `typecheck` taks, which will raises error on typescript errors
- Upgrade dependencies related to lint
- Use lint config from snapshot-labs (also removing the local eslint and prettier config file)


wa0x6e:
Fixed by #62

PR #61, opened 2023-07-07
fix: fix failing ENS resolver

### LINKED ISSUE #59: bug: ENS provider failing
There is some issue with ENS provider, test is failing, using `provider.getAvatar` seem to always return null.

wa0x6e:
fix #59

### DIFF
--- a/src/resolvers/ens.ts
+++ b/src/resolvers/ens.ts
@@ -2,10 +2,10 @@ import axios from 'axios';
 import { getProvider, resize } from '../utils';
 import { max } from '../constants.json';
 
-export default async function resolve(address: string) {
+export default async function resolve(name: string) {
   try {
     const provider = getProvider(1);
-    const url = await provider.getAvatar(address);
+    const url = await provider.getAvatar(name);
     if (!url) return false;
     const input = (await axios({ url, responseType: 'arraybuffer' })).data as Buffer;
     return await resize(input, max, max);
--- a/src/resolvers/ens.ts
+++ b/src/resolvers/ens.ts
@@ -1,22 +1,11 @@
 import axios from 'axios';
 import { getProvider, resize } from '../utils';
 import { max } from '../constants.json';
-import { Contract } from '@ethersproject/contracts';
 
-async function resolveName(address) {
-  const provider = getProvider(1);
-  const abi = ['function getNames(address[]) view returns (string[])'];
-  const contract = new Contract('0x3671aE578E63FdF66ad4F3E12CC0c0d71Ac7510C', abi, provider);
-  const names = await contract.getNames([address]);
-  return names[0];
-}
-
-export default async function resolve(address) {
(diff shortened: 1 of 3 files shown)


src/resolvers/ens.ts
```diff
@@ -1,22 +1,11 @@
...
-}
-
-export default async function resolve(address) {
+export default async function resolve(address: string) {
   try {
     const provider = getProvider(1);
-    const name = await resolveName(address);
-    if (!name) return false;
-    const url = await provider.getAvatar(name);
+    const url = await provider.getAvatar(address);
```
ChaituVR, line 8: This method always return null 🤔

ChaituVR (changes requested):

src/resolvers/ens.ts
wa0x6e: Why do you think so ? The tests are passing, and it does return a correct url when image exists. What are you passing in the arg ?

function argument name is misleading, and should be `name` and not `address`, I will fix that

wa0x6e: the resolver only accepts an ENS name as agument, and not an eth address. Should we also support that ?

wa0x6e: In fact, this resolver will always get an ENS name as argument (https://github.com/snapshot-labs/stamp/blob/641daa145c4a3c542270468888073aa6d21c4f7f/src/utils.ts#L61)

ChaituVR: so here is what i did to reproduce.

If i go here https://stamp.fyi/avatar/fabien.eth i see 
image
same as here https://app.ens.domains/fabien.eth

Then if i run this code, and go to http://localhost:3000/avatar/fabien.eth i see:
image

so i've added a `console.log(url);` in `line 9` - It returns `null`
and with previous changes also it is `null` 🙈

ChaituVR: image
Here getAvatar is top of `resolver` not on top of `provider`. could be because of that, but i am not 100% sure

wa0x6e: Ok, I pushed another fix, still getting nothing from Fabien.eth, but everything else is working

Test with:

- http://localhost:3000/avatar/ricmoo.eth
- http://localhost:3000/avatar/snowowl.eth

Issue with fabien.eth is that the avatar url is not a http link but an eip address (`eip155:1/erc1155:0x495f947276749Ce646f68AC8c248420045cb7b5e/61751686310106735452369515624497384717106902022558848289817630643456812515329`), making `getAvatar()` return false.

The fix is now returning the raw text, but still need a way to convert that eip155 address to an http one

wa0x6e: Do you have some pointer on how to convert the eip155 address to some more conventional http or something else ?

ChaituVR (approved):
tAck

PR #73, opened 2023-07-28
chore: use snapshot-sentry to handle sentry

wa0x6e:
Use https://www.npmjs.com/package/@snapshot-labs/snapshot-sentry to handle everything related to Sentry


ChaituVR (approved):
utAck

Issue #76, opened 2023-08-04
Fix NFT profile pictures


ChaituVR:
Issue with fabien.eth is that the avatar url is not a http link but an eip address (`eip155:1/erc1155:0x495f947276749Ce646f68AC8c248420045cb7b5e/61751686310106735452369515624497384717106902022558848289817630643456812515329`), making `getAvatar()` return false.

The fix is now returning the raw text, but still need a way to convert that eip155 address to an http one

_Originally posted by @wa0x6e in https://github.com/snapshot-labs/stamp/pull/61#discussion_r1280689987_

PR #80, opened 2023-08-07
fix: return false when ENS avatar is not an URL

wa0x6e:
The current ENS resolver is returning the raw value from ENS, and it's working great when the value is an http url.

But when it's an NFT address, such as `eip155:1/erc1155:0x495f947276749Ce646f68AC8c248420045cb7b5e/61751686310106735452369515624497384717106902022558848289817630643456812515329` we should return `false`, and not continue with trying to fetch the image.

See https://github.com/snapshot-labs/stamp/issues/76


ChaituVR (approved):
utAck

PR #102, opened 2023-10-10
fix: fix missing `cb` query params when clearing cache

wa0x6e:

When clearing the cache, the `cb` query params is ignored

## 💊 Fixes / Solution

- Add the `cb` query params when computing the cache key


ChaituVR (approved):
utAck

PR #113, opened 2023-11-17
feat: add addresses resolvers

wa0x6e:
From https://github.com/snapshot-labs/pitches/issues/58

Add new endpoint to return Ens, Lens, or unstoppableDomains handle associated to an address


src/addressResolvers/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as ensResolver from './ens';
+import * as lensResolver from './lens';
+import * as unstoppableDomainsesolver from './unstoppableDomains';
+import { Address } from './utils';
+import cache from './cache';
+
+const RESOLVERS = [ensResolver, unstoppableDomainsesolver, lensResolver];
+
+export async function lookupAddresses(addresses: Address[]) {
+  const normalizedAddresses = addresses.slice(0, 250).map(a => getAddress(a));
```
ChaituVR, line 11: Better to return an error if we pass more than `250` addresses

ChaituVR:
```suggestion
  const normalizedAddresses = addresses.slice(0, 250).map(getAddress);
```

src/addressResolvers/cache.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  return transaction.exec();
+}
+
+export default async function cache(addresses: Address[], callback) {
+  const cache = await getCache(addresses);
+  const cachedAddresses = Object.keys(cache);
+
+  const results = await callback(addresses.filter(a => !cachedAddresses.includes(a)));
```
ChaituVR, line 36: we never use callbacks 🙈

src/addressResolvers/cache.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+  const results = await transaction.exec();
+
+  return Object.fromEntries(
+    addresses
+      .map((address, index) => {
+        return [address, results[index]];
+      })
+      .filter(([, handle]) => handle !== null)
+  );
```
ChaituVR, line 20:
```suggestion
  return Object.fromEntries(
    addresses
      .map((address, index) => [address, results[index]])
      .filter(([, handle]) => handle)
  );
```

src/addressResolvers/cache.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  return transaction.exec();
+}
+
+export default async function cache(addresses: Address[], callback) {
+  const cache = await getCache(addresses);
+  const cachedAddresses = Object.keys(cache);
+
+  const results = await callback(addresses.filter(a => !cachedAddresses.includes(a)));
+  await setCache(results);
```
bonustrack, line 37: We should not wait for `setCache`, it's not useful to resolve the request, would be faster without it

bonustrack: Also here do you cache the cached results? We shouldn't otherwise there might be some address where cache is never removed we should only cache fresh results

wa0x6e: Only caching fresh results, since it's filtering out already cached addresses.

src/addressResolvers/cache.ts
wa0x6e: Suggestion is not working, since we only want to filter out `null` values, and keep empty string

src/addressResolvers/cache.ts
wa0x6e: We do, in requestDeduplicator

ChaituVR: requestDeduplicator is different case but i haven't seen us using callbacks for normal usecase

src/addressResolvers/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (addresses.length > MAX_LOOKUP_ADDRESSES) {
+    return Promise.reject({
+      error: `params must contains less than ${MAX_LOOKUP_ADDRESSES} addresses`,
+      code: 400
+    });
+  }
+
+  let normalizedAddresses: Address[];
+  try {
+    normalizedAddresses = addresses.map(a => getAddress(a));
```
ChaituVR, line 21:
```suggestion
    normalizedAddresses = addresses.map(getAddress);
```

bonustrack:
We should only return addresses with result, the ones with empty results should not be part of the response (but still should be cached)

wa0x6e:
> We should only return addresses with result, the ones with empty results should not be part of the response (but still should be cached)

done

src/addressResolvers/cache.ts
wa0x6e: How will you accomplish this without passing a function to the `cache` function, and without coupling the cache logic inside addressResolvers ?

And how is requestDeduplicator different ? We pass a function to it, as well as the expected args to pass to the function, and return the results depending on the function result.

src/addressResolvers/cache.ts
```diff
@@ -0,0 +1,38 @@
+import { Address, Handle } from './utils';
+import redis from '../helpers/redis';
+
+const KEY = 'address-resolvers';
+const TTL = 24 * 60 * 60 * 1e3;
+
+export async function getCache(addresses: Address[]): Promise<Record<Address, Handle>> {
+  const transaction = redis.multi();
```
ChaituVR, line 8: redis/cache should be optional no?

wa0x6e: Fixed

ChaituVR:
Something is weird, it doens't resolve name for address `0x24F15402C6Bb870554489b2fd2049A85d75B982f` 
It works if i pass this
```JSON
{
    "method": "lookup_addresses",
    "params": [
        "0x24F15402C6Bb870554489b2fd2049A85d75B982f"
    ]
}
```
But not like this
```JSON
{
    "method": "lookup_addresses",
    "params": [
        "0xE6D0Dd18C6C3a9Af8C2FaB57d6e6A38E29d513cC",
        "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
        "0x0C67A201b93cf58D4a5e8D4E970093f0FB4bb0D1",
        "0x24F15402C6Bb870554489b2fd2049A85d75B982f"
    ]
}
```

ChaituVR:
Also, any idea why test cases fail 
image

In local i get this 
image

wa0x6e:
> Also, any idea why test cases fail image
> 
> In local i get this image

Will try to fix in another PR

src/addressResolvers/cache.ts
```diff
@@ -0,0 +1,42 @@
+import { Address, Handle } from './utils';
+import redis from '../helpers/redis';
+
+const KEY = 'address-resolvers';
+const TTL = 24 * 60 * 60 * 1e3;
```
ChaituVR, line 5: This is actually two years 😄 
image

ChaituVR:
```suggestion
const TTL = 24 * 60 * 60;
```

ChaituVR: Actually there is a config file https://github.com/snapshot-labs/stamp/blob/add-addresses-resolvers/src/constants.json

src/addressResolvers/index.ts
```diff
@@ -0,0 +1,40 @@
+import { getAddress } from '@ethersproject/address';
+import * as ensResolver from './ens';
+import * as lensResolver from './lens';
+import * as unstoppableDomainsesolver from './unstoppableDomains';
+import { Address } from './utils';
+import cache from './cache';
+
+const RESOLVERS = [ensResolver, unstoppableDomainsesolver, lensResolver];
```
ChaituVR, line 8:
```suggestion
import * as unstoppableDomainresolver from './unstoppableDomains';
import { Address } from './utils';
import cache from './cache';

const RESOLVERS = [ensResolver, unstoppableDomainresolver, lensResolver];
```

ChaituVR (approved):
tAck

bonustrack (approved):
tACK

PR #115, opened 2023-11-20
feat: add handle resolver

wa0x6e:
This is part 2 of pitch https://github.com/snapshot-labs/pitches/issues/58

Following on https://github.com/snapshot-labs/stamp/pull/113

Adding `resolveNames` function to convert handle (domain) to address

The goal will be to remove from this file https://github.com/snapshot-labs/snapshot/blob/master/src/helpers/utils.ts

- resolveEns
- resolveLens 

And replace by a simple `resolveName` to call stamp api.

We currently do not resolve unstoppableDomains

Cache engine also refactored to be able to cache data from both `resolveNames` and `lookupAddresses`.

The method accepts an array of domains, up to 5 domains (limit is for unstoppableDomains, as the other resolvers uses graphql to resolve)

### Test

Testing with a mix of existing and non-existing domains

```bash
curl -X POST localhost:3008/ -H "Content-Type: application/json" -d '{"method": "resolve_names", "params": ["fabien.eth", "chaitu.eth", "snapshot.crypto", "test.snapshot", "google.com"] }'
# return
{"jsonrpc":"2.0","result":{"fabien.eth":"0xeF8305E140ac520225DAf050e2f71d5fBcC543e7","chaitu.eth":"0x24F15402C6Bb870554489b2fd2049A85d75B982f","snapshot.crypto":"0xeF8305E140ac520225DAf050e2f71d5fBcC543e7"},"id":null}
```

Non-existing domains are omitted from the results

### Todo 

- [x] Add cache
- [x] Add tests


src/addressResolvers/cache.ts
```diff
@@ -1,42 +1,42 @@
...
+export default async function cache(input: string[], callback) {
+  const cache = await getCache(input);
+  const cachedKeys = Object.keys(cache);
 
-  const results = await callback(addresses.filter(a => !cachedAddresses.includes(a)));
-  setCache(results);
+  const callbackArgs = input.filter(a => !cachedKeys.includes(a));
+  if (callbackArgs.length > 0) {
+    const results = await callback(callbackArgs);
+    setCache(results);
```
ChaituVR, line 36: Better name, `callbackArgs` is bit confusing
```suggestion
  const uncachedInputs = input.filter(a => !cachedKeys.includes(a));
  if (uncachedInputs.length > 0) {
    const results = await callback(uncachedInputs);
    setCache(results);
```

src/addressResolvers/ens.ts
```diff
@@ -38,3 +38,12 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
     return {};
   }
 }
+
+export async function resolveName(handle: string): Promise<string | undefined> {
+  try {
+    const addressResolved = await provider(NETWORK).resolveName(handle);
```
ChaituVR, line 44: Better to move `provider(NETWORK)` outside this function so it can be reused, - NOT too important

src/addressResolvers/unstoppableDomains.ts
```diff
@@ -21,3 +22,21 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
...
   }
 }
+
+export async function resolveName(handle: string): Promise<Address | undefined> {
+  const abi = ['function ownerOf(uint256 tokenId) external view returns (address address)'];
+
+  try {
+    const tokenId = new Resolution().namehash(handle, NamingServiceName.UNS);
+    return await snapshot.utils.call(
+      provider(NETWORK),
```
ChaituVR, line 32: same here

src/addressResolvers/lens.ts
```diff
@@ -25,13 +36,29 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
...
+        `);
 
     return Object.fromEntries(items.map(i => [i.ownedBy, i.handle])) || {};
   } catch (e) {
     capture(e);
     return {};
   }
 }
+
+export async function resolveName(handle: string): Promise<string | undefined> {
```
ChaituVR, line 48: But `params` passed in `src/api.ts` is an array 🤔

ChaituVR (changes requested):
Could share me the curl to test this, not able to get this working

src/addressResolvers/lens.ts
wa0x6e: params is a string in the case of `resolve_name`

ChaituVR: like this? 
```JSON
{
    "method": "resolve_name",
    "params": "chaitu.eth"
}
```

Doesn't work, i get `ResolutionError: Domain chaitu.eth is not supported`

src/addressResolvers/index.ts
```diff
@@ -38,3 +38,22 @@ export async function lookupAddresses(addresses: Address[]) {
     ).filter(([, handle]) => handle)
   );
 }
+
+export async function resolveName(handle: Handle): Promise<Address | undefined> {
+  const results = await cache([handle], async (h: Handle[]) => {
+    let address: Address | undefined;
+    const _handle = h[0];
+
+    if (_handle.endsWith('.ens')) {
```
ChaituVR, line 47: why is it .ens 🙈

wa0x6e:
> Could share me the curl to test this, not able to get this working

Original post update with test instructions, and expected output

src/addressResolvers/lens.ts
ChaituVR: Anyway i believe it should be an array if it is named `params`

wa0x6e: Fixed, api should work now.

What var name would be appropriate ?

ChaituVR: I would still keep it array and limit it to length of one 😄 I checked eth RPC calls https://composer.alchemy.com/ and there are many requests that use `params` with a single item 

else maybe `input` or `name` is more relevant

src/api.ts
```diff
@@ -15,11 +15,14 @@ router.post('/', async (req, res) => {
...
+    if (!Array.isArray(params)) return rpcError(res, 400, 'params must be an array of string', id);
+
     if (method === 'lookup_addresses') result = await lookupAddresses(params);
+    else if (method === 'resolve_name') result = await resolveName(params[0]);
     else return rpcError(res, 400, 'invalid method', id);
 
-    if (result.error) return rpcError(res, result.code || 500, result.error, id);
-    return rpcSuccess(res, result, id);
+    if (result?.error) return rpcError(res, result.code || 500, result.error, id);
+    return rpcSuccess(res, { address: result[params[0]] }, id);
```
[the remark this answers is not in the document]
wa0x6e: I feel accepting array may confuse people, as they may think they can pass multiple values, but it will always only resolve the first one.

Would it be better if we just add support for multiple values ?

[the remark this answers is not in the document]
ChaituVR: Yea, feel like better to return an error if more than one name is passed, but if it is easy, we can add a limit up to 5 names for now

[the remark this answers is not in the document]
wa0x6e: Ok, will proceed to a refactoring to add support to multiple handle resolbver

wa0x6e:
Refactoring in progress to add support for multiple name resolvers

src/addressResolvers/unstoppableDomains.ts
```diff
@@ -1,23 +1,59 @@
...
+          return;
+        }
+      })
+    );
+
+    return withoutEmptyValues(
+      Object.fromEntries(handles.map((handle, index) => [handle, results[index]]))
+    );
+  } catch (e) {
+    console.log(e);
```
ChaituVR, line 55:
```suggestion
```

ChaituVR (approved):
tAck

PR #118, opened 2023-11-21
fix: Clear cache doesn't work with `cb` param

ChaituVR:
## Issues
- Not able to clear cache for images with `cb` param, for example https://cdn.stamp.fyi/space/wen%F0%9F%8C%95.eth?s=96&cb=c178562f8f12066b
- If there is nothing to clear, It returns `ok` in response 

How to test:
- Try going to http://localhost:3008/space/thanku.eth?s=160&cb=ce8ed162479340899ca4cae882bd0a0e
- Clear cache by going to http://localhost:3008/clear/space/thanku.eth?s=160&cb=ce8ed162479340899ca4cae882bd0a0e

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -29,14 +29,14 @@ router.post('/', async (req, res) => {
 router.get(`/clear/:type(${TYPE_CONSTRAINTS})/:id`, async (req, res) => {
   const { type, id } = req.params;
   try {
-    const { address, network, w, h, fallback } = await parseQuery(id, type, {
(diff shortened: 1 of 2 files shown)


wa0x6e:
This does take care of clearing the cache when passing the exact same params as when creating it, but it will not clear the cache for all resized versions

ChaituVR:
It deletes all items inside the folder

we saving different sizes inside a folder based on `cb` param

wa0x6e (approved):
utAck

PR #121, opened 2023-11-22
fix: ignore invalid addresses

wa0x6e:
There's a high error rate since introduction of addressResolvers (https://snapshot-labs.sentry.io/issues/4651451195/?project=4505506866593792&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=0)

Most of the errors come from invalid addresses given as params.

This PR will:
- ignore all invalid addresses, and process only valid one, instead of crashing and returning a 400 error for the whole request.
- log the addresses input given to the request, to debug why is there invalid addresses

### DIFF
--- a/src/addressResolvers/ens.ts
+++ b/src/addressResolvers/ens.ts
@@ -34,7 +34,7 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
         .filter((_, index) => !!validNames[index])
     );
   } catch (e) {
-    capture(e);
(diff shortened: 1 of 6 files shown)


ChaituVR (changes requested):
Feel like better to return error if an invalid address is passed
If there is wrong address, it is invalid input right?

- better to just ignore this error on sentry
- we need to identify who is/from where we are sending these invalid addresses

ChaituVR (approved):
utAck

PR #123, opened 2023-11-22
fix: skip resolver call when all addresses are already cached

### LINKED ISSUE #122: TypeError: Cannot read properties of null (reading 'profiles')
Sentry Issue: [STAMP-9](https://snapshot-labs.sentry.io/issues/4654544992/?referrer=github_integration)

```
TypeError: Cannot read properties of null (reading 'profiles')
  File "/workspace/dist/src/addressResolvers/lens.js", line 12, in Object.lookupAddresses
    const { data: { data: { profiles: { items } } } } = await (0, axios_1.default)({
  File "/workspace/dist/src/addressResolvers/index.js", line 52, in <anonymous>
    const results = await Promise.all(RESOLVERS.map(r => r.lookupAddresses(addresses)));
  File "/workspace/dist/src/addressResolvers/cache.js", line 34, in cache
    const results = await callback(addresses.filter(a => !cachedAddresses.includes(a)));
  File "/workspace/dist/src/addressResolvers/index.js", line 51, in lookupAddresses
    return Object.fromEntries(Object.entries(await (0, cache_1.default)(normalizedAddresses, async (addresses) => {
  File "/workspace/dist/src/api.js", line 23, in <anonymous>
    result = await (0, addressResolvers_1.lookupAddresses)(params);
...
(3 additional frame(s) were not displayed)
```

wa0x6e:

When all keys are already cached, the args passed to the callback function is an empty array

## 💊 Fixes / Solution

Fix #122 
Fix [STAMP-9](https://snapshot-labs.sentry.io/issues/4654544992/?referrer=github_integration)

We should skip calling the callback function when all keys are already cached

### DIFF
--- a/src/addressResolvers/cache.ts
+++ b/src/addressResolvers/cache.ts
@@ -34,9 +34,14 @@ export function setCache(payload: Record<Address, Handle>) {
 export default async function cache(addresses: Address[], callback) {
   const cache = await getCache(addresses);
   const cachedAddresses = Object.keys(cache);
+  const uncachedAddresses = addresses.filter(a => !cachedAddresses.includes(a));
 
-  const results = await callback(addresses.filter(a => !cachedAddresses.includes(a)));
-  setCache(results);
+  if (uncachedAddresses.length > 0) {
+    const results = await callback(uncachedAddresses);
+    setCache(results);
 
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

PR #128, opened 2023-11-23
fix: filter out invalid input before processing

wa0x6e:
This PR validates handles input, and filter out invalid domains before calling the resolver.

- ENS handle validation coming from ens_normalize package
- Lens validation regex coming from their error graphql error message (`Expected value of type \"Handle!\", found \"Fabien@.lens\"; Handle only supports lower case characters, numbers, - and _. Handle must be minimum of 5 length and maximum of 31 length - fabien@.lens`)
- UnstoppableDomains validation regex coming from https://github.com/unstoppabledomains/resolution/blob/3c10dced96d25c7de5247ca5234a29d8c9699830/src/utils/prepareAndValidate.ts


ChaituVR (approved):
tAck

PR #129, opened 2023-11-23
fix: Some ENS doesn't have resolved address

ChaituVR:
Ens like https://app.ens.domains/volet.eth doesn't have a resolved address


wa0x6e (approved):
tAck

PR #134, opened 2023-11-25
fix: decrease max number of `lookupAddresses` params

wa0x6e:

Current limit of addresses accepted for `lookupAddresses` is 250. This was a generous arbitrarily number chosen, without any rational.

We're now using graphql API to resolve the addresses, and they have a different limit

## 💊 Fixes / Solution

Decrease the number of accepted addresses, to match the resolvers API limit

- Lens limit is 50
- ENS limit is 1000

## 🚧 Changes

- Decrease the limit from 250 to 50 addresses. (this should not have any impact on UI, which is sending addresses in batch of 20)


ChaituVR (approved):
utAck

PR #144, opened 2023-11-29
fix: remove invalid and duplicate input

wa0x6e:
Before processing addressResolvers: 

- remove duplicate input
- remove invalid handles
- refactor regex comparison to use `test` instead of `match`


src/addressResolvers/unstoppableDomains.ts
```diff
@@ -13,7 +13,7 @@ const NETWORK = '137';
 const provider = getProvider(NETWORK);
 
 function normalizeHandles(handles: Handle[]): Handle[] {
-  return handles.map(h => (h.match(RegExp('^[.a-z0-9-]+$')) ? h : ''));
+  return handles.map(h => (/^[.a-z0-9-]+$/.test(h) ? h : ''));
 }
```
ChaituVR, line 17: Could we replace this with `normalizeHandles` in src/addressResolvers/utils.ts

ChaituVR (approved):
utAck

src/addressResolvers/unstoppableDomains.ts
wa0x6e: The one in utils is more global and lax, and is only filtering out invalid values like empty string, string with spaces, etc ...

The one inside each resolver are more tailored to the resolver, as they all have their own requirements 

- lens handle between 5 and 31 chars, [a-z0-0-_]
- ens allow emoji
- unstoppableDomains only allow [a-z0-0] and dash (no _)

ChaituVR: makes sense 👍

PR #146, opened 2023-12-01
fix: use ENS subgraph to query ENS data

wa0x6e:
Use ENS Subgraph (https://thegraph.com/hosted-service/subgraph/ensdomains/ens) to lookup addresses

### TODO 

- [ ] Check that this issue is fixed https://github.com/snapshot-labs/snapshot/issues/3987


wa0x6e:
Closing, will stay the the current contract call implementation due to https://github.com/snapshot-labs/snapshot/issues/3987

Current implementation is still failing on some addresses, such as `0x9c3aC02Cd616a82C83830e40D45c9534b32c4934` (may be related to expired domains)

PR #150, opened 2023-12-03
fix: filter out expired ENS handles from results

### LINKED ISSUE #147: fix(ens): resolve_names returns addresses for expired ENS names

wa0x6e:

When resolving ENS address from ENS's subgraph, we're returning all names associated to the address, without taking into account expiration date.

## 💊 Fixes / Solution

Fix #147

We should filter out expired domains from the results.

## 🚧 Changes

- Filter out expired domains from ENS results.
- Add relevant tests

## 🛠️ Tests

Test with the following query

```bash
curl -X POST http://localhost:3008/ -H "Content-Type: application/json" -d '{"method": "resolve_names", "params": ["49415.eth"] }'
```

This domain is expired (https://app.ens.domains/49415.eth)

Before fix, it returns

`{"jsonrpc":"2.0","result":{"49415.eth":"0xD3F0DC8F0aE3b5cd6FCB4a7e5dBC3154554d17F3"},"id":null}`

After fix, it returns

`{"jsonrpc":"2.0","result":{},"id":null}`

### DIFF
--- a/src/addressResolvers/ens.ts
+++ b/src/addressResolvers/ens.ts
@@ -65,7 +65,7 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
       `query Domains {
         domains(where: {
           name_in: ["${normalizedHandles.join('","')}"],
-          expiryDate_lt: "${Math.floor(Date.now() / 1e3)}"}
+          expiryDate_gt: "${Math.floor(Date.now() / 1e3)}"}
         ) {
           name
           resolvedAddress {
--- a/src/addressResolvers/ens.ts
+++ b/src/addressResolvers/ens.ts
@@ -63,7 +63,10 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
     } = await graphQlCall(
       'https://api.thegraph.com/subgraphs/name/ensdomains/ens',
       `query Domains {
-        domains(where: {name_in: ["${normalizedHandles.join('","')}"]}) {
+        domains(where: {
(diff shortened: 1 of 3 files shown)


src/addressResolvers/ens.ts
```diff
@@ -63,7 +63,10 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
     } = await graphQlCall(
       'https://api.thegraph.com/subgraphs/name/ensdomains/ens',
       `query Domains {
-        domains(where: {name_in: ["${normalizedHandles.join('","')}"]}) {
+        domains(where: {
+          name_in: ["${normalizedHandles.join('","')}"],
+          expiryDate_gt: "${Math.floor(Date.now() / 1e3)}"}
```
ChaituVR, line 68: `expiryDate` is null if name is a subdomain example `1.snapspace.eth` should be fine.?

wa0x6e: Will add some tests for that

wa0x6e: mhhh... subdomain does not return any address, even without the expiryDate filter.

Should we just convert all subdomains to domains, and use the domain's resolved address ? (and map it back to the subdomain)

ChaituVR: mmm getting complicated 😢 may be better to return the domain if doesn't have expiryDate for now

wa0x6e: Issue is that subdomain does not return any result, even on master.

wa0x6e:
Stale, need another PR

PR #156, opened 2023-12-13
fix: keep original input formatting when returning

### LINKED ISSUE #155: feat: address resolver should return addresses in the input format
Currently the address resolver API return object of addresses and names with addresses in the checksummed format. The issue is that it make harder on UI to use directly, it require mapping address manually because the input format is not the same as output. We should instead keep same format than the input addresses.

wa0x6e:
Fix #155

Keep the original input case formatting when returning result

When sending `['a.eth', 'A.ETH']`, will return result like 

```ts
{
'a.eth': '0x0',
'A.ETH': '0x0'
}
```

-> return a value for each original input, to avoid any processing from the caller side, even if the address os the same. This is done only after getting result from api and caching, so we don't pass `a.eth` and `A.ETH` to api.

## Test

```bash
curl -X POST http://localhost:3008/ -H "Content-Type: application/json" -d '{"method": "resolve_names", "params": ["snapshot.crypto", "SNAPSHOT.CRYPTO"] }'

// result
{"jsonrpc":"2.0","result":{"snapshot.crypto":"0xeF8305E140ac520225DAf050e2f71d5fBcC543e7","SNAPSHOT.CRYPTO":"0xeF8305E140ac520225DAf050e2f71d5fBcC543e7"},"id":null}
```

```bash
curl -X POST http://localhost:3008/ -H "Content-Type: application/json" -d '{"method": "lookup_addresses", "params": ["0xeF8305E140ac520225DAf050e2f71d5fBcC543e7", "0xEF8305E140AC520225DAF050E2F71D5FBCC543E7"] }'

// result
{"jsonrpc":"2.0","result":{"0xeF8305E140ac520225DAf050e2f71d5fBcC543e7":"fabien.eth","0xEF8305E140AC520225DAF050E2F71D5FBCC543E7":"fabien.eth"},"id":null}

### DIFF
--- a/src/addressResolvers/index.ts
+++ b/src/addressResolvers/index.ts
@@ -2,7 +2,14 @@ import * as ensResolver from './ens';
 import * as lensResolver from './lens';
 import * as unstoppableDomainResolver from './unstoppableDomains';
 import cache from './cache';
-import { Address, Handle, normalizeAddresses, normalizeHandles, withoutEmptyValues } from './utils';
+import {
+  Address,
+  Handle,
+  normalizeAddresses,
+  normalizeHandles,
+  withoutEmptyValues,
+  mapOriginalInput
+} from './utils';
 import { timeAddressResolverResponse as timeResponse } from '../helpers/metrics';
 
 const RESOLVERS = [ensResolver, unstoppableDomainResolver, lensResolver];
@@ -48,17 +55,21 @@ async function _call(fnName: string, input: string[], maxInputLength: number) {
 }
 
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
-  return await _call(
+  const result = await _call(
     'lookupAddresses',
     Array.from(new Set(normalizeAddresses(addresses))),
     MAX_LOOKUP_ADDRESSES
   );
+
+  return mapOriginalInput(addresses, result);
 }
 
 export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Address>> {
-  return await _call(
+  const result = await _call(
     'resolveNames',
     Array.from(new Set(normalizeHandles(handles))),
     MAX_RESOLVE_NAMES
   );
+
+  return mapOriginalInput(handles, result);
 }
--- a/src/addressResolvers/utils.ts
+++ b/src/addressResolvers/utils.ts
@@ -35,7 +35,7 @@ export function normalizeAddresses(addresses: Address[]): Address[] {
   return addresses
     .map(a => {
       try {
-        return getAddress(a);
+        return getAddress(a.toLowerCase());
       } catch (e) {}
     })
     .filter(a => a) as Address[];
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by ChaituVR — CHANGES_REQUESTED** [review]
Sorry for delay on this,

It works well with names but not with addresses, try
```
{
    "method": "lookup_addresses",
    "params": [
        "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
        "0xEF8305E140AC520225DAF050E2F71D5FBCC543e7"
    ]
}
```

then

```
{
    "method": "lookup_addresses",
    "params": [
        "0xEF8305E140AC520225DAF050E2F71D5FBCC543e7"
    ]
}
```
Second will not return any response, this is fine?

**Comment by wa0x6e** [comment]
> Sorry for delay on this,
> 
> It works well with names but not with addresses, try
> 
> ```
> {
>     "method": "lookup_addresses",
>     "params": [
>         "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
>         "0xEF8305E140AC520225DAF050E2F71D5FBCC543e7"
>     ]
> }
> ```
> 
> then
> 
> ```
> {
>     "method": "lookup_addresses",
>     "params": [
>         "0xEF8305E140AC520225DAF050E2F71D5FBCC543e7"
>     ]
> }
> ```
> 
> Second will not return any response, this is fine?

No, this is not fine

**Comment by wa0x6e** [comment]
> 0xEF8305E140AC520225DAF050E2F71D5FBCC543e7

I don't get same result:

```bash
curl -X POST http://localhost:3008/ -H "Content-Type: application/json" -d '{"method": "lookup_addresses", "params": ["0xeF8305E140ac520225DAf050e2f71d5fBcC543e7", "0xEF8305E140AC520225DAF050E2F71D5FBCC543e7"] }'

{"jsonrpc":"2.0","result":{"0xeF8305E140ac520225DAf050e2f71d5fBcC543e7":"fabien.eth","0xEF8305E140AC520225DAF050E2F71D5FBCC543e7":"fabien.eth"},"id":null}
```

wa0x6e:
> > Sorry for delay on this,
> > It works well with names but not with addresses, try
> > ```
> > {
> >     "method": "lookup_addresses",
> >     "params": [
> >         "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
> >         "0xEF8305E140AC520225DAF050E2F71D5FBCC543e7"
> >     ]
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > then
> > ```
> > {
> >     "method": "lookup_addresses",
> >     "params": [
> >         "0xEF8305E140AC520225DAF050E2F71D5FBCC543e7"
> >     ]
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Second will not return any response, this is fine?
> 
> No, this is not fine

Should be fixed, please retest

ChaituVR (approved):
tAck

PR #164, opened 2024-01-13
fix: fix ENS avatar resolver when avatar is not HTTP

### LINKED ISSUE #76: Fix NFT profile pictures

### LINKED ISSUE #109: Not getting image for ENS values, it returns blockie icon

wa0x6e:
Fix #76
Fix #109 

Fix missing image on ENS name, when the avatar metadata is not starting by `http`.
This fix uses ENS own metadata resolver as fallback, as pointed out here https://github.com/snapshot-labs/stamp/issues/109#issuecomment-1798361605

### DIFF
--- a/src/resolvers/ens.ts
+++ b/src/resolvers/ens.ts
@@ -10,9 +10,9 @@ export default async function resolve(name: string) {
       return false;
     }
 
-    const url = await ensResolver.getText('avatar');
+    let url = await ensResolver.getText('avatar');
     if (!url || !url.startsWith('http')) {
-      return false;
+      url = `https://metadata.ens.domains/mainnet/avatar/${name}`;
     }
 
     const input = (await axios({ url, responseType: 'arraybuffer' })).data as Buffer;
--- a/src/resolvers/ens.ts
+++ b/src/resolvers/ens.ts
@@ -11,9 +11,7 @@ export default async function resolve(name: string) {
     }
 
     let url = await ensResolver.getText('avatar');
-    if (!url || !url.startsWith('http')) {
-      url = `https://metadata.ens.domains/mainnet/avatar/${name}`;
-    }
+    url = url?.startsWith('http') ? url : `https://metadata.ens.domains/mainnet/avatar/${name}`;
 
     const input = (await axios({ url, responseType: 'arraybuffer' })).data as Buffer;
 


src/resolvers/ens.ts
```diff
@@ -10,9 +10,9 @@ export default async function resolve(name: string) {
       return false;
     }
 
-    const url = await ensResolver.getText('avatar');
+    let url = await ensResolver.getText('avatar');
     if (!url || !url.startsWith('http')) {
-      return false;
+      url = `https://metadata.ens.domains/mainnet/avatar/${name}`;
     }
```
ChaituVR, line 16: Not important but there is a solarlint notice (I use) 
```
Prefer using an optional chain expression instead, as it's more concise and easier to read.sonarlint(typescript:S6582)
```
Maybe we can try this
```
url = url?.startsWith('http') ? url : `https://metadata.ens.domains/mainnet/avatar/${name}`;
```

src/resolvers/ens.ts
```diff
@@ -10,9 +10,9 @@ export default async function resolve(name: string) {
       return false;
     }
 
-    const url = await ensResolver.getText('avatar');
+    let url = await ensResolver.getText('avatar');
     if (!url || !url.startsWith('http')) {
-      return false;
+      url = `https://metadata.ens.domains/mainnet/avatar/${name}`;
```
ChaituVR, line 15: Not sure if there is some rate limit 😄 but we can try

ChaituVR (approved):
tAck

src/resolvers/ens.ts
wa0x6e: We're using this as fallback, only on NFT image, should be fine

src/resolvers/ens.ts
wa0x6e: Done

PR #167, opened 2024-01-28
fix: handle jsonRpcError when silencing contrat call errors

wa0x6e:
When silencing timeout error, also take into account timeout error from brovider JsonRpcError

Also add `ECONNRESET` to list of silenced error (https://snapshot-labs.sentry.io/issues/4495021166/?project=4505506866593792&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=3)


ChaituVR (approved):
utAck

PR #169, opened 2024-02-01
feat: add starknet address/name resolver

### LINKED ISSUE #143: feat: support Starknet.id domains
We should support https://www.starknet.id, this is the main registry for domains on Starknet, I have myself an id checkpoint.stark on this address: `0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d`. We should also separate EVM and Starknet chains, addresses are different and it will be better if we dont try to resolve ENS or Lens for non-EVM address and not trying to resolve Starknet.id for non-Starknet addresses.

wa0x6e:
Fix #143 

Add support for starknet.id domain

## lookup_address

Test with

```bash
curl -X POST localhost:3008/ -H "Content-Type: application/json" -d '{"method": "lookup_addresses", "params": ["0x23d62cEf5622343dD1991E8bb2873D1532C943F2", "0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d"] }'
```

Should accept EVM and non-EVM addresses, and resolve both ENS and starknet.id domains, and return:

```json
{"jsonrpc":"2.0","result":{"0x23d62cEf5622343dD1991E8bb2873D1532C943F2":"altitudefi.eth","0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d":"checkpoint.stark"},"id":null}
```

## resolve_name

Test with

```bash
curl -X POST localhost:3008/ -H "Content-Type: application/json" -d '{"method": "resolve_names", "params": ["fabien.eth", "checkpoint.stark"] }'
```

Should return 

```json
{"jsonrpc":"2.0","result":{"fabien.eth":"0xeF8305E140ac520225DAf050e2f71d5fBcC543e7","checkpoint.stark":"0x07ff6b17f07c4d83236e3fc5f94259a19d1ed41bbcf1822397ea17882e9b038d"},"id":null}
```

## Other changes

- There are some other changes, mostly related to adding support for both evn/non-evm addresses
- Refactoring test (rename unit test to integration), since they depends on external api
- Add e2e tests


src/addressResolvers/ens.ts
```diff
@@ -25,33 +26,44 @@ function normalizeHandles(names: string[]) {
   });
 }
 
+function normalizeAddresses(addresses: Address[]): Address[] {
+  return addresses.filter(isEvmAddress);
+}
+
+function normalizeHandles(names: Handle[]): Handle[] {
+  return normalizeEns(names).filter(h => h);
+}
```
ChaituVR, line 35: why seperate function to just filter? we can filter in `normalizeEns` itself correct?

ChaituVR: In lens resolver, both `map` and `filter` are done in same function `normalizeHandles`, we can do same here

src/addressResolvers/lens.ts
```diff
@@ -28,15 +28,23 @@ async function apiCall(filterName: string, filters: string[]) {
...
-  return handles.map(h =>
-    /^[a-z0-9-_]{5,31}\.lens$/.test(h) ? `lens/${h.replace(/\.lens$/, '')}` : ''
-  );
+  return handles
+    .map(h => (/^[a-z0-9-_]{5,31}\.lens$/.test(h) ? `lens/${h.replace(/\.lens$/, '')}` : ''))
+    .filter(h => h);
 }
 
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
+  const normalizedAddresses = normalizeAddresses(addresses);
```
ChaituVR, line 42: I feel like we can just update `addresses` variable, so that we don't have to update many things. but yea we can keep it this way if needed

src/addressResolvers/unstoppableDomains.ts
```diff
@@ -7,41 +7,52 @@ import {
...
+    const multi = new snapshot.utils.Multicaller(NETWORK, provider, ABI);
+    normalizedAddresses.forEach(address =>
+      multi.call(address, CONTRACT_ADDRESS, 'reverseNameOf', [address])
     );
 
     const names = (await multi.execute()) as Record<Address, Handle>;
 
     return withoutEmptyValues(names);
   } catch (e) {
+    console.log(e);
```
ChaituVR, line 46:
```suggestion
```

src/addressResolvers/utils.ts
```diff
@@ -34,9 +44,13 @@ export function graphQlCall(url, query: string) {
 export function normalizeAddresses(addresses: Address[]): Address[] {
```
ChaituVR, line 44: This has same function name as all resolvers `normalizeAddresses` could we change to something else?

ChaituVR: Actually we are not using this anywhere else. maybe we can just move this to addressResolvers/index.ts

ChaituVR (changes requested):

src/addressResolvers/utils.ts
```diff
@@ -9,12 +9,22 @@ const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.brovider.xyz';
...
   return snapshot.utils.getProvider(network, { broviderUrl });
 }
 
-export function withoutEmptyValues(obj: Record<string, any>) {
-  return Object.fromEntries(Object.entries(obj).filter(([, value]) => value));
+export function withoutEmptyValues(obj: any[] | Record<string, any>) {
+  const data = Array.isArray(obj) ? obj : Object.entries(obj);
+
+  return Object.fromEntries(data.filter(([, value]) => value));
 }
```
ChaituVR, line 28: We now accept a array just for starknet? better to convert and pass it from starknet resolver itself than accepting both array and object

src/addressResolvers/ens.ts
wa0x6e: `normalizeHandles` and `normalizeAddresses` have the same signature everywhere.

`normalizeEns` is just for ENS, and is also used in another case, to detect variable name, without filtering out invalid names. Merging everything was the cause of the previous issue last week, about wrong ENS assosiated to wrong address.

src/addressResolvers/lens.ts
wa0x6e: Updating incoming argument is never a good idea, as array as passed by reference, thus it's also updating it outside this function

src/addressResolvers/utils.ts
wa0x6e: We could move it to index, but also will require importing 2 new dependencies inside `index.ts` inside of just one

```
import { isStarknetAddress } from './utils';
```

ChaituVR (approved):
tAck

PR #170, opened 2024-02-02
refactor: improve axios connection

wa0x6e:
- Improve the axios fetch request by adding timestamp and keepAlive
- Refactor image resolvers by extracting the axios image fetcher to utils.ts

### DIFF
--- a/src/resolvers/space.ts
+++ b/src/resolvers/space.ts
@@ -1,6 +1,7 @@
 import axios from 'axios';
 import { getUrl, resize } from '../utils';
 import { max } from '../constants.json';
+import { fetchHttpImage, axiosDefaultParams } from './utils';
 
 const HUB_URL = process.env.HUB_URL || 'https://hub.snapshot.org';
 
@@ -12,12 +13,13 @@ export default async function resolve(key) {
         method: 'post',
         data: {
           query: `query { space(id: "${key}") { avatar } }`
-        }
+        },
+        ...axiosDefaultParams
       })
     ).data.data.space;
     if (!space || !space.avatar) return false;
     const url = getUrl(space.avatar);
-    const input = (await axios({ url, responseType: 'arraybuffer' })).data as Buffer;
+    const input = await fetchHttpImage(url);
     return await resize(input, max, max);
   } catch (e) {
     return false;
--- a/src/resolvers/utils.ts
+++ b/src/resolvers/utils.ts
@@ -2,21 +2,20 @@ import axios from 'axios';
 import http from 'http';
 import https from 'https';
 
-export async function fetchHttpImage(url: string, options = {}): Promise<Buffer> {
-  return (
-    await axios({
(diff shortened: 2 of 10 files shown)


src/resolvers/space.ts
```diff
@@ -12,12 +13,13 @@ export default async function resolve(key) {
         method: 'post',
         data: {
           query: `query { space(id: "${key}") { avatar } }`
-        }
+        },
+        ...axiosDefaultParams
```
ChaituVR, line 17: I think we should pass `axiosDefaultParams` in `fetchHttpImage` itself, is there any reason to import it here on all resolvers? 🤔

src/resolvers/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    await axios({
+      url,
+      ...{
+        responseType: 'arraybuffer',
+        timeout: 5e3,
+        ...options
+      }
+    })
+  ).data;
+}
```
ChaituVR, line 16: I mean like this
```typescript
export async function fetchHttpImage(url: string, options = {}): Promise<Buffer> {
  return (
    await axios({
      url,
      ...{
        responseType: 'arraybuffer',
        timeout: 5e3,
        ...options,
        ...axiosDefaultParams
      }
    })
  ).data;
}
```

we can also make it optional based on some param passed to `fetchHttpImage`

ChaituVR (changes requested):

src/resolvers/utils.ts
ChaituVR: Only problem is future generations may not remember about `axiosDefaultParams` 😋

wa0x6e: `fetchImage` does not need keepalive, as I don’t think it has benefits when the url is not from same host. We could still uses it there though

src/resolvers/space.ts
wa0x6e: These params are injected in the graphql axios query, and not fetchImate

ChaituVR: ahh makes sense. i thought its the same

src/resolvers/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const axiosDefaultParams = {
+  httpAgent: new http.Agent({ keepAlive: true }),
+  httpsAgent: new https.Agent({ keepAlive: true }),
+  timeout: 5e3
+};
+
+export async function fetchHttpImage(url: string): Promise<Buffer> {
+  return (
+    await axios({
```
ChaituVR, line 13:
```suggestion
  return (
    axios({
```

ChaituVR (approved):
utAck

src/resolvers/utils.ts
wa0x6e: We need this one, else the `.data` will not resolve

PR #171, opened 2024-02-04
refactor: extract cache logic into dedicated class

### LINKED ISSUE #119: Space image cache is wrong
In some cases, even if we pass new `cb` param, the image stored in cache is old avatar

Example https://cdn.stamp.fyi/space/wen🌕.eth?s=96&cb=c178562f8f12066b here is this case `c178562f8f12066b` belong to new image hash but somehow in cache we still have old image

i may delete the cache soon so for future reference:

when i visit link above, I see this 
image

if i change `cb` to something else it works 
image

On UI `cb` is generated with `const hash = sha256(props.space.avatar).slice(0, 16);` 
while `props.space.avatar` value is `ipfs://bafybeiatkvzda3w6xquzv2dkn5gn7f3jrfarxh6bwy5t3lid2hut5qooie`

wa0x6e:

The image resolver cache is tightly coupled inside express controller, making the controller fat and hard to follow. This can also leads to bug (such as cache not clearing properly), as there is no single module for cache handling

## 💊 Fixes / Solution

Extract the cache logic from express controller inside a dedicated `Cache` class.
This allow adding logging and metrics to the cache module, without complicating the express controller.

This is step 1 of a refactoring process, which should also resolve https://github.com/snapshot-labs/stamp/issues/119 after all refactoring PR will be done

## 🚧 Changes

- Move the image resolver cache handling logic inside `Cache` class
- Add logging to the cache module
- Instrument the cache module
- Test the cache module

## 🛠️ Tests

This is an internal refactoring, and does not change the end-user usage/behavior.

- yarn dev
- Go to http://localhost:3008/avatar/0xeF8305E140ac520225DAf050e2f71d5fBcC543e7
- In the yarn console, should see 2 GET and 2 SET cache
- Refresh the page
- You should see only 1 GET cache

By adding other params such as size and cb will SET a new cache, then subsequent call to the same url should only trigger a single GET cache

## Summary by CodeRabbit

- **New Features**
  - Enhanced test output verbosity for clearer insights during testing.
  - Introduced a caching mechanism for image data to streamline storage and retrieval.
  - Added metrics tracking for image resolver cache performance.

- **Bug Fixes**
  - Improved error handling and logging for cache operations.

- **Documentation**
  - Updated descriptions for key functions to clarify their usage and expected return types.

- **Tests**
  - Created integration tests for the new caching functionality to verify image storage and retrieval.


wa0x6e:
Been half a year since this PR opened. May be better to start anew again.

We are now using redis as cache, in addition to aws.

So maybe have a new cache folder, with aws and redis file inside, instead of having addressResolvers/cache.ts and resolvers/cache.ts

bonustrack:
@coderabbitai full review

PR #173, opened 2024-02-05
refactor: move resolver logic inside resolvers

wa0x6e:


ChaituVR:
This still relevant?

wa0x6e:
> This still relevant?

Yes

wa0x6e:
Not even sure of how this work now.

but if resolver is set to something, it should be something valid, and throw when invalid.

src/api.ts
```diff
@@ -70,6 +70,10 @@ router.get(`/:type(${TYPE_CONSTRAINTS})/:id`, async (req, res) => {
     return res.status(500).json({ status: 'error', error: 'failed to load content' });
   }
 
+  if (resolver && !resolverExists(type, resolver)) {
```
wa0x6e, line 73: What if we just use resolve as usual, and just throw an error if it does not exist, instead of using a condition ?

src/resolvers/index.ts
```diff
@@ -26,4 +28,30 @@ export default {
   zapper,
   starknet,
   farcaster
-};
+} as const;
+
+export function resolve(
+  type: string,
```
wa0x6e, line 34: We could use `ResolverType` from utils.

We're always using a known type (a constrain from the express routes), maybe we could remove the fallback on line 40 below

src/resolvers/index.ts
```diff
@@ -26,4 +28,30 @@ export default {
...
+  address: string,
+  network: string,
+  resolvers?: string[]
```
wa0x6e, line 37: This could also use a better type, we have the list of know valid resolvers

src/api.ts
wa0x6e: Validation should be done on the api.ts side, we should consider that everything passed to the resolver to be valid, hence throwing an error would make sense.

I agree there's currently no proper validation on the resolver query params, which should be fine, as it's a debug params, used only for internal purpose.

PR #174, opened 2024-02-11
fix: silence 504 error code from provider

wa0x6e:
Silence 504 (timeout) error code, returned by brovider as JSON-RPC error

### DIFF
--- a/src/addressResolvers/utils.ts
+++ b/src/addressResolvers/utils.ts
@@ -50,8 +50,8 @@ export function isSilencedError(error: any): boolean {
     ['invalid token ID', 'is not supported', 'execution reverted'].some(m =>
       error.message?.includes(m)
     ) ||
-    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET'].some(c =>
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

src/addressResolvers/utils.ts
```diff
@@ -50,8 +50,8 @@ export function isSilencedError(error: any): boolean {
     ['invalid token ID', 'is not supported', 'execution reverted'].some(m =>
       error.message?.includes(m)
     ) ||
-    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET'].some(c =>
-      (error.error?.code || error.code)?.includes(c)
+    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET', 504].some(c =>
+      (error.error?.code || error.error?.status || error.code)?.includes(c)
```
ChaituVR, line 54: what's error code in this case? 🤔 the list is getting bigger though, we may not need to capture errors for network calls maybe?

PR #177, opened 2024-02-26
feat: add support for starknet avatar

### LINKED ISSUE #175: feat: resolve starknet avatar
See https://docs.starknet.id/devs/starknetid-api#domain-to-data

wa0x6e:
Add support for Starknet avatar from domain (see https://docs.starknet.id/devs/starknetid-api#domain-to-data)

Fix #175 

Test with 

- http://localhost:3008/avatar/checkpoint.stark
- http://localhost:3008/avatar/0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d

Should return the same image

Coming later in another PR: resolving Starknet avatar from address

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ cdn.stamp.fyi/**token**/0x0bc529c00C6401aEF6D220BE8C6Ea1667F6Ad93e
 
 ### Identifier
 
-The identifier can be an address (case insensitive), ENS name, CAIP-10, EIP-3770 or DID.
+The identifier can be an address (case insensitive), ENS name, CAIP-10, EIP-3770, DID or Starknet domain.
 
 #### Examples
 
@@ -36,6 +36,7 @@ cdn.stamp.fyi/avatar/**0xeF8305E140ac520225DAf050e2f71d5fBcC543e7**
 cdn.stamp.fyi/avatar/**fabien.eth**  
 cdn.stamp.fyi/token/**eth:0x0bc529c00C6401aEF6D220BE8C6Ea1667F6Ad93e**  
 cdn.stamp.fyi/token/**eip155:1:0x0bc529c00C6401aEF6D220BE8C6Ea1667F6Ad93e**
+cdn.stamp.fyi/token/**checkpoint.stark**
 
 ### Params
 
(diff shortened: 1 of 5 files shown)


bonustrack:
Will it work with address too? Like https://cdn.stamp.fyi/avatar/0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d (checkpoint.stark)

wa0x6e:
Address coming later

wa0x6e:
> Will it work with address too? Like https://cdn.stamp.fyi/avatar/0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d (checkpoint.stark)

Added in this same PR

ChaituVR (changes requested):
if you go here http://localhost:3008/avatar/0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d
It display this image
image

If you go here http://localhost:3008/avatar/x.stark it is not same. is it expected?

ChaituVR:
> Test with

> http://localhost:3008/avatar/checkpoint.stark
> http://localhost:3008/avatar/0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d
> Should return the same image

I see different images

wa0x6e:
> if you go here http://localhost:3008/avatar/0x07FF6B17f07c4d83236E3Fc5f94259A19D1Ed41bBCf1822397EA17882E9b038d It display this image image
> 
> If you go here http://localhost:3008/avatar/x.stark it is not same. is it expected?

Should be not the same, since they're not x.stark do not have an avatar set, so it fallback to blockie

ChaituVR:
Untitled
Maybe we should remove this test case as it can take forever

ChaituVR:
Also adding some timeout for fetch can be good idea

wa0x6e:
> Also adding some timeout for fetch can be good idea

Every outgoing `fetch` call have a timeout of 5seconds, set via `axiosDefaultParams`

wa0x6e:
> Untitled Maybe we should remove this test case as it can take forever

Used another domain/address for the tests, which does not use their slow IPFS resolver.

Issue #180, opened 2024-02-28
feat: add Farcaster avatar as resolver


bonustrack:
Add a new resolver for avatar returning Farcaster avatar image for a specific Ethereum address. Here is an example of resolver that we did for Lens: https://github.com/snapshot-labs/stamp/blob/master/src/resolvers/lens.ts. We should have the same for Farcaster. The function take as input an Ethereum address and return a resized image or false. 

This would allow site like Snapshot and others using Stamp to return Farcaster avatars when available.

Mr94t3z:
okay, i'll try my best to do it.

SamuelLHuber:
The issue I see here is that Wallet Address for any farcaster user can be queried directly on hubs, but the inverse going from wallet address to find a farcaster user name is not available on Hubs.

Hubs are the farcaster protocol nodes.

So you'd need to go through all Farcaster users and find if there is any wallet connected.

There are API providers that do the lookup quickly for you and provide the Wallet Address to Farcaster user query, but they are paid. I am unsure if that would satisfy your requirements for Stamp. adding the API key to .env requirements would solve for it. anyone running would then need a API key there though which is paid.

once one has a farcaster username (fname) or id (FID) the lookup of the profile picture is free using publicly available hubs (farcaster nodes)

bonustrack:
@SamuelLHuber Did you tried the GRPC API "GetVerification" function resolving fid based on an Ethereum address? https://docs.farcaster.xyz/reference/hubble/grpcapi/verification

SamuelLHuber:
> @SamuelLHuber Did you tried the GRPC API "GetVerification" function https://docs.farcaster.xyz/reference/hubble/grpcapi/verification
![Screenshot 2024-03-31 at 00 27 12](https://github.com/snapshot-labs/stamp/assets/40248495/547251c3-b406-4736-b466-b6397da15aaa)


this also requires and FID

bonustrack:
Ok then we can use the paid API

Mr94t3z:
I use this free API to get `avatarUrl`
https://searchcaster.xyz/api/profiles?connected_address=0xcB46Bfb7315eca9ECd42D02C1AE174DA4BBFf291

bonustrack:
Good catch @Mr94t3z

SamuelLHuber:
[...]

Issue #194, opened 2024-03-23
feat: resolve off-chain Farcaster domain


samuveth:
We should support off-chain Farcaster domains (fnames) like `alice.fcast.id`

DOCS: 
https://docs.farcaster.xyz/learn/architecture/ens-names
https://docs.farcaster.xyz/reference/fname/api

developerfred:
@bonustrack
about this: https://github.com/snapshot-labs/stamp/pull/201#issuecomment-2028804541

I will do this issue, but this api  neynar does not support resolvers with address , only if fname is available
https://docs.neynar.com/reference/fname-availability

So I will need to create two query scripts, to associate the address with the fid and after we associate it through the fid to get the corresponding domain, it will work.

As I don't have a Neynar key, I'll use the Docs key

PR #201, opened 2024-03-31
feat(farcaster): add farcaster resolver

developerfred:
#180

I created an api to correlate addresses with FIDs, so we can get images or any other information

https://github.com/developerfred/far-users

This solution is consuming API


bonustrack:
We would prefer to use an API with some guarantee it will be maintained for a long time.

developerfred:
@bonustrack do you have neynar key ?

bonustrack:
@developerfred No, I'll create one, but be aware there is another person selected for the bounty https://github.com/snapshot-labs/stamp/issues/180#issuecomment-2028782668 I don't recommend continue on the PR unless it's not working with the other person.

developerfred:
> @developerfred No, I'll create one, but be aware there is another person selected for the bounty [#180 (comment)](https://github.com/snapshot-labs/stamp/issues/180#issuecomment-2028782668) I don't recommend continue on the PR unless it's not working with the other person.

Its ok, I'll wait. I get the data from the warcast api, and send it to supabase, I have a query for fnames also in this api. ref: https://github.com/snapshot-labs/stamp/issues/194

I'm going to put the scraping code in the api too to make it easier for everyone to understand

```
query Query($address: String!) {
  getFnamesByAddress(address: $address) 
  {    
    name,
    fname,
    fid
  }
  
}```

graphql: https://far-users.vercel.app/api/graphql

**Comment by bonustrack** [comment]
@developerfred If you want to do #194 here is an example with Lens: https://github.com/snapshot-labs/stamp/blob/master/src/addressResolvers/lens.ts it must support querying multiple addresses. We should also use the same API (Neynar) for both resolving avatar and domain.

**Comment by developerfred** [comment]
https://github.com/snapshot-labs/stamp/assets/57037080/eb346429-36b4-40bd-a2be-7db825be3a68

**Review by wa0x6e — CHANGES_REQUESTED** [review]

**Comment by developerfred** [comment]
done ✅ 
```
 PASS  test/integration/resolvers/farcaster.test.ts
  farcaster
    user details by address
      ✓ should return false for invalid or non-existent addresses (135 ms)
      ✓ should return false if user details cannot be fetched (18 ms)
      ✓ should return a Buffer if user details are successfully fetched and the user has a profile picture (286 ms)
      ```

wa0x6e (approved):
tAck

PR #202, opened 2024-03-31
added Farcaster resolver for both custody and verified address

s29papi:
On the protocol level Farcaster doesn't store verified and connected addresses, only custody addresses. Initially my suggestion was on using the Farcaster API, but this only covers custody addresses. Here I use Neynar API to resolve both custody and connected addresses. 
N.B Neynar lets us use their free API keys, but we may need to upgrade to paid, depending on the request being made second.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage
 # Remove some common IDE working directories
 .idea
 .vscode
+try.ts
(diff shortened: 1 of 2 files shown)


bonustrack:
@s29papi Can you tell me what your username on Farcaster? Thanks!

s29papi:
@onchainbets-bot

s29papi:
Hey @bonustrack do you already have someone taking on the eip6963 bounty, I showed interest in it

PR #205, opened 2024-04-03
feat: return followings from address

### LINKED ISSUE #197: feat: API to resolve following list
We should have an API method to receive a list of a specific account's followings, the list would return:

- All the addresses you follow from Farcaster
- ... from EFP https://efp.vercel.app
- ... from Lens

This will allows us later to show for example what your following are voting for.

wa0x6e:
Close #197 

Add `followings` API to return addresses followed by a given address.

Will return following from:

- [x] Lens
- [ ] Farcaster
- [ ] ~~EFP~~

## TODO 

- [ ] Add cache

## 🛠️ Tests

Run 

```
--data '{
    "method": "following",
    "params": "0x70d9b2084d24823789073d3732a327b2c18a7e6f"
}'
```


bonustrack:
We dont need following for multiple addresses, we will always specify 1 address to get list of following.

wa0x6e:
Already the case, using an array just to conform to the rest of the API, array length must be 1

bonustrack:
Why we need to conform to the rest of the API? This is different method of the API we dont need use same argument everywhere

PR #212, opened 2024-04-25
fix: fix starknet avatar

### LINKED ISSUE #186: bug: starknet avatar stopped working
See https://github.com/snapshot-labs/stamp/pull/185#issuecomment-1991201976

Starknet.id API has changed, and does not return same result anymore.

wa0x6e:

Starknet avatar resolver stopped working, due to starknet ID api update, which does not return expected data anymore.

Fix #186 

## 💊 Fixes / Solution

Migrate to use their starknet.js library, which still return the user profile and avatar.

## 🚧 Changes

- Refactor starknet avatar resolver to use starknet.js lib instead of their API
- Add `SN_INFURA_API_KEY` to .env.example
- Add tests for starknet resolver

## 🛠️ Tests

- Run `yarn jest test/integration/resolvers/starknet.test.ts`
- Tests should pass 

You can also test directly with:

- http://localhost:3008/avatar/fricoben.stark
- http://localhost:3008/avatar/0x00C09F8D8CEfb5f094547f0a8B26c88Ca9fD684B4810F3f14C4ef6602FFa7516
- http://localhost:3008/avatar/0x072d4f3fa4661228ed0c9872007fc7e12a581e000fad7b8f3e3e5bf9e6133207


ChaituVR (approved):
tAck

.env.example
```diff
@@ -9,3 +9,4 @@ REDIS_URL=
 SENTRY_DSN=
 SENTRY_TRACE_SAMPLE_RATE=
 NEYNAR_API_KEY=
+SN_INFURA_API_KEY=
```
bonustrack, line 12: Can we rename to this:
```suggestion
INFURA_API_KEY=
```

bonustrack: I've added already the key in Stamp server (with and without SN_ prefix)

wa0x6e: Can you also add it to github so tests pass?

bonustrack: Should be ok now
image

PR #217, opened 2024-04-29
fix: ignore non-valid starknet address from starknet resolver

### LINKED ISSUE #216: bug: fallback image returning default avatar for starknet
After #212, the fallback for addresses `0x...` always returns the default Starknet avatar instead of the jazzicon or blockie as was returned in the past.

I've also tried modifying our query with the `resolver` query param, but no luck so far.

`https://cdn.stamp.fyi/avatar/${address}?resolver=ens&fb=jazzicon&cb=40`

image

wa0x6e:
Fix #216 

Fix an issue where all starknet was resolving on ETH address, thus making all addresses without images using the starknet fallback image

### DIFF
--- a/src/resolvers/starknet.ts
+++ b/src/resolvers/starknet.ts
@@ -14,6 +14,12 @@ function isStarknetDomain(domain: string): boolean {
   return domain.endsWith('.stark');
 }
 
+function normalizeAddress(address: string): string {
+  if (!address.match(/^(0x)?[0-9a-fA-F]{64}$/)) throw new Error('Invalid starknet address');
+
+  return address;
+}
+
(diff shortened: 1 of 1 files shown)


src/resolvers/starknet.ts
```diff
@@ -14,6 +14,12 @@ function isStarknetDomain(domain: string): boolean {
   return domain.endsWith('.stark');
 }
 
+function normalizeAddress(address: string): string {
+  if (!address.match(/^(0x)?[0-9a-fA-F]{64}$/)) throw new Error('Invalid starknet address');
```
ChaituVR, line 18: I think this check should be inside `isStarknetDomain`

ChaituVR (approved):
utAck, added a comment for small improvement

PR #218, opened 2024-04-29
fix: fix ipfs gateway always timing out on test env

wa0x6e:
In test env. use 4everland instead of cloudflare as ipfs gateway, as cloudflare is always timing out, and failing tests

After this fix, `snapshot` and `space` resolvers should not fail in test anymore


ChaituVR (approved):
utAck

PR #219, opened 2024-04-29
chore: capture more http 504 errors

wa0x6e:
Improve the filter capturing http 504 error.

This will let us capture 504 error returned by brovider


ChaituVR (approved):
utAck

PR #220, opened 2024-04-29
fix: fix `resolver` query params being ignored

wa0x6e:
Fix issue where the `?resolver=lens` query params is being ignored


ChaituVR (approved):
utAck

PR #228, opened 2024-05-28
fix: fix case sensitive graphql filter

wa0x6e:
Fix issue where some thegraph API are expecting checksum formatted addresses.

This PR will fix the issue by always passing both lowercase and checksumed address


bonustrack (approved):
utACK

PR #229, opened 2024-05-29
fix: fix address formatter for starknet address

wa0x6e:
This PR skip the address formatting when it is a starknet address, and also add/fix more tests


src/resolvers/space-sx.ts
```diff
@@ -15,13 +16,18 @@ const SUBGRAPH_URLS = [
...
+  }
+
   const data = await axios({
     url,
     method: 'POST',
     data: {
       query: `
         query {
-          spaces(where: { id_in: ["${key}", "${getAddress(key)}"] }) {
+          spaces(where: { id_in: ["${id.join('", "')}"] }) {
```
ChaituVR, line 30: maybe this is much better
```suggestion
          spaces(where: { id_in: [${id.map(item => `"${item}"`).join(', ')}] }) {
```

ChaituVR (changes requested):

ChaituVR:
Looks like test cases are still failing some how 🤔

wa0x6e:
> Looks like test cases are still failing some how 🤔

Failing tests are unrelated, they're now failing due to something wrong with the starknet library itself, and not due to tests themselves, issue has been open https://github.com/snapshot-labs/stamp/issues/230

ChaituVR (approved):
utAck

PR #231, opened 2024-05-30
feat: add `lookupDomains` method to return an address owned ENS domains

wa0x6e:
This PR add a new API method `lookupDomains` to return an address owned ENS domains

The goal is to move https://github.com/snapshot-labs/snapshot/blob/1b1c37021a53470512c840a69d6b31c51068b912/src/composables/useEns.ts#L8 here to stamp, so that we only have one source of truth for a controller owned domains.

After this PR is merged

- we should refactor snapshot UI to fetch list of owned ENS from stamp (we keep the invalid domains validation on UI side)
- Add a new `controller` filter to Space on the hub, so that SX could return all spaces controlled by an address (like on SX API)

Only 2 files are changed (api.ts and lookupDomains.ts), as well as tests, the rest are refactoring because we moved a function

### Test

Run 

```shell
curl -X POST http://localhost:3008/ -H "Content-Type: application/json" -d '{"method": "lookup_domains", "params": "0x24F15402C6Bb870554489b2fd2049A85d75B982f" }'
```

It should return 

```json
{"jsonrpc":"2.0","result":["chaitu.eth","validation.snapspace.eth","snapspace.eth","2.snapspace.eth","1.snapspace.eth","thanku.eth","3.snapspace.eth","4.snapspace.eth"],"id":null}
```

### TODO 

- [x] Add tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@self.id/core": "^0.3.0",
     "@snapshot-labs/snapshot-metrics": "^1.4.1",
     "@snapshot-labs/snapshot-sentry": "^1.5.5",
-    "@snapshot-labs/snapshot.js": "^0.11.27",
+    "@snapshot-labs/snapshot.js": "^0.11.28",
     "@unstoppabledomains/resolution": "^9.2.2",
     "axios": "^0.25.0",
     "canvas": "^2.9.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2312,10 +2312,10 @@
   dependencies:
     "@sentry/node" "^7.81.1"
 
-"@snapshot-labs/snapshot.js@^0.11.27":
-  version "0.11.27"
-  resolved "https://registry.yarnpkg.com/@snapshot-labs/snapshot.js/-/snapshot.js-0.11.27.tgz#f341f810e61d6dce72f00ddd1a57bde338b9a6b3"
-  integrity sha512-oCKY0thlLeSh2fYrVzPCzHb9HHyfRqLD55IGBrCgUUORJP78w3TvlscXwP0LgdNPuN3pNaDNwLGKUrx4GayB/w==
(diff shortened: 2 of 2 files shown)


src/lookupDomains.ts
```diff
@@ -0,0 +1,85 @@
+import { capture } from '@snapshot-labs/snapshot-sentry';
+import { FetchError, isSilencedError } from './addressResolvers/utils';
+import { Address, graphQlCall } from './utils';
+import { isAddress } from '@ethersproject/address';
+
+const ENS_GRAPHQL_URL = 'https://api.thegraph.com/subgraphs/name/ensdomains/ens';
```
ChaituVR, line 6: Could we allow network as input? to get domains owned on sepholia for testnet spaces? not needed for now?

ChaituVR (changes requested):

src/lookupDomains.ts
wa0x6e: Good idea

src/lookupDomains.ts
```diff
@@ -0,0 +1,97 @@
...
+    ...domain,
+    name: labelName ? domain.name.replace(`[${hash}]`, labelName) : domain.name
+  };
+}
+
+export default async function lookupDomains(
+  address: Address,
+  chainId = DEFAULT_CHAIN_ID
+): Promise<Address[]> {
+  if (!isAddress(address) || !ENS_GRAPHQL_URL[chainId]) return [];
```
ChaituVR, line 51: Maybe better to return error?

ChaituVR (approved):
tAck

src/lookupDomains.ts
wa0x6e: Maybe, but all the other methods do not return any errors, and just return `[]` instead.

ChaituVR: i see.. ok ok 👍

PR #240, opened 2024-06-13
feat: add support for user cover

wa0x6e:
Related to the new profile page introduced in https://github.com/snapshot-labs/sx-monorepo/pull/394

An user has now a `cover` property, in addition to `avatar`, similar to the sx `Space` object.

This PR will add a new `user-cover` type when requesting image, allowing to fetch an user cover image with 

```
http://localhost:3008/user-cover/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3
```

The `snapshot.ts` has been refactored to be similar to `space-sx.ts`


ChaituVR (changes requested):
Weirdly i get avatar instead of a cover image 

image

ChaituVR (approved):
tAck

PR #242, opened 2024-06-17
fix: allow larger max size for all cover types

wa0x6e:
This PR allow a larger max size for all image type including `-cover`.

Else, the newly introduced `user-cover` type is capped to only 500, but should behave like `space-cover-xxx`


ChaituVR (approved):
utAck

Issue #243, opened 2024-06-18
chore: cover image are too heavy


bonustrack:
I can see cover image taking few seconds to load. Compared to Twitter cover image (having same size) it's about 4x more heavy on Stamp. Is there a way we can reduce the size without losing much on quality?

<img width="287" alt="image" src="https://github.com/snapshot-labs/stamp/assets/16245250/f7c3e508-7f37-4da5-936d-b48f11f74f60">

wa0x6e:
We can test with nearlossless compression instead

bonustrack:
Don't we do that already? But yes we can try

bonustrack:
Currently with my internet it takes about 4sec for the cover image to appear

Sekhmet:
`nearlossless` doesn't make great difference (142kB vs 166kB). Making it lossy with quality 80% gives us 17.7kB.

PR #247, opened 2024-06-19
feat: use lossy webp encoding (quality 80).

Sekhmet:
Using reduced quality greatly reduces image size (images are 90% smaller). Quality is reduced slightly, but not really noticable. You can try what quality 80 for WebP looks like on https://squoosh.app/

Closes: https://github.com/snapshot-labs/stamp/issues/243

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -36,7 +36,7 @@ export function sha256(str) {
 export async function resize(input, w, h) {
   return sharp(input)
     .resize(w, h)
-    .webp({ lossless: true })
+    .webp()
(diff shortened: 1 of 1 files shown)


wa0x6e:
Did you try with just `nearlossless` ?

wa0x6e (approved):
utAck

Sekhmet:
> Did you try with just nearlossless ?

`nearlossless` doesn't make great difference (142kB vs 166kB). Making it lossy with quality 80% gives us 17.7kB.

PR #251, opened 2024-07-05
fix: use address formatter from SDK

wa0x6e:
Depends on https://github.com/snapshot-labs/snapshot.js/pull/1032 

This PR will use the address formatter from snapshot.js to prepare address before sending to graphql

This address formatter is more robust, and use by sequencer and the hub

### DIFF
--- a/src/resolvers/snapshot.ts
+++ b/src/resolvers/snapshot.ts
@@ -1,8 +1,8 @@
+import snapshot from '@snapshot-labs/snapshot.js';
 import axios from 'axios';
 import { getUrl, resize } from '../utils';
 import { max } from '../constants.json';
 import { fetchHttpImage, axiosDefaultParams } from './utils';
-import { getChecksummedAddress } from '../utils';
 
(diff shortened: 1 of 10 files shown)


src/resolvers/snapshot.ts
```diff
@@ -1,8 +1,8 @@
+import snapshot from '@snapshot-labs/snapshot.js';
 import axios from 'axios';
 import { getUrl, resize } from '../utils';
 import { max } from '../constants.json';
 import { fetchHttpImage, axiosDefaultParams } from './utils';
-import { getChecksummedAddress } from '../utils';
```
ChaituVR, line 5: We are not using getChecksummedAddress anywhere, maybe we can delete it from utils

ChaituVR (changes requested):

src/utils.ts
ChaituVR, line 1: Untitled 5

ChaituVR (changes requested):

wa0x6e:
Closing, will let hub decide how to format the address

PR #252, opened 2024-07-05
fix: fix starknet avatar not working

### LINKED ISSUE #230: bug: starknet avatar not resolving

wa0x6e:
Fix #230 

Fix issue where starknet avatar where not working.

The call to the contract was returning an exception, upgrading the lib to the latest version fixed the issue


ChaituVR (approved):
utAck

PR #259, opened 2024-07-19
fix: ignore starknet default avatar image

### LINKED ISSUE #257: feat: ignore Starknet.id default image
Starknet avatars working well when you setup one with Starknet.id but when there isn't any avatar available Starknet.id return the default avatar. We should detect this default avatar and ignore it to show instead our fallback avatar (with blockies)

image

wa0x6e:
Fix #257 

Ignore starknet default image on avatar resolver

### Test 

- Go to https://snapshot.box/#/sn:0x07c251045154318a2376a3bb65be47d3c90df1740d8e35c9b9d943aa3f240e50/proposal/1/votes
- Select one address with default avatar
- Go to http://localhost:3008/avatar/ADDRESS with the copied address
- It should return blockicon avatar
- Do the same with an address with a non-default avatar
- It should return the avatar

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@self.id/core": "^0.3.0",
     "@snapshot-labs/snapshot-metrics": "^1.4.1",
     "@snapshot-labs/snapshot-sentry": "^1.5.5",
-    "@snapshot-labs/snapshot.js": "^0.11.41",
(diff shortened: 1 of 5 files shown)


ChaituVR (approved):
utAck

PR #261, opened 2024-07-19
fix: remove unecessary address formatting

wa0x6e:

We format user address before sending them to the hub, which is not needed anymore, as we've added the formatting directly on the hub side.

## 💊 Fixes / Solution

Remove address formatting, delegating the logic to the hub

## Test 

Both urls should return the same image

- http://localhost:3008/avatar/0x91fd2c8d24767db4ece7069aa27832ffaf8590f3?resolver=snapshot
- http://localhost:3008/avatar/0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3?resolver=snapshot


ChaituVR (approved):
utAck

PR #262, opened 2024-07-19
feat: add snapshot to addressResolvers

wa0x6e:
Fix #262 

Lookup snapshot user's name from addresses

### 🛠️ Tests
 
Run 

```shell
--data '{
    "method": "lookup_addresses",
    "params": [
        "0x91fd2c8d24767db4Ece7069AA27832ffaf8590f3", "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7", "0x24F15402C6Bb870554489b2fd2049A85d75B982f", "0x76A6D08b82034b397E7e09dAe4377C18F132BbB8"
    ]
}'
```

It should return a list of addresses, with the username from snapshot for the first 3 addresses, and ENS name for the last one

### Other changes

- Add `address` and `name` to the cache clearing endpoint, so that calling `http://localhost:3008/clear/address/0x91fd2c8d24767db4Ece7069AA27832ffaf8590F3` will clear the address cache for the given address. This will be called by sequencer on user's name change, via https://github.com/snapshot-labs/snapshot-sequencer/pull/410

### DIFF
--- a/src/addressResolvers/index.ts
+++ b/src/addressResolvers/index.ts
@@ -2,6 +2,7 @@ import * as ensResolver from './ens';
 import * as lensResolver from './lens';
 import * as unstoppableDomainResolver from './unstoppableDomains';
 import * as starknetResolver from './starknet';
+import * as snapshotResolver from './snapshot';
 import cache from './cache';
 import {
   normalizeAddresses,
@@ -12,7 +13,13 @@ import {
 import { Address, Handle } from '../utils';
 import { timeAddressResolverResponse as timeResponse } from '../helpers/metrics';
 
-const RESOLVERS = [ensResolver, unstoppableDomainResolver, lensResolver, starknetResolver];
+const RESOLVERS = [
+  snapshotResolver,
+  ensResolver,
+  unstoppableDomainResolver,
+  lensResolver,
+  starknetResolver
+];
 const MAX_LOOKUP_ADDRESSES = 50;
 const MAX_RESOLVE_NAMES = 5;
 
--- a/src/addressResolvers/index.ts
+++ b/src/addressResolvers/index.ts
@@ -3,7 +3,7 @@ import * as lensResolver from './lens';
 import * as unstoppableDomainResolver from './unstoppableDomains';
 import * as starknetResolver from './starknet';
 import * as snapshotResolver from './snapshot';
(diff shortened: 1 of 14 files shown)


wa0x6e:
mhhh. wait , test are failing

wa0x6e:
Test passing now, just a temp thing

src/addressResolvers/index.ts
```diff
@@ -12,7 +13,13 @@ import {
 import { Address, Handle } from '../utils';
 import { timeAddressResolverResponse as timeResponse } from '../helpers/metrics';
 
-const RESOLVERS = [ensResolver, unstoppableDomainResolver, lensResolver, starknetResolver];
+const RESOLVERS = [
+  snapshotResolver,
```
ChaituVR, line 17: What happens when the user edits his profile? 🤔 we will still have it in our cache right

wa0x6e: Good point … what to do

bonustrack: Ideally the sequencer would trigger stamp for clear cache of that specific user who updated his profile

wa0x6e: Cache clearing for `address` added. will make a PR on sequencer side to call it on user profile update

src/api.ts
```diff
@@ -1,15 +1,15 @@
...
 import resolvers from './resolvers';
 import constants from './constants.json';
 import { rpcError, rpcSuccess } from './helpers/utils';
-import { lookupAddresses, resolveNames } from './addressResolvers';
+import { lookupAddresses, resolveNames, clearCache } from './addressResolvers';
 import lookupDomains from './lookupDomains';
 
 const router = express.Router();
-const TYPE_CONSTRAINTS = Object.keys(constants.resolvers).join('|');
+const TYPE_CONSTRAINTS = [...Object.keys(constants.resolvers), 'address', 'name'].join('|');
```
ChaituVR, line 12: I don't remember where we use `name` resolver 🤔

ChaituVR (approved):
tAck

src/api.ts
wa0x6e: It's for resolveName, when passing a domain name (like chaitu.eth), and return an address.

It's there, but not also sure where we use it

PR #264, opened 2024-07-28
refactor: export typed client instance from redis helper

mktcode:
This makes sure the redis client isn't typed as _any_, so your IDE can show you its internals (which I was missing when trying to build something).

Changes in the other two files are due to the exposed possible null pointer exception. (accessing prop of undefined)

### DIFF
--- a/src/addressResolvers/cache.ts
+++ b/src/addressResolvers/cache.ts
@@ -46,6 +46,8 @@ export default async function cache(input: string[], callback) {
 }
 
 export async function clear(input: string): Promise<boolean> {
+  // TODO: When redis is not available, it should probably throw instead of returning false
+  // causing the api the return "failed to clear cache" instead of "not found"
   if (!redis) return false;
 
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
utAck

test/integration/addressResolvers/index.test.ts
```diff
@@ -3,6 +3,12 @@ import { getCache, setCache } from '../../../src/addressResolvers/cache';
 import redis from '../../../src/helpers/redis';
 import randomAddresses from '../../fixtures/addresses';
 
+function purge() {
+  if (!redis) return;
+
+  return redis.flushDb();
+}
+
```
wa0x6e, line 11: One of the thing that is long due is to have a different setup files for integration/unit/e2e tests with jest, so we could move this to the setup files.

This will avoid to boot the redis instance on unit tests, where it is not needed

bonustrack (approved):

PR #267, opened 2024-07-29
chore: add husky and lint-staged for pre-commit linting

mktcode:
Adding this pre-commit hook to prevent accidentally committing unformatted code. Applies only to staged changes.

(Unfortunately when using yarn, husky requires the addition dependency [`pinst`](https://www.npmjs.com/package/pinst/v/2.1.0))

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx lint-staged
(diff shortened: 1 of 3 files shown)


bonustrack:
Do we need this? The other approach is just running GitHub actions to tell you if there is a lint issue, we had husky previously in some of our repo, one thing that bother me is that it's taking much longer to commit changes, was very slow for some codebase

PR #268, opened 2024-07-29
chore: update lint and test workflows to target node 18

mktcode:
Tried running the workflows on node 18. Works nicely. @bonustrack anything blocking this? What version does production run on?

Cc: https://github.com/stamp-labs/stamp/issues/206 https://github.com/stamp-labs/stamp/issues/48 https://github.com/stamp-labs/stamp/pull/267 https://github.com/stamp-labs/stamp/pull/278


ChaituVR:
Production is using v16.20.2 but i think it should be fine

PR #270, opened 2024-07-30
refactor: use graphql helper for all graphql calls

wa0x6e:
This refactor is using the `graphQlCall` helper for all graphql calls

This will enable us in #265 to put the api key handling only inside this function

### DIFF
--- a/src/resolvers/space-sx.ts
+++ b/src/resolvers/space-sx.ts
@@ -1,8 +1,7 @@
-import axios from 'axios';
 import { getAddress } from '@ethersproject/address';
-import { getUrl, resize } from '../utils';
+import { getUrl, graphQlCall, resize } from '../utils';
 import { max } from '../constants.json';
-import { fetchHttpImage, axiosDefaultParams } from './utils';
+import { fetchHttpImage } from './utils';
 import { isStarknetAddress } from '../addressResolvers/utils';
 
 const SUBGRAPH_URLS = [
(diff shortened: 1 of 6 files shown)


ChaituVR (approved):
utAck

src/resolvers/space-sx.ts
```diff
@@ -21,25 +20,22 @@ async function getSpaceProperty(key: string, url: string, property: 'avatar' | '
...
-        }`
-    },
-    ...axiosDefaultParams
-  });
-
-  const result = data.data?.data?.spaces
-    ?.map(space => space.metadata?.[property])
-    .filter(Boolean)[0];
+      data: { spaces }
+    }
```
ChaituVR, line 26: should be fine in case of an error?

wa0x6e: Mhhh it’s been tested only in case of no results

ChaituVR: For now, i think we should go back to the previous way `data.data?.data?.spaces` and we should be good to merge

ChaituVR: > Also, constructing the queries like this is a bit error prone:

What kind of errors you are referring to?

ChaituVR:
> @wa0x6e @ChaituVR @bonustrack In favor of moving forward, I will create an issue for the API response validation and GraphQL variables. Let's make this here just work. It does, right?

Yep let's go with this for now. it looks like API response validation doesn't work before as well

Issue #276, opened 2024-08-03
Tests should run against prod build


mktcode:
https://github.com/stamp-labs/stamp/blob/3c00957b63d4e061227dbe77628aa6de80668be5/package.json#L12

`yarn dev` starts `nodemon` and compiles on the fly, using `ts-node`. Instead `build` / `start` should be used, to .

In this case, we test without type-checking, if I'm not missing something, and on deployment we could still run into compile errors.

[This workflow](https://github.com/snapshot-labs/actions/blob/main/.github/workflows/test.yml) has been [moved](https://github.com/stamp-labs/stamp/pull/274) to this repo and needs adjustment, to add the build step. I guess you could also chain it together in the `yarn test` command.

wa0x6e:
The risk this raised, compile errors reaching deploy because tests ran through ts-node without type-checking, is covered now: the Lint workflow runs `yarn typecheck` (`tsc --noEmit`, same `tsconfig.json` as `yarn build`) on every push, and tests run in-process under ts-jest, which type-checks. The `yarn dev` test server was removed in [87adab2](https://github.com/snapshot-labs/stamp/commit/87adab223680440eed4430a710209745837aa9c7) (#396). Tests still don't execute `dist/`; reopen if that is wanted for its own sake.

PR #278, opened 2024-08-04
chore: update start:test script to include build

mktcode:
Resolves: #276

Since it's also required to run it on the correct Node.js version, I also added the ["engines"](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) key in package.json. I chose 18.20.4 because that is the version used when I run `nvm use 18` and everything works. I know it breaks on the latest node version 22 and 21 was working still but I didn't track it down further because compatibility with later node.js version should be achived instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,5 +56,8 @@
     "prettier": "^1.19.1",
     "start-server-and-test": "^2.0.3",
     "ts-jest": "^28.0.4"
+  },
+  "engines": {
+    "node": ">=18.20.4 <19"
   }
 }


wa0x6e:
I understand that it running the test on the compiled js is better.

Could we do it only on the github CI, and avoid doing it when developing locally ? It'll be a pain to wait for compilation before each unit tests

package.json
```diff
@@ -56,5 +56,8 @@
     "prettier": "^1.19.1",
     "start-server-and-test": "^2.0.3",
     "ts-jest": "^28.0.4"
+  },
+  "engines": {
+    "node": ">=18.20.4 <19"
   }
```
ChaituVR, line 62: Tests use node version 16 by default, https://github.com/snapshot-labs/actions/blob/main/.github/workflows/test.yml

if we want to change, we should change test.yml and lint.yml too

ChaituVR (changes requested):

PR #282, opened 2024-08-05
feat: tokenlists resolver

### LINKED ISSUE #24: feat: add Tokenlists resolver for tokens images
We should add a resolver for Tokenlists for tokens images, and use the Uniswap + Zerion + Coingecko lists for now see https://tokenlists.org/token-list?url=tokenlist.zerion.eth

mktcode:
Closes: #24

### Tests

The (e2e) tests check previously generated fingerprints of the returned images. Of courtse that will make tests fail, when one of these images changes. The images we test should be chosen accordingly. Correct invalidation of changed/updated images should probably be a test of its own.

The tests could be far more comprehensive but I didn't want to cause too much refactoring. The fingerprinting should help to test more things already and reduce manual review.

### Implementation

The tokenlists data is aggregated into a map, containing all found image URLs for a `<chainId>-<address>` pair.

https://github.com/stamp-labs/stamp/blob/0ee529ec4b2ebf1cbb24c921ee6ad3ea024d8994/src/helpers/tokenlists.ts#L9

The URLs are sorted based on keywords they contain (in their path).

https://github.com/stamp-labs/stamp/blob/0ee529ec4b2ebf1cbb24c921ee6ad3ea024d8994/src/helpers/tokenlists.ts#L85-L126

Before that, known URL patterns related to image size, are replaced. Like for coingecko, we know we can replace _small_ and _thumb_ with _large_ to get a larger image.

https://github.com/stamp-labs/stamp/blob/c0f8a9c2503971b999c86da45a7d90a0d4d5876e/src/helpers/tokenlists.ts#L71-L85

When the resolver runs for the first time, it waits for the tokenlists data to be fetched and aggregated.

https://github.com/stamp-labs/stamp/blob/c0f8a9c2503971b999c86da45a7d90a0d4d5876e/src/resolvers/tokenlists.ts#L6-L10
https://github.com/stamp-labs/stamp/blob/51bff9d637a1771fcf532fff1d67706f46c2e802/src/helpers/tokenlists.ts#L135-L140

When the cache expires, an update is triggered by the first incoming request. This request waits for the update to complete. Other concurrent requests will use the existing list, which may be empty after a server restart. While they could wait for the first request to finish updating, I'm still struggling to understand the rationale behind this approach.

I'm curious about the reasons for not running updates at regular intervals, starting with the server initialization. Additionally, I'm wondering why it's necessary for the resolver to determine the TTL starting timestamp.
Personally, I might consider a different approach, such as running a daily or hourly GitHub workflow to update the data in the Redis cache. This way, the application wouldn't need to handle these updates directly. The dependabot-like approach I suggested earlier is not even necessary.

Btw, I didn't have the time to look into sentry integration yet.

@bonustrack I'm working on another project now, so, like I said, I won't have a lot of time anymore to work on this one. But I will check back in a couple of weeks/months. Let me know when release day get's closer.


bonustrack:
Shouldn't the change on cache be in another PR? I don't see how it's related to tokenlists

src/api.ts
```diff
@@ -122,14 +122,18 @@ router.get(`/:type(${TYPE_CONSTRAINTS})/:id`, async (req, res) => {
...
       return res.send(resizedImage);
     }
   }
 
   // Resize and return image
   const resizedImage = await resize(baseImage, w, h);
-  setHeader(res);
+  const extraHeaders = {};
+  if (resolver) extraHeaders[`x-stamp-${type}-resolver`] = resolver;
+  setHeader(res, 'LONG_CACHE', extraHeaders);
```
bonustrack, line 136: I mean these changes, how is this related to tokenlists?

bonustrack:
There should be a way to change the URL like: 
https://assets.coingecko.com/coins/images/35810/thumb/enjoy_tech.jpeg?1709871790
To the large version of the image like this:
https://assets.coingecko.com/coins/images/35810/large/enjoy_tech.jpeg?1709871790

bonustrack:
@mktcode Thank you, code looks good, I've tried the PR locally, for some reason I can't resolve image for some token that are in the tokenlists, here is what I've tried:

http://localhost:3008/token/oeth:0xe0BB0D3DE8c10976511e5030cA403dBf4c25165B?s=100

This token is available on the Optimism list with an image here (2nd position): https://static.optimism.io/optimism.tokenlist.json

Any idea what's wrong?

Also I've disabled CodeRabbit

bonustrack:
What do you mean by missing prefix? The URLs looks the same to me

bonustrack:
@mktcode We should add a test case with a token and network prefix, we could have network prefix with both chain id and network short name like `eip155:1:0x123` and `arb1:1:0x123`. Token URL with network prefix should already work with others resolvers.

src/resolvers/tokenlists.ts
```diff
@@ -0,0 +1,16 @@
+import { resize } from '../utils';
+import { max } from '../constants.json';
+import { fetchHttpImage } from './utils';
+import { findImageUrl, updateExpiredAggregatedTokenList } from '../helpers/tokenlists';
+
+export default async function resolve(address: string, chainId: string) {
+  try {
+    await updateExpiredAggregatedTokenList();
```
wa0x6e, line 8: If we remove the `await` here:

- the first request triggering the token list refresh will always return immediately with nothing, with the list refreshing async
- a request waiting for list refresh will return immediately with current list, instead of waiting for new list, which is refreshing async

test/e2e/tokenlists.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  it('returns correct image for ERC3770 address', async () => {
+    const response = await getImageResponse(ERC3770_ADDRESS);
+
+    const image = sharp(response.data);
+    const imageBuffer = await image.raw().toBuffer();
+
+    const fingerprint = getImageFingerprint(imageBuffer.toString('hex'));
+    const expectedFingerprint = 'ac601f072065d4d03e6ef906c1dc3074d7ad52b9c715d0db6941ec89bf2073a1';
```
wa0x6e, line 47: As mentioned here https://github.com/stamp-labs/stamp/pull/282#issuecomment-2292623553 this is not the expected image

wa0x6e: This is not working due to something unrelated to this PR: the shortname `oeth` is not supported:

https://github.com/stamp-labs/stamp/blob/3b50d97093d4f038003cdb8a4abb401e311bc8a0/src/utils.ts#L50-L58

wa0x6e (changes requested):

wa0x6e:
Covered by the new defilama resolver

PR #294, opened 2024-09-12
fix: update sx subgraph url

wa0x6e:
Update the SX subgraph to the new `.box` domains, instead of old `x.xyz` domains


ChaituVR (approved):
utAck

PR #295, opened 2024-09-12
fix: update brovider default url

wa0x6e:
Update brovider url to use the one from documentation, the .xyz one is deprecated and should not be used anymore


ChaituVR (approved):
utAck

PR #312, opened 2024-10-09
fix: fix deprecated sepolia subgraph url

wa0x6e:
Update ENS sepolia subgraph url

### DIFF
--- a/src/constants.json
+++ b/src/constants.json
@@ -14,7 +14,7 @@
   },
   "ensSubgraph": {
     "1": "https://subgrapher.snapshot.org/subgraph/arbitrum/5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH",
-    "11155111": "https://api.studio.thegraph.com/proxy/49574/enssepolia/version/latest"
+    "11155111": "https://subgrapher.snapshot.org/subgraph/arbitrum/DmMXLtMZnGbQXASJ7p1jfzLUbBYnYUD9zNBTxpkjHYXV"
   },
   "defaultOffchainNetwork" : "s",
   "offchainNetworks": ["s", "s-tn"]
--- a/src/lookupDomains.ts
+++ b/src/lookupDomains.ts
@@ -81,6 +81,7 @@ export default async function lookupDomains(
       ) || []
     );
   } catch (e) {
+    console.log(e);
     if (!isSilencedError(e)) {
       capture(e, { input: { address } });
     }


src/constants.json
```diff
@@ -14,7 +14,7 @@
   },
   "ensSubgraph": {
     "1": "https://subgrapher.snapshot.org/subgraph/arbitrum/5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH",
-    "11155111": "https://api.studio.thegraph.com/proxy/49574/enssepolia/version/latest"
+    "11155111": "https://subgrapher.snapshot.org/subgraph/arbitrum/DmMXLtMZnGbQXASJ7p1jfzLUbBYnYUD9zNBTxpkjHYXV"
```
bonustrack, line 17: Looks like the right "official" endpoint is `https://api.studio.thegraph.com/query/49574/enssepolia/version/latest`

ChaituVR: It is rate limited though

ChaituVR: Yes yes, It's a tricky situation 

Official and Ratelimited
`https://api.studio.thegraph.com/query/49574/enssepolia/version/latest`

Not Official and not Ratelimited
`https://subgrapher.snapshot.org/subgraph/arbitrum/DmMXLtMZnGbQXASJ7p1jfzLUbBYnYUD9zNBTxpkjHYXV`

ChaituVR:
> Can't we use our API key on `https://api.studio.thegraph.com/query/49574/enssepolia/version/latest`?

No, because it is not published, if we know the subgraph ID, it should work, I've asked about it on ENS discord but no one replied

bonustrack:
Maybe ask TheGraph directly

src/constants.json
wa0x6e: Since we were using the official one before, does it mean it was also rate limited ?

ChaituVR: Yes it was

ChaituVR: Thegraph team confirmed that both are same subgraphs ✅ 

we can go with `https://subgrapher.snapshot.org/subgraph/arbitrum/DmMXLtMZnGbQXASJ7p1jfzLUbBYnYUD9zNBTxpkjHYXV`



> The subgraph ID of https://api.studio.thegraph.com/query/49574/enssepolia/version/latest is indeed DmMXLtMZnGbQXASJ7p1jfzLUbBYnYUD9zNBTxpkjHYXV

> You can see that the deployment ID it returns when you query the Studio endpoint above with the query below is the same as on 
https://thegraph.com/explorer/subgraphs/DmMXLtMZnGbQXASJ7p1jfzLUbBYnYUD9zNBTxpkjHYXV?view=About&chain=arbitrum-one

> Query:
```
{
 _meta{deployment} 
}
```

ChaituVR (approved):
utAck

PR #314, opened 2024-10-12
fix: fix legacy snapshot resolver not working

wa0x6e:
This PR fix an issue where the legacy url for the `space/ADDRESS` endpoint (without network prefix) was not working as intended, due to an issue when setting the missing network prefix 

This result in those 2 url not returning the same result

- http://localhost:3008/space/s:test.wa0x6e.eth?resolver=space
- http://localhost:3008/space/test.wa0x6e.eth?resolver=space


ChaituVR (approved):
utAck

PR #315, opened 2024-10-12
test: add more tests for snapshot legacy format

wa0x6e:
This PR adds some tests to ensure that the snapshot resolver legacy format (space/address) always return same result as the new one (space/network:address), for both v1 and sx


ChaituVR (approved):
utAck

PR #320, opened 2024-10-28
feat: add web3-name-sdk resolve name to address

minghuipei:


src/addressResolvers/snapshot.ts
```diff
@@ -33,6 +34,16 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
   }
 }
 
-export async function resolveNames(): Promise<Record<Handle, Address>> {
-  return {};
+export async function resolveName(name: string): Promise<Address | undefined> {
```
bonustrack, line 37: You shouldn't replace `snapshot.ts` resolver but instead making a new file `spaceid.ts`

package.json
```diff
@@ -43,7 +44,8 @@
     "sharp": "^0.30.1",
     "starknet": "^6.11.0",
     "ts-node": "^10.8.1",
-    "typescript": "^4.7.3"
+    "typescript": "^5.6.3",
+    "viem": "^2.21.35"
```
bonustrack, line 48: I don't see this being used, could you remove it?

package.json
```diff
@@ -28,6 +28,7 @@
     "@snapshot-labs/snapshot-sentry": "^1.5.5",
     "@snapshot-labs/snapshot.js": "^0.12.28",
     "@unstoppabledomains/resolution": "^9.2.2",
+    "@web3-name-sdk/core": "^0.2.0",
```
bonustrack, line 31: Isn't there a way to do onchain call to get the name instead of using that library?

package.json
minghuipei: > I don't see this being used, could you remove it?

Our library need these.

package.json
minghuipei: > Isn't there a way to do onchain call with Ethers.js to get the name instead of using that library?

Our library is specifically designed to achieve that functionality and also supports additional parsing features. For more details, please refer to https://docs.space.id/developer-guide/web3-name-sdk

src/addressResolvers/snapshot.ts
minghuipei: I got it.

PR #325, opened 2024-12-03
fix: convert network/networkId using lookup table

Sekhmet:
Closes: https://github.com/snapshot-labs/workflow/issues/310

Currently we use networkId for cacheKeys but this is not always set. This means that we might end up using wrong cached asset.

To solve this we generate lookup table using chains data that will minimize this issue (there can be still some cases of missing mapping, but way less likely now).

## Test plan

- Log `key1`.
- Try http://localhost:3008/token/eip155:137:0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee?s=64
- Try http://localhost:3008/token/eip155:1:0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee?s=64

You get different images and logged key1 is different.


src/chains.json
```diff
@@ -0,0 +1,4124 @@
...
+    "heco": "128",
+    "Innovator": "129",
+    "unichain": "130",
+    "tgram": "131",
+    "nfic": "132",
+    "HSKT": "133",
+    "rlc": "134",
+    "AlyxTestnet": "135",
+    "deam": "136",
+    "pol": "137",
```
bonustrack, line 2200: This may require some change in our UI, to use `pol` instead of `matic`

src/scripts/generate-chains.ts
```diff
@@ -0,0 +1,45 @@
...
+  const chains: Chain[] = JSON.parse(inputData);
+
+  const CHAIN_ID_TO_SHORTNAME = [
+    ...chains.map(chain => [String(chain.chainId), chain.shortName]),
+    ...CUSTOM_CHAINS
+  ];
+
+  const output = {
+    CHAIN_ID_TO_SHORTNAME: CHAIN_ID_TO_SHORTNAME.reduce((acc, chain) => {
+      acc[chain[0]] = chain[1];
```
bonustrack, line 33: Maybe we can enforce lowercase for the shortname wdyt

src/chains.json
Sekhmet: Added compatibility for matic for now (Safe still uses `matic` for example).

If you use `pol` and `matic` it will both map to 137.
If you use 137 it will map to matic.

bonustrack (approved):
utACK

PR #328, opened 2024-12-14
fix: update urls to SX api

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/331

Update SX APIs urls to use subgrapher, to avoid rate limit


bonustrack (approved):
utACK

PR #340, opened 2025-01-30
fix: set timeout for unstoppable domains resolver

wa0x6e:
Depends on https://github.com/snapshot-labs/snapshot.js/pull/1114

This PR will use the new option in the SDK to set a custom timeout when connecting to brovider.

This will let us set the same timeout value (5 seconds) for all address resolvers.

UnnstoppableDomain not having a timeout was making all address resolvers request respond after 25 seconds, when brovider node is having issues.


ChaituVR:
snapshot.js is updated for stamp

ChaituVR (approved):
utAck

PR #347, opened 2025-02-20
feat: add params to customize resize fit

wa0x6e:
Toward https://github.com/snapshot-labs/sx-monorepo/pull/1153

This PR adds a new query params `fit` when querying image, to control how the image is resized.

It accepts any values accepted by https://sharp.pixelplumbing.com/api-resize#resize (`contain`, `cover`, etc).

The current fit (cover) is not always ideal, as sometimes, we need the image in its original ratio

### How to test

- Go to http://localhost:3008/space-cover/test.wa0x6e.eth?h=200&w=300&fit=fill
- Go to http://localhost:3008/space-cover/test.wa0x6e.eth?h=200&w=300&fit=inside

It should return images with different resize ratio (missing `fit` query params will default to `cover`)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -45,12 +45,13 @@ router.get(`/clear/:type(${TYPE_CONSTRAINTS})/:id`, async (req, res) => {
     if (type === 'address' || type === 'name') {
       result = await clearCache(id);
     } else {
-      const { address, network, w, h, fallback, cb } = await parseQuery(id, type, {
+      const { address, network, w, h, fallback, cb, fit } = await parseQuery(id, type, {
         s: constants.max,
         fb: req.query.fb,
-        cb: req.query.cb
+        cb: req.query.cb,
+        fit: req.query.fit
       });
-      const key = getCacheKey({ type, network, address, w, h, fallback, cb });
+      const key = getCacheKey({ type, network, address, w, h, fallback, cb, fit });
       result = await clear(key);
     }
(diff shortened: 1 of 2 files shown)


ChaituVR (approved):
tAck

wa0x6e:
> Looks like in this case width and height are being ignored image

What's the full url ? then no height or width are passed, it default to 44px

ChaituVR:
> > Looks like in this case width and height are being ignored image
> 
> What's the full url ? then no height or width are passed, it default to 44px

Same as mentioned in description 
- Go to http://localhost:3008/space-cover/test.wa0x6e.eth?h=200&w=300&fit=fill
- Go to http://localhost:3008/space-cover/test.wa0x6e.eth?h=200&w=300&fit=inside

wa0x6e:
> > > Looks like in this case width and height are being ignored image
> > 
> > 
> > What's the full url ? then no height or width are passed, it default to 44px
> 
> Same as mentioned in description
> 
> * Go to http://localhost:3008/space-cover/test.wa0x6e.eth?h=200&w=300&fit=fill
> * Go to http://localhost:3008/space-cover/test.wa0x6e.eth?h=200&w=300&fit=inside

Should be the expected behavior then, `inside` will keep the ratio, and use the biggest dimension (in that case, the width)

ChaituVR (approved):
tAck

PR #354, opened 2025-03-15
feat: return domain from reverse registrar

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/485

The `lookup_domains` method now returns domains set in the reverse registrar

### Test

```shell
--data '{
    "method": "lookup_domains",
    "params":
        "0x7aeB96261e9dC2C9f01BaE6A516Df80a5a98c7eB"

}' | jq
```

It should return

```json
{
  "jsonrpc": "2.0",
  "result": [
    "appdefi.eth",
    "defi.app"
  ],
  "id": null
}
```

Note that this address->name feature is only added to `lookup_domains`, and not also to `lookup_address`, as it's an expensive request. It will mainly be used for offchain space creation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@ethersproject/address": "^5.7.0",
     "@ethersproject/bignumber": "^5.7.0",
     "@ethersproject/contracts": "^5.7.0",
-    "@ethersproject/hash": "^5.8.0",
+    "@ethersproject/hash": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
     "@metamask/jazzicon": "^2.0.0",
     "@self.id/core": "^0.3.0",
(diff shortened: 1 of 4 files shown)


wa0x6e:
Still need to confirm that the `getENSOwner` utils function in the SDK see those owners correctly

ChaituVR:
> Still need to confirm that the `getENSOwner` utils function in the SDK see those owners correctly

I don't get it 🤔

wa0x6e:
> > Still need to confirm that the `getENSOwner` utils function in the SDK see those owners correctly
> 
> I don't get it 🤔

If we merge this PR, the address `0x7aeB96261e9dC2C9f01BaE6A516Df80a5a98c7eB` will be able to see its domain `defi.app` in the list of available domains the offchain space creation.

Remaining issue is that the verification that this address is truly the owner of this domain will fail on sequencer, as there's no concept of owner for TLD domain.

Screenshot 2025-03-16 at 23 02 21

The address is not registered as owner, still looking for a way to verify ownership of TLD, maybe with https://github.com/ensdomains/ensjs/blob/main/docs/dns/function.getDnsOwner.md

PR #363, opened 2025-04-01
fix: make sure chainIds are strings and unique

ChaituVR:
Right now UI send chainIds as Int, we need to convert it into string

### How to test:

Send request with:
```bash
--data '{
    "method": "lookup_domains",
    "params": "0x24F15402C6Bb870554489b2fd2049A85d75B982f",
    "network": 1
}'
```


wa0x6e (approved):
tAck

PR #367, opened 2025-04-12
fix: mute 429 errors from lens

wa0x6e:
Fix https://snapshot-labs.sentry.io/issues/4651451195/events/latest/?notification_uuid=94935e9b-e1ed-438b-abfa-5941ed814c91&project=4505506866593792&referrer=latest-event

This PR will stop logging errors when lens api is returning 429 response.

Those errors can not be fixed, and are just consuming sentry quota


ChaituVR (approved):
utAck

PR #368, opened 2025-04-12
refactor: use d3 api to fetch shibarium domains

wa0x6e:
Similar to https://github.com/stamp-labs/stamp/pull/366 , use d3 api to retrieve all owned shib domains for a given address.

Also some small refactoring on `getOwner` to avoid repetition.

### Test

Use same tests as https://github.com/stamp-labs/stamp/pull/362


ChaituVR (approved):
tAck

PR #371, opened 2025-04-17
fix: handle unclaimed names when fetching shib names owner

wa0x6e:
This PR fix an issue with unclaimed .shib names.

### What is an unclaimed name ?

.shib names can be bought with web2 means, such as credit card. Names bought via such mean will not be associated to an EOA wallet, and retrieving the owner via d3 api will not return owner field.

Screenshot 2025-04-14 at 13 05 28

### Fix

In those cases, we should fallback, and return the resolved address instead, and consider the resolved address as owner, like we already do with the other offchain resolved address (like coinbase .cb.id names).

This will fix the error encountered in the ui pr: https://github.com/snapshot-labs/sx-monorepo/actions/runs/14441352799/job/40492017489?pr=1312

Resolving names are done via https://docs.d3.app/resolve-d3-names#d3-connect-sdk

### Tests

```shell
--data '{
    "method": "get_owner",
    "params": "renato.shib",
    "network": "157"
}'
```

It should returned the resolved address as stated here instead of the empty address: https://puppyscan.shib.io/name-domains?name=renato.shib&only_active=true

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,35 @@
 name: Test
+
 on: [push]
+
 jobs:
   test:
-    uses: snapshot-labs/actions/.github/workflows/test.yml@main
-    secrets: inherit
-    with:
-      redis: true
+    name: Test
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    env:
+      D3_API_KEY_MAINNET: ${{ secrets.D3_API_KEY_MAINNET }}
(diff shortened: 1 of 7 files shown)


ChaituVR:
@wa0x6e yarn install fails in my local as well, similar to test and lint workflows.

wa0x6e:
@ChaituVR `yarn install` issue should be solved

wa0x6e:
Tests failing are unrelated to this PR

ChaituVR (approved):
tAck

PR #375, opened 2025-04-29
fix: lookup_domains should default to use all supported mainnet

wa0x6e:
When running the `lookup_domains`, like 

```shell
--data '{
    "method": "lookup_domains",
    "params":
        "0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6"

}' | jq
```

it currently only returns ens name, since when missing, the networks params default to `1`

This PR fix the issue by adding shibarium chain id to the default list


ChaituVR (approved):
tAck

PR #412, opened 2025-08-27
chore: remove polygon from snapshot resolver

wa0x6e:
This PR removes the polygon network from the list of supported network by snapshot-x api

Resolver is dead since the removal of the polygon network on sx, leading to failing test.


ChaituVR (approved):
utAck

PR #413, opened 2025-09-05
fix: support pagination when fetching domains from unstoppable domains

wa0x6e:
Reported by unstoppable domains team, in telegram group

### Issue

Current implementation only fetch data from first page, when looking for domains from unstoppable domains api.

While this is enough in most scenario (the pagination is returning 100 results per page), it may lead to some empty results when the domain we're looking for is on subsequent pages.

### Fix

This PR will add pagination support when fetching data from their API

### Test

This is an example where the .sonic domain is on page 2

```shell
curl \
  -X POST \
  -d '  {
    "method": "lookup_domains",
    "params": "0x64F23b8c6261f8717E486955383679b9aEC67A56"
  }' \
  'http://localhost:3008'
```

Should return 

```json
{
  "jsonrpc": "2.0",
  "result": [
    "partner-engineering.sonic"
  ],
  "id": null
}
```

The query above is returning no results on master.

Old regular queries should still continue working

```shell
curl \
  -X POST \
  -d '  {
    "method": "lookup_domains",
    "params": "0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6"
  }' \
  'http://localhost:3008'
```

Should return 

```json
{
  "jsonrpc": "2.0",
  "result": [
    "boorger.eth",
    "boorger.shib",
    "boorger.sonic"
  ],
  "id": null
}
```

### DIFF
--- a/src/lookupDomains/unstoppableDomains.ts
+++ b/src/lookupDomains/unstoppableDomains.ts
@@ -38,7 +38,7 @@ export default async function lookupDomains(address: Address, chainId: string): 
   try {
     while (cursor !== null) {
       const data = await fetchDomains(address, cursor);
-      cursor = data.next?.split('=').pop() || null;
+      cursor = data.next?.split('cursor=').pop() || null;
       domains.push(...data.data.map((domain: any) => domain.meta.domain));
     }
     return normalizeHandles(domains);


src/lookupDomains/unstoppableDomains.ts
```diff
@@ -10,26 +10,38 @@ function normalizeHandles(handles: Handle[]): Handle[] {
...
-
-    const data = await resp.json();
+  const domains: string[] = [];
+  let cursor = '0';
 
-    return normalizeHandles(data.data.map((domain: any) => domain.meta.domain));
+  try {
+    while (cursor !== null) {
+      const data = await fetchDomains(address, cursor);
+      cursor = data.next?.split('=').pop() || null;
```
ChaituVR, line 41: For extra safety
```suggestion
      cursor = data.next?.split('cursor=').pop() || null;
```

ChaituVR (approved):
tAck

PR #424, opened 2025-11-18
chore: use starknet provider which use brovider

ChaituVR:
### Summary
- removed INFURA_API_KEY from .env.example
- updated Starknet provider initialization to use getProvider utility

### How to test
- Remove `INFURA_API_KEY` from your env
- (optional) clear cache with `curl --location 'http://localhost:3008/clear/avatar/starknet.stark'`
- Run `curl --location 'http://localhost:3008/avatar/starknet.stark?s=200'` 
- It should not return a image before and should return an image after fix

### TODO
- [ ] Remove `INFURA_API_KEY` and `SN_INFURA_API_KEY` from DO


bonustrack (approved):
utACK

PR #429, opened 2026-04-16
fix: use shortTtl for empty address resolver cache entries

ChaituVR:
### Issue

Test case on sx-monorepo fail with
image

This was because at some point, `wanki.moe`. returned null and the cache stored it

## Summary
- Failed name resolutions were cached with the same 12h TTL as successful ones
- A transient failure (e.g. a 5s CCIP-read timeout) could block a valid name from resolving for up to 12 hours — hit this with `wanki.moe`, which resolves fine via ethers but returned `{}` from stamp because of a stale empty cache entry
- Empty values now expire after `shortTtl` (1h); real resolutions keep the full `ttl` (12h)

## Test plan

Run locally (`yarn start`, listens on :3008) with Redis available.

**Trigger a successful resolution — should be cached 12h:**
```bash
curl -s -X POST http://localhost:3008 \
  -H "Content-Type: application/json" \
  -d '{"method":"resolve_names","params":["vitalik.eth"],"network":1}'
# {"jsonrpc":"2.0","result":{"vitalik.eth":"0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"},"id":null}

redis-cli TTL address-resolvers:vitalik.eth
# ~43200 (ttl)
```

**Trigger an unresolvable name — should be cached 1h (shortTtl), not 12h:**
```bash
curl -s -X POST http://localhost:3008 \
  -H "Content-Type: application/json" \
  -d '{"method":"resolve_names","params":["nonexistent-random-name-xyz.com"],"network":1}'
# {"jsonrpc":"2.0","result":{},"id":null}

redis-cli TTL address-resolvers:nonexistent-random-name-xyz.com
# ~3600 (shortTtl) — before this fix it was ~43200
```

**Clear cache endpoint still works:**
```bash
curl -s http://localhost:3008/clear/name/vitalik.eth
# {"status":"ok"}
```

- [ ] Existing cache integration tests still pass
- [ ] TTL values on Redis keys match expected (12h positive, 1h empty)

### DIFF
--- a/src/addressResolvers/cache.ts
+++ b/src/addressResolvers/cache.ts
@@ -21,7 +21,9 @@ export function setCache(payload: Record<string, string>) {
 
   const transaction = redis.multi();
   Object.entries(payload).map(([key, value]) =>
-    transaction.set(`${KEY_PREFIX}:${key}`, value || '', { EX: constants.ttl })
+    transaction.set(`${KEY_PREFIX}:${key}`, value || '', {
+      EX: value ? constants.ttl : constants.shortTtl
+    })
   );
 
   return transaction.exec();


ChaituVR:
Can review later, not urgent

wa0x6e (commented):
This does not seem to solve the issue, and just a band-aid.

Reducing the cache time to 1h will still cache and fail the mentioned CI task for 1h, and also have the side-effect of decreasing cache time for all genuine name without value.

A more definitive fix would be to address the root cause: do not cache if the empty value is due to an error (will be a bigger fix)

Each resolver would throw a consistent error when needed (currently, some resolver throw, and some other swallow them), and the caching will be skipped when no value are found due to resolver error

ChaituVR:
hmm not sure if there was some error at that time, but 12 hour cache is bit strict no?

PR #435, opened 2026-05-23
fix(addressResolvers): silence axios 504 errors (STAMP-7)

wa0x6e:
## Summary
- `isSilencedError` chained candidate fields with `||`, so axios's `error.code = 'ERR_BAD_RESPONSE'` short-circuited and the 504 on `error.response.status` was never checked
- Iterate each candidate independently and include `error.response?.status`, which is where axios surfaces the upstream HTTP status
- Follow-up to #434, which only handled the wrapped-ethers shape

## Context
[STAMP-7](https://snapshot-labs.sentry.io/issues/STAMP-7) is the top recurring error on the avatar endpoints (256k+ occurrences). Breadcrumbs trace it to `hub.snapshot.org/graphql` returning 504, captured at `src/addressResolvers/snapshot.ts:33` because `isSilencedError` didn't recognize the axios error shape.

## Test plan
- [x] Added two unit tests covering the axios 504 shape and a non-504 axios error (to guard against over-silencing)
- [x] Existing wrapped-ethers 504 test still passes
- [ ] Monitor STAMP-7 occurrence rate after deploy

Fixes STAMP-7

### DIFF
--- a/src/addressResolvers/utils.ts
+++ b/src/addressResolvers/utils.ts
@@ -54,10 +54,11 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
     'status=504',
     ...(additionalMessages || [])
   ];
+  const codes = [error.error?.code, error.error?.status, error.code, error.response?.status];
   return (
     messages.some(m => error.message?.includes(m) || error.error?.message?.includes(m)) ||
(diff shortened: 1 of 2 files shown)


ChaituVR (approved):
utAck

PR #436, opened 2026-05-23
test: disable coverage by default for faster local runs

wa0x6e:
## Summary
- `jest.config.ts`: default `collectCoverage` to `false` so `yarn test` runs without coverage overhead locally
- `package.json`: add `test:coverage` script; drop now-redundant `--collectCoverage=false` from `test:integration` and `test:e2e`
- `.github/workflows/test.yml`: CI now runs `yarn test:coverage` so Codecov still gets reports

## Test plan
- [ ] `yarn test` runs without generating a `coverage/` directory
- [ ] `yarn test:coverage` generates coverage as before
- [ ] CI run uploads coverage to Codecov successfully

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
         run: yarn install --frozen-lockfile
 
       - name: Test
-        run: yarn test
+        run: yarn test:coverage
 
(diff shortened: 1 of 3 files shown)


ChaituVR (approved):
utAck

PR #437, opened 2026-05-23
fix(addressResolvers): skip non-EVM addresses in shibarium lookup

wa0x6e:
## Summary

Adds an `isEvmAddress` filter to `shibarium.lookupAddresses`, matching the pattern already used by every other EVM resolver (`ens`, `lens`, `spaceId`, `unstoppableDomains`).

## Root cause

64-hex Starknet-shaped inputs produce a first DNS label of `0x` + 64 hex = 66 characters, exceeding the RFC 1035 §2.3.4 limit of 63 octets. Cloudflare DoH responds with `{"error":"Invalid query name ..."}` and no `Status` field, so `@webinterop/dns-connect`'s DoH resolver throws ``Received error status from DNS server: undefined.``. Reproduced by `curl`ing the DoH endpoint directly with the exact `input.addresses` from Sentry. Behavior is deterministic for every Starknet address — consistent with the 15k+ event volume on STAMP-36.

The throw is caught at `src/addressResolvers/index.ts:56` (`catch (e) {}`), so user requests are unaffected, but `capture()` floods Sentry on every Starknet lookup.

## Test plan

- [x] New test in `test/integration/addressResolvers/shibarium.test.ts` feeds the exact STAMP-36 input and asserts `lookupAddresses(...)` resolves to `{}` without throwing or hitting DNS.
- [x] Existing shibarium integration tests still pass (7/7).
- [x] `yarn typecheck` clean.
- [x] `yarn lint` clean.

Fixes STAMP-36

### DIFF
--- a/src/addressResolvers/shibarium.ts
+++ b/src/addressResolvers/shibarium.ts
@@ -1,6 +1,6 @@
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import { Address, Handle } from '../utils';
-import { FetchError, isSilencedError, withoutEmptyValues } from './utils';
+import { FetchError, isEvmAddress, isSilencedError, withoutEmptyValues } from './utils';
 import { DNSConnect } from '@webinterop/dns-connect';
 import constants from '../constants.json';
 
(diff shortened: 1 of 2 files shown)


ChaituVR (approved):
tAck

Just blocked by #441

PR #438, opened 2026-05-23
chore(lookupDomains): log subgraph response on ENS errors

wa0x6e:
## Summary
- Refs [STAMP-49](https://snapshot-labs.sentry.io/issues/STAMP-49): `TypeError: Cannot read properties of undefined (reading 'account')` thrown by the strict destructure in `src/lookupDomains/ens.ts`.
- Single occurrence so far, not reproducible against the live ENS subgraph today — the trigger appears to be a transient response shape from the `subgrapher.snapshot.org` proxy (HTTP 200 with no top-level `data` field).
- This PR is **diagnostic only**, not a fix. It hoists the axios response out of the `await` so the existing `capture()` call can attach the response `status` and `body` as a `response` context block. The next occurrence will reveal the actual shape so we can pick a permanent fix (defensive destructure, silenced GraphQL-error envelope handling, etc).

## Test plan
- [x] `tsc --noEmit` passes
- [x] Happy path unchanged (the destructure still runs the same way; only the variable scope moved)
- [ ] On next STAMP-49 occurrence, verify Sentry event shows the new `response` context with `status` + `body`

### DIFF
--- a/src/lookupDomains/ens.ts
+++ b/src/lookupDomains/ens.ts
@@ -44,9 +44,12 @@ export default async function lookupDomains(
 ): Promise<Handle[]> {
   if (!constants.ensSubgraph[chainId]) return [];
 
-  let response: Awaited<ReturnType<typeof graphQlCall>> | undefined;
   try {
-    response = await graphQlCall(
+    const {
+      data: {
+        data: { account }
+      }
+    } = await graphQlCall(
       constants.ensSubgraph[chainId],
       `query Domain($id: String!) {
         account(id: $id) {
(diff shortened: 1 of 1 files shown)


src/lookupDomains/ens.ts
```diff
@@ -66,6 +63,12 @@ export default async function lookupDomains(
       { id: address.toLowerCase() }
     );
 
+    const {
+      data: {
+        data: { account }
+      }
+    } = response;
```
ChaituVR: But why this change ? 😅 we can just extract `status`

ChaituVR (approved):
utAck (comments are optional, not imp)

ChaituVR (approved):
utAck

PR #422, opened 2025-11-04
tests: disable puppynet tests

wa0x6e:
Tests are failing since credentials to the puppynet API services are invalid

PR #441, opened 2026-05-25
chore: drop unused D3 testnet (puppynet) test cases

ChaituVR:
## Summary
- Removes the last puppynet (chainId `157`) test in `getOwner.test.ts` (the one currently failing with `Error fetching owner: Unauthorized` because the CI `D3_API_KEY_TESTNET` is expired/invalid).
- Removes the three already-disabled puppynet tests left commented by #422 (dead code).
- Drops `D3_API_KEY_TESTNET` from `.github/workflows/test.yml` since nothing in the test suite references it anymore.

## Why
After #422, only one `'157'` test remained (`should return an empty address when the name does not have a primary names`). It would pass *either* without a key (short-circuits to `EMPTY_ADDRESS`) *or* with a valid key, but an invalid/expired key makes the resolver throw — the current CI state. Rather than chase a rotating testnet credential to keep one always-empty assertion green, drop the testnet coverage entirely. Mainnet (chainId `109`) tests are unaffected and still depend on `D3_API_KEY_MAINNET`.

`src/getOwner/shibarium.ts` and `src/lookupDomains/shibarium.ts` still reference `process.env.D3_API_KEY_TESTNET` for runtime use, so production testnet support is unchanged — only the CI dependency is removed.

## Test plan
- [x] Local typecheck passes.
- [ ] CI on this branch should now stop failing on the puppynet `Unauthorized` assertion.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,6 @@ jobs:
     timeout-minutes: 10
     env:
       D3_API_KEY_MAINNET: ${{ secrets.D3_API_KEY_MAINNET }}
-      D3_API_KEY_TESTNET: ${{ secrets.D3_API_KEY_TESTNET }}
       UNSTOPPABLE_DOMAINS_API_KEY: ${{ secrets.UNSTOPPABLE_DOMAINS_API_KEY }}
 
     steps:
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

PR #394, opened 2025-07-13
feat: add UnstoppableDomains to lookup_domains

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/598

This PR will return list of domains owned by an address for domains from unstoppable domains.

### Test

Set `UNSTOPPABLE_DOMAINS_API_KEY` in your .env file
Need to get a key from https://docs.unstoppabledomains.com/openapi/metadata/

```shell
curl --request POST \
  --url http://localhost:3008/ \
  --header 'content-type: application/json' \
  --data '{
  "method": "lookup_domains",
  "params": "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7"
}'
```

It should return this list:

```json
{
  "jsonrpc": "2.0",
  "result": [
    "fabien.eth",
    "textcoin.eth",
    "test1.snapback.eth",
    "fabien.sismo.eth",
    "test2.snapback.eth",
    "space.fabien.eth",
    "textcoin.crypto",
    "snapshot.dao",
    "snapshot.crypto"
  ],
  "id": null
}
```

`snapshot.cryto` is a domain from unstoppableDomain.

Note that unstoppableDomains also return result from ENS registry.

Need to add the `UNSTOPPABLE_DOMAINS_API_KEY` to the repo env var for the CI tests to pass

PR #362, opened 2025-03-31
refactor: refactor lookupDomains to support multiple resolvers

wa0x6e:

PR #448, opened 2026-05-31
fix(lookupDomains): absorb per-resolver failures

### LINKED ISSUE #493: lookupDomains is not best effort: one failing resolver fails the whole request
`src/lookupDomains/index.ts` runs all resolvers (ENS, Shibarium,
Unstoppable Domains) through a bare `Promise.all`. If any single
resolver throws, the whole call rejects:

- ENS and Shibarium results for the request are discarded
- the client gets a 500 instead of partial results
- the api.ts catch captures the error a second time (`FetchError`
  extends `Error` with no `code`, so the `code !== 400` guard passes),
  duplicating whatever the resolver already sent to Sentry

Seen in practice with STAMP-6X: a UD rate-limit burst turned every
`lookup_domains` request for chain 146 into a 500, even though ENS and
Shibarium were healthy.

### Proposed fix

Per-resolver `.catch(() => [])` in index.ts so a failing resolver
contributes an empty result instead of rejecting the batch. Resolvers
already capture their own errors, so nothing is lost silently.

wa0x6e:
Fix https://github.com/snapshot-labs/stamp/issues/493

## Roadmap context

Step **2 of 3** in cleaning up resolver error handling. End goal: resolvers throw raw errors, orchestrators own capture/silence policy, `FetchError` sentinel deleted, api.ts catch-all is a pure safety net for unexpected errors, and `lookup_domains` is best-effort.

1. **#492 (merged)** — stopgap: api.ts skips re-capturing `FetchError`, class gets a name/message. Fixed the Sentry noise (STAMP-6V / STAMP-35) at the boundary, changed no behavior.
2. **This PR** — behavior change, isolated on purpose: one failing resolver/chain no longer rejects the whole `lookup_domains` request; it contributes `[]` and the client gets a partial 200 instead of a 500.
3. **#496** — refactor, no behavior change: move capture into the orchestrators, delete per-resolver try/catch and the `FetchError` sentinel entirely. Blocked on this PR: its `lookupDomains` half (capture + swallow at index) only makes sense once partial success is the accepted contract.

## Summary

`lookupDomains` fans out over `[ens, shibarium, unstoppableDomains]` × chains via `Promise.all` with **no per-resolver error handling**. Each resolver captures (with context) or silences its own error and then throws a bare `FetchError`, which rejects the whole `Promise.all` and turns the entire request into a 500 — even when other resolvers succeeded.

> **Note (post-#492):** this PR originally also fixed the Sentry double-capture (STAMP-35): the escaped `FetchError` was re-captured context-free at api.ts. #492 has since fixed that at the boundary, so the Sentry motivation is gone. What remains — and what this PR is now solely about — is the **request behavior**: best-effort partial results instead of all-or-nothing.

## Why this is the right fix

Propagation was correct in the **single-resolver origin** (#231), where `lookupDomains` was one ENS call and api.ts was its only error boundary. It was never revisited when the function became a fan-out (#362, #394) — both pure feature additions that didn't touch error handling. The sibling fan-out `addressResolvers._call` already swallows per-resolver failures with `try { … } catch {}` for exactly this reason.

This change brings `lookupDomains` in line with that established pattern: catch each resolver's rejection and return `[]`, so one failing chain no longer rejects the lookup or escapes to api.ts. The api.ts catch remains a genuine safety net for *unexpected* errors (e.g. `getOwner`); informative capture + silencing stay at the resolver layer (until #496 centralizes them).

## Behavior change

Previously, **any** single resolver/chain failure rejected the whole `lookup_domains` request → the client got a **500**. Now the failing resolver contributes `[]` and the other chains' results are still returned → the client gets a **partial 200**.

This path is **not cached** (no redis/cache layer on `lookup_domains`, unlike `lookup_addresses`/`resolve_names`), so the partial result is per-request only and self-corrects on the next call — there is no risk of caching an incomplete result. The new behavior matches how `addressResolvers` already treats per-resolver failures.

Side effect on #492's code: once failures are absorbed here, no `FetchError` reaches the api.ts catch-all from this path anymore, making the `instanceof FetchError` skip dead code on the live path. That's expected — #496 deletes it along with the class.

## Test plan

- [x] Added `test/unit/lookupDomains.test.ts` (fully mocked resolvers; CI runs it via `test:coverage`) — a resolver rejecting no longer propagates and the other resolvers' results are returned; all-fail returns `[]`. Fails before the change, passes after.
- [x] `yarn typecheck`
- [x] `yarn lint`

### DIFF
--- a/src/lookupDomains/index.ts
+++ b/src/lookupDomains/index.ts
@@ -26,7 +26,7 @@ export default async function lookupDomains(
 
   RESOLVERS.forEach(resolver => {
     chainIds.forEach(chain => {
-      promises.push(resolver(address, chain));
+      promises.push(resolver(address, chain).catch(() => []));
     });
   });
 
--- a/test/unit/lookupDomains.test.ts
+++ b/test/unit/lookupDomains.test.ts
@@ -1,0 +1,42 @@
+import { FetchError } from '../../src/addressResolvers/utils';
+import lookupDomains from '../../src/lookupDomains';
+import ens from '../../src/lookupDomains/ens';
+import shibarium from '../../src/lookupDomains/shibarium';
+import unstoppableDomains from '../../src/lookupDomains/unstoppableDomains';
+
+jest.mock('../../src/lookupDomains/ens', () => ({
+  __esModule: true,
+  DEFAULT_CHAIN_ID: '1',
+  default: jest.fn()
+}));
+jest.mock('../../src/lookupDomains/shibarium', () => ({
+  __esModule: true,
+  DEFAULT_CHAIN_ID: '109',
+  default: jest.fn()
+}));
+jest.mock('../../src/lookupDomains/unstoppableDomains', () => ({
+  __esModule: true,
+  DEFAULT_CHAIN_ID: '146',
+  default: jest.fn()
+}));
+
+const VALID_ADDRESS = '0x24F15402C6Bb870554489b2fd2049A85d75B982f';
+const CHAINS = ['1', '109', '146'];
+
+describe('lookupDomains - resolver failures', () => {
+  it('does not propagate a resolver failure and returns the other resolvers results', async () => {
+    (ens as jest.Mock).mockRejectedValue(new FetchError());
+    (shibarium as jest.Mock).mockResolvedValue(['boorger.shib']);
+    (unstoppableDomains as jest.Mock).mockResolvedValue([]);
+
+    await expect(lookupDomains(VALID_ADDRESS, CHAINS)).resolves.toEqual(['boorger.shib']);
+  });
+
+  it('returns an empty array when every resolver fails', async () => {
+    (ens as jest.Mock).mockRejectedValue(new FetchError());
+    (shibarium as jest.Mock).mockRejectedValue(new FetchError());
+    (unstoppableDomains as jest.Mock).mockRejectedValue(new FetchError());
+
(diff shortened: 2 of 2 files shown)


tony8713:
On the merge question: I'd close #493 with this PR rather than fold it into #496, since its proposed fix is this diff verbatim and its double-capture bullet went with #492, so there's no remainder left to merge. #496 stands on its own and is bigger than these lines: only its lookupDomains half touches them, and the addressResolvers half doesn't depend on this PR at all. Agreed that #495 is a separate concern.

tony8713 (approved):
LGTM. I checked the behaviour rather than just the diff: on master a `lookup_domains` where ENS fails and Shibarium succeeds comes back `500` with no result, with your line it comes back `200` with `["boorger.shib"]`.

One thing before merge, this needs a rebase. `test/unit/api.test.ts` landed with #492 after you cut the branch, and both of its `lookup_domains` cases assert `expect(response.status).toBe(500)`, which is exactly the behaviour you're changing. There's no textual conflict so the PR still shows mergeable, but merged as-is those two go red. Flipping both to 200 is the whole fix (ran it: `test/unit` 5/5 green, typecheck and lint clean).

Worth fixing rather than deleting them, since they're the only place the 500 to partial 200 claim gets asserted through api.ts instead of at the orchestrator.

Unrelated to you: master's Test workflow has been red since before this PR (ens, starknet, zapper integration suites), so discount that part of the CI signal here.

PR #452, opened 2026-06-07
fix: drop node-canvas, render blockies via SVG + sharp

tony8713:
blockie.ts was the only consumer of node-canvas: it painted an NxN rect grid onto a canvas, then handed the buffer to sharp for resize/encode.

This rewrites it to emit an equivalent SVG (a background rect plus one rect per non-background cell, at the same 8x8 grid / scale 64 geometry) and rasterize with sharp, mirroring the approach the jazzicon resolver already uses.

Output is unchanged. The Xorshift PRNG, color and grid logic are ported verbatim from @download/blockies and called in the same order (color, bgcolor, spotcolor, then grid), so the PRNG stream does not diverge. Verified against the original lib for several addresses: background color, foreground color and the exact set of non-background cells are byte-identical. The only differences are sub-perceptual per-channel rasterizer rounding between canvas and sharp on identical HSL values and geometry.

This removes the node-canvas dependency entirely (it was the sole user), which also eliminates the macOS sharp-vs-canvas dual-GLib dylib clash that surfaces during local test runs.

Checks: yarn lint, yarn typecheck (tsc --noEmit) and yarn build all pass; no remaining canvas imports; canvas removed from package.json and yarn.lock.

Note: please verify blockie output parity in review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "@unstoppabledomains/resolution": "^9.2.2",
     "@webinterop/dns-connect": "^0.3.1",
     "axios": "^0.25.0",
-    "canvas": "^3.2.3",
     "compression": "^1.7.4",
     "cors": "^2.8.5",
     "dotenv": "^16.0.0",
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
utAck

PR #454, opened 2026-06-08
test(resolvers): pixel reference tests for deterministic identicons

tony8713:
## What this tests

Pixel-comparison (image-snapshot) tests for the avatar resolvers whose output is **deterministic from the input address**, so a committed baseline image is stable across runs.

### Resolvers covered
- **blockie** (`src/resolvers/blockie.ts`) - blockie identicon rendered locally from the address.
- **jazzicon** (`src/resolvers/jazzicon.ts`) - jazzicon rendered locally from the address.

### Resolvers excluded (and why)
All remote-fetch resolvers are intentionally left on size-only assertions: `ens`, `lens`, `coingecko`, `farcaster`, `snapshot`, `basename`, `trustwallet`, `zapper`, `selfid`. They resize an image fetched over the network, so a pixel snapshot would be flaky and network-bound.

**`starknet` is also excluded**: despite being avatar-ish, it queries the Starknet provider for a profile-picture URL and fetches that image over HTTP (`fetchImageOrMetadata` / `fetchHttpImage`). It does not generate an identicon locally, so it is not reproducible offline.

## How it works
- Uses the **`jest-image-snapshot`** library (which bundles `pixelmatch` + `pngjs` internally) instead of a hand-rolled matcher.
- `configureToMatchImageSnapshot` is registered in `test/setup-jest.ts` with `failureThreshold: 0.01` / `failureThresholdType: 'percent'` to absorb anti-aliasing noise across platforms / sharp versions.
- Resolver output is WebP, which `jest-image-snapshot` cannot read, so each test converts the output to a PNG buffer via `sharp(result).png().toBuffer()` before asserting `toMatchImageSnapshot()`.
- A `customSnapshotIdentifier` per address is set because `jest.retryTimes()` is enabled globally.
- Baselines are managed by the library under `test/integration/resolvers/__image_snapshots__/` (4 images: blockie + jazzicon for 2 fixed fixture addresses), committed to the repo.
- Tests: `test/integration/resolvers/blockie.pixel.test.ts`, `test/integration/resolvers/jazzicon.pixel.test.ts`.

## Regenerating baselines
```
yarn jest test/integration/resolvers/*.pixel.test.ts --ci=false -u
```
Regenerate only intentionally (e.g. an approved rendering change).

## devDeps
- added `jest-image-snapshot` + `@types/jest-image-snapshot`
- removed the now-redundant direct `pixelmatch` + `pngjs` (they were only used by the deleted custom matcher; `jest-image-snapshot` bundles them)

## Verification
- `yarn jest test/integration/resolvers/{blockie,jazzicon}.pixel.test.ts` -> 4 passed against the committed baselines.
- `yarn lint` clean, `yarn typecheck` clean.

For @wa0x6e to review.


wa0x6e (approved):
utAck

PR #455, opened 2026-06-08
test(resolvers): pixel snapshot tests for remote-fetch resolvers

tony8713:
## What

Adds jest-image-snapshot pixel tests for the **remote-fetch** resolvers that #454 did NOT cover (#454 handled the deterministic identicons: blockie + jazzicon).

## Resolvers covered (9)

ens, lens, snapshot (user avatar), space-sx, basename, coingecko, farcaster, trustwallet, zapper.

### Skipped (+ why)

- **selfid** - instantiates a `@self.id/core` `Core` (Ceramic gateway) at module load; mocking it cleanly would need a src-level seam, out of scope for an additive-only PR.
- **starknet** - uses a custom two-stage fetch (`provider.getStarkProfile` + a bespoke `fetchImageOrMetadata` axios call that branches on content-type); deterministic mocking would require either editing src or a brittle double mock. Left for a follow-up.

The deterministic resize/encode pipeline these would exercise is already covered by the 9 resolvers above.

## Approach: mock the fetch seam

All these resolvers fetch a remote image and then `resize()` it via sharp. The tests mock the network seam (the shared `src/resolvers/utils.fetchHttpImage`, plus the per-resolver URL-discovery layer: `graphQlCall`, `getProvider`, `getAvatar`, `node-fetch`, global `fetch`) so a **committed sample source image** is fed in. The test then asserts the **OUTPUT** of the deterministic sharp resize/encode pipeline via `toMatchImageSnapshot`. This guards what a sharp / image-lib bump would change, not the network.

Source fixtures: `test/fixtures/images/source/sample-avatar.png` + `sample-cover.png` (small fixed PNGs). Per-resolver inputs live in a NEW fixtures file `test/fixtures/remote-resolver-fixtures.ts` (the existing `image-snapshot-addresses.ts` is untouched).

## Additive only

New test files + new fixtures + new baselines only. **No existing test file modified, no resolver src modified** (`git diff master` shows additions only). Keeps this isolated from other in-flight PRs.

## Regenerating baselines

```
yarn jest "test/integration/resolvers/.*\.pixel\.test\.ts" --ci=false -u
```

## Verification

- New pixel tests pass against committed baselines (13 pixel snapshots total, incl. the 4 from #454, all green)
- `yarn lint` clean, `yarn typecheck` clean
- Existing tests untouched and still pass


tony8713:
Superseded by the real-fetch image-snapshot refactor (no mocks, per wa0x6e).

PR #456, opened 2026-06-08
test(resolvers): real-fetch image-snapshot comparison

tony8713:
## What changed

Refactors the resolver integration tests (`test/integration/resolvers/*.test.ts`) so that:

1. **Every resolver is tested for the real no-avatar path** using a NORMAL, non-special address (`0x556B...A70`, `NO_AVATAR_ADDRESS`), in addition to its existing real-avatar test. The zero address is deliberately NOT used as the no-avatar input: trustwallet/zapper special-case it (and `0xEeee...EEeE`) to the base-asset (ETH) icon, so it does not exercise the genuine no-avatar path.
2. **Only real image tests remain**: all `instanceof Buffer` / Buffer-length / `toBeImage(n)` size assertions are gone, replaced by real `toMatchImageSnapshot` pixel comparisons. The only non-image assertions left are the documented `returns false` / no-fallback cases.

No network mocking. Each remote-fetch resolver makes a real call to its actual upstream and the decoded pixels are compared to a committed baseline.

Supersedes #455 (the mocked-fetch approach), per wa0x6e's direction.

## No-avatar behaviour (corrected)

- **No resolver returns a fallback image for a genuine no-avatar input.** With a normal, non-special address, EVERY remote resolver returns `false` (ens, lens, snapshot all surfaces, space-sx, basename, trustwallet, zapper, coingecko, farcaster, starknet, selfid). These are asserted as `false`.
- **trustwallet** and **zapper** still special-case the native/ETH sentinel (the zero address / `0xEeee...EEeE`, which are in their ETH list) to the base-asset (ETH) icon, but that zero-address native-asset image case is NOT snapshot-tested here (it was dropped, since it is a sentinel branch rather than a genuine avatar/no-avatar path). Each resolver is tested only for its real-avatar (image) and genuine no-avatar (`false`) paths.
- **starknet** has no fallback image of its own: when an address has no custom pfp, starknet.id serves a DEFAULT identicon (`starknet.id/api/identicons/0`) which the resolver EXPLICITLY REJECTS -> `false`. Documented in the test. **Open question for wa0x6e:** should starknet stop rejecting the default identicon and return/snapshot it as a fallback image instead? That is a behavior change and not made here.

## Reliability handling

Real-fetch snapshots can drift with benign CDN re-encodes, so remote snapshots use `failureThreshold: 1.5%` (`remoteSnapshotOptions`); deterministic resolvers (blockie/jazzicon) keep the strict 0.01% default from `setup-jest.ts`. Stable, long-lived identities are used as inputs (consolidated in `test/fixtures/image-snapshot-addresses.ts`).

## Coverage overview

Thumbnails below are the actual committed baseline PNGs from `test/integration/resolvers/__image_snapshots__/` on this branch.

No-avatar inputs are NORMAL, non-special addresses (the zero address is special-cased by trustwallet/zapper and so is the wrong input for a no-avatar test). The trustwallet/zapper zero-address native-asset (ETH) sentinel is a separate branch and is not snapshot-tested here (the column below is `n/a` for it).

| Resolver (chain/surface) | Real avatar | No-avatar (non-special addr) | Native-asset sentinel |
|---|---|---|---|
| blockie |  | n/a (deterministic from any address)  | n/a |
| jazzicon |  | n/a (deterministic from any address)  | n/a |
| ens |  | returns false | n/a |
| lens (by handle) |  | returns false | n/a |
| lens (by address) |  | returns false | n/a |
| basename (by name) |  | returns false | n/a |
| basename (by address) |  | returns false | n/a |
| snapshot user-avatar |  | returns false | n/a |
| snapshot user-cover |  | returns false | n/a |
| snapshot space-avatar |  | returns false | n/a |
| snapshot space-cover |  | returns false | n/a |
| space-sx (arbitrum) |  | returns false | n/a |
| space-sx (optimism) |  | returns false | n/a |
| space-sx (sepolia) |  | returns false | n/a |
| space-sx (starknet) |  | returns false | n/a |
| space-sx (starknet sepolia) |  | returns false | n/a |
| space-sx (eth) | it.todo | returns false | n/a |
| trustwallet |  | returns false | n/a (zero-address native-asset case dropped) |
| zapper |  | returns false | n/a (zero-address native-asset case dropped) |
| coingecko |  (gated on `COINGECKO_API_KEY`) | returns false | n/a |
| farcaster |  (gated on `NEYNAR_API_KEY`) | returns false | n/a |
| starknet (simple image) |  | returns false | n/a |
| starknet (NFT by handle) |  | returns false | n/a |
| starknet (NFT by address) |  | returns false | n/a |
| selfid | it.todo (Ceramic gateway deprecated, suite skipped) | returns false | n/a |

## Baselines

25 baselines committed under `test/integration/resolvers/__image_snapshots__/`: 4 deterministic (blockie/jazzicon x2) + 21 remote/real. The new real-avatar baselines were minted on a temporary push-triggered GitHub Actions job (real network, with `COINGECKO_API_KEY` + `NEYNAR_API_KEY` present in CI), downloaded as an artifact, committed, and the temp workflow removed.

## CI

Lint + Test green on the branch against the committed baselines (23 image snapshots pass, 157 tests pass).

## Re-audit (live, correct inputs per resolver)

Per wa0x6e: re-verified every remote resolver's no-avatar path live with the CORRECT input format, to rule out false "returns false" findings caused by wrong inputs. Confirmed map (input used -> result):

- **ens** -> `false`. Tested the actual fallback URL path: a name WITH a reverse record but NO `avatar` text record (`registrar.firefly.eth`, `0cf5e.eth`) makes the resolver fall back to `metadata.ens.domains/mainnet/avatar/<name>`, which returns **404** for no-avatar names (not a default image) -> `fetchHttpImage` throws -> `false`. So even via the fallback URL, ens has no default image.
- **lens** -> `false` (zero address + no-account EOA; no Lens default pfp).
- **basename** -> `false` (zero address + base EOA without a basename).
- **trustwallet** -> **base-asset (ETH) image** ONLY for the ETH sentinel/zero address (in its ETH list). A non-ETH token with no logo (junk address, chains '', 1, 137, 8453) -> 404 -> `false`. Fallback image snapshotted.
- **zapper** -> same as trustwallet: **base-asset image** for ETH/zero only; junk token -> `false`. Fallback image snapshotted.
- **coingecko** -> `false` (zero + junk on chains 1/8453). Re-verified LIVE in CI with `COINGECKO_API_KEY` present (key set: true); the no-fallback `false` assertions pass against the live Pro API.
- **farcaster** -> `false` (zero, no-account, junk). Re-verified LIVE in CI with `NEYNAR_API_KEY` present (key set: true); Neynar returns no user -> no default pfp -> `false`.
- **starknet** -> `false` (valid 64-hex felt; upstream default identicon `starknet.id/api/identicons/0` is explicitly rejected). Confirmed genuine-false.
- **selfid** -> suite skipped (Ceramic gateway deprecated); no live retest possible.
- **snapshot** -> `false` (known no-fallback resolver; confirmed for zero address across user/space avatar+cover).

Keyed resolvers (coingecko/farcaster) were re-tested on a temporary push-triggered GitHub Actions job carrying the CI secrets (`--updateSnapshot` + artifact upload); the existing baselines matched with no drift (no baselines re-minted, nothing faked) and the temp workflow + probe were removed afterwards. No table changes were needed - the fallback map already matched reality.


wa0x6e (approved):
utAck

PR #457, opened 2026-06-09
refactor: migrate axios to native fetch

### LINKED ISSUE #206: refactor: migrate `axios` to `node-fetch`

tony8713:
## What

Moves the remaining direct HTTP calls from `axios` and `node-fetch` to Node's built-in `fetch`, then removes both direct dependencies.

## Behavior

- `graphQlCall` returns the GraphQL body directly, and still attributes envelope failures and non-2xx responses to the upstream host and status.
- The 5s budget on the GraphQL and Starknet metadata paths is now a total wall-clock deadline rather than the socket-inactivity timeout axios gave. That is a change in kind, not just in transport, and it matches what `fetchHttpImage` already did.
- A response prefixed by a UTF-8 BOM still parses. axios stripped one itself, so this preserves existing behavior rather than adding it, and the spec decode in `response.text()` is what keeps it working.
- Non-2xx still rejects, and on the GraphQL, image and Starknet paths it now rejects before the body is read. Since that leaves the response unconsumed, the deadline releases its socket rather than stranding it until GC.
- Transient upstream failures are still silenced rather than reported, including the socket errors native fetch reports differently from axios.
- Updates transport-coupled tests and fixtures that described axios as the live transport.

Closes #206


wa0x6e (commented):
## Review: axios → native fetch migration

Happy path is clean (timeouts ported to `AbortSignal.timeout`, `graphQlCall` return shape preserved, deps removed consistently). The risk is **error semantics**: axios threw on any non-2xx and carried `.code`/message error shapes; native fetch does neither. The code that depended on both wasn't fully ported.

### 🔴 Blocker

- **`src/resolvers/utils.ts` — `fetchHttpImage` no longer checks `response.ok`.** Native fetch resolves on 4xx/5xx, so a 404/500 error body is read as image bytes. On cover/logo paths the buffer returns unvalidated (`src/api.ts` resize) **outside** the resolver fallback, so an expired/rate-limited image URL can bypass fallback and 500 the request. Restore axios semantics: check `response.ok`, throw with status/url on non-2xx. Add a test for `space-cover`/`user-cover` where the upstream returns 404 → assert fallback, not a sharp error.

### 🟡 Major

- **`src/resolvers/starknet.ts` — `fetchImageOrMetadata` no longer checks `response.ok`.** Same class of bug: a non-2xx body is JSON-parsed or fed as image bytes into `resize()`/sharp instead of failing cleanly.
- **`src/addressResolvers/utils.ts` — `isSilencedError` no longer matches native-fetch timeouts.** `AbortSignal.timeout()` rejects with a DOMException `name: 'TimeoutError'`, no `.code`; the `includes('TIMEOUT')` check is case-sensitive and won't match. axios timeouts carried `ECONNABORTED`/`ETIMEDOUT` which did match. Routine upstream timeouts now flood Sentry.
- **`src/addressResolvers/utils.ts:64` — `isSilencedError` never inspects `error.cause`, so socket-level errors escape.** undici surfaces network failures (`ECONNRESET`, `ETIMEDOUT`, `ENOTFOUND`, `ECONNREFUSED`) as a `TypeError` with `message: 'fetch failed'` and the real errno on `error.cause.code`. The `codes` array reads `error.code` / `error.response?.status` / `error.error?.code` but not `error.cause?.code`, and the message is `'fetch failed'`. So `ECONNRESET`/`ETIMEDOUT` — which *are* in the silenced list — no longer match: the migration moved them from `error.code` to `error.cause.code`. Fix: add `error.cause?.code` to the `codes` array.
- **`src/addressResolvers/lens.ts` — `MUTED_ERRORS` no longer matches.** The new FetchError message is `GraphQL request failed with status 503`; the old substrings (`status code 503`/`429`) don't appear, and 503/429 aren't in the silenced-codes list. Lens rate-limit/outage now floods Sentry.
- **`src/lookupDomains/ens.ts` — Sentry context reads `err.response.data`, now always `undefined`.** The new FetchError sets only `.response.status`. Diagnostic regression (optional chaining avoids a crash, but the upstream error body is lost in reports).

### 🔵 Minor / ⚪ Nit

- `src/utils.ts` — new `FetchError(message, status)` collides by name with the empty marker `FetchError` in `src/addressResolvers/utils.ts`. Different shapes, same name → `instanceof`/import hazard. Consider `GraphQlError`.
- `src/utils.ts` — `FetchError.response.status` is set at the throw site but read by no caller. Dead state.
- `5e3` timeout hardcoded in `src/utils.ts` + `src/addressResolvers/starknet.ts` while `src/resolvers/utils.ts` exports `DEFAULT_TIMEOUT` for the same value — three expressions of one timeout.
- `src/resolvers/utils.ts` — removed the `keepAlive` http/https Agents; pooling now relies on undici defaults. Possible handshake overhead under high-volume image resolution.
- `src/utils.ts` — `method: 'post'` → convention is `'POST'` (fetch normalizes it; harmless).

---

**Bottom line:** the blocker plus the three silencing/diagnostic findings are the same root cause — a parity pass is needed over every site that branched on axios error shape (`response.ok` gating + `.code`/message-keyed silencing). Findings 1–3 worth covering with tests since they're silent regressions.

tony8713:
Addressed the review (commit ea0a193):

**Blocker** — `fetchHttpImage` (src/resolvers/utils.ts) now checks `response.ok` and throws on non-2xx, so 404/500 bodies are no longer returned as image bytes; cover/logo paths fall back instead of 500-ing through sharp. Added a unit test for the non-2xx throw.

**Major**
- `fetchImageOrMetadata` (src/resolvers/starknet.ts): added `response.ok` check + throw.
- `isSilencedError` (src/addressResolvers/utils.ts): now inspects `error.cause?.code` (undici socket errors: ECONNRESET/ETIMEDOUT/ENOTFOUND/ECONNREFUSED) and `error.name` (AbortSignal.timeout → TimeoutError), matches codes case-insensitively. Native-fetch timeouts and socket errors are silenced again.
- `addressResolvers/starknet.ts`: thrown error now carries `response.status` so transient 504s stay silenced.
- Lens `MUTED_ERRORS` + ENS Sentry context: the GraphQL error message now includes `status code <n>` and `error.response.data` carries the response body, so MUTED_ERRORS substring matching and Sentry capture work again — no changes needed in those files.

**Minor/nits**
- Renamed `FetchError` → `GraphqlError` in src/utils.ts to avoid the name collision with the marker class in addressResolvers/utils.ts.
- `response.status` is now actually consumed (message + isSilencedError); also added `response.data`.
- Hoisted `DEFAULT_TIMEOUT` to src/utils.ts, re-exported from resolvers/utils.ts — single source for the 5e3 timeout (utils.ts + addressResolvers/starknet.ts now use it).
- `method: 'post'` → `'POST'`.

Added unit tests for the native-fetch error shapes in isSilencedError and the fetchHttpImage non-2xx throw. Left the keepAlive/undici-pooling nit as-is (out of scope; relies on undici defaults). Also updated the PR title to "native fetch" since the code uses global fetch (Node 22), not node-fetch.

tony8713:
Rebased onto master (23 commits of drift). Two things I resolved by judgement rather than mechanically, both worth a second opinion:

**Dropped `ENOTFOUND` and `ECONNREFUSED` from the `isSilencedError` code list** that ea0a193 added. Master's shibarium case "reports a failure that carries no HTTP status" asserts the opposite, and I think master is right: a host that does not resolve is a persistent misconfiguration, not a transient blip, so it should stay visible in Sentry. Genuine socket drops are still silenced through `ECONNRESET`/`ETIMEDOUT` on `error.cause.code`, which master already reads, and nothing in this branch's own coverage depended on the two entries I removed.

**Swapped the shibarium test's transport double.** `test/integration/lookupDomains/shibarium.test.ts` landed on master after this branch was cut and mocks `node-fetch`, which this branch removes. The mock had quietly stopped intercepting anything, so the suite would have gone to the live API; it now doubles the global `fetch`.

Also worth knowing before this merges: master has independently absorbed most of ea0a193's `isSilencedError` work (`error.cause?.code`, cause-message matching, 429), so that commit is now much smaller than it reads. And this PR rewrites `graphQlCall`, which #506 also rewrites, and rewrites `addressResolvers/starknet.ts`, which #491 deletes outright.

Gate after the rebase: typecheck, lint, unit and e2e all pass, and integration matches master's failure set exactly (11, all pre-existing) with no new failures.

src/resolvers/utils.ts
```diff
@@ -1,21 +1,6 @@
...
-      url,
-      ...{
-        responseType: 'arraybuffer',
-        ...axiosDefaultParams
-      }
-    })
-  ).data;
+  const response = await fetch(url, { signal: AbortSignal.timeout(DEFAULT_TIMEOUT) });
+  return Buffer.from(await response.arrayBuffer());
 }
```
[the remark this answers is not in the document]
tony8713: Addressed on the current architecture. fetchHttpImage rejects non-2xx, and the /space-cover/:id regression proves a 404 reaches the fallback response instead of sharp.

src/resolvers/starknet.ts
```diff
@@ -34,13 +33,9 @@ async function getImage(domainOrAddress: string): Promise<string | null> {
...
-  const response = await axios({
-    url,
-    responseType: 'arraybuffer',
-    ...axiosDefaultParams
-  });
-  const contentType: string = response.headers['content-type'] || '';
-  const data = Buffer.from(response.data);
+  const response = await fetch(url, { signal: AbortSignal.timeout(DEFAULT_TIMEOUT) });
+  const contentType: string = response.headers.get('content-type') || '';
+  const data = Buffer.from(await response.arrayBuffer());
```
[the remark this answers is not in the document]
tony8713: Addressed. The moved Starknet image resolver checks response.ok and throws an HTTP error before parsing bytes or metadata.

src/addressResolvers/starknet.ts
```diff
@@ -18,9 +17,17 @@ async function apiCall(
...
+  const requests = needles.map(async needle => {
+    const response = await fetch(buildApiUrl(resolve_type, needle), {
+      signal: AbortSignal.timeout(5e3)
+    });
+
+    if (!response.ok) {
+      throw new Error(`Starknet API request failed with status ${response.status}`);
+    }
+
+    return response.json() as Promise<Record<string, string>>;
```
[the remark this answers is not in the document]
tony8713: This direct Starknet address API path was removed on master before the rebase, so there is no remaining request here. The branch keeps the current snapshot.js-based address architecture.

src/utils.ts
```diff
@@ -183,31 +182,48 @@ export const getBaseAssetIconUrl = (chainId: string) => {
...
+
+class FetchError extends Error {
+  response: { status: number };
+
+  constructor(message: string, status: number) {
+    super(message);
+    this.name = 'FetchError';
+    this.response = { status };
+  }
 }
```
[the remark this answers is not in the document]
tony8713: Addressed. GraphQL non-2xx errors retain response text on error.response.data, with status preserved in both classifier locations.

wa0x6e (changes requested):
**YAGNI: drop `withInactivityTimeout` / `fetchHttpResponse`, keep `withDeadline` everywhere.**

The idle-refresh timer reproduces axios's per-chunk `timeout` semantics, but nothing in prod needs it. Last 7d of `address_resolver_response_duration_seconds` (whole resolver call, fetch included):

| provider | p50 | p99 | share > 5s |
|---|---|---|---|
| Snapshot | 0.19s | 2.27s | 0.01% |
| Ens | 0.37s | 2.23s | 0.26% |
| Space ID | 0.19s | 1.39s | 0.08% |
| Unstoppable | 0.20s | 1.09s | 0.07% |
| Lens | 0.09s | 0.74s | 0.004% |
| Starknet | 0.003s | 0.015s | 0% |

`lookup_domains_response_duration_seconds` p99: Ens 1.53s, Shibarium 0.90s, Unstoppable 0.49s.

Every p99 is under 2.3s against a 5s budget. The ≤0.26% above 5s are hangs, not bodies still streaming at second five — these are small GraphQL/JSON responses. So the case this helper protects (a slow-but-active response that must outlive the budget) does not occur, while the helper costs ~40 lines, a hand-rolled `getReader()` loop, and a second timeout primitive to reason about. It also removes the upper bound entirely: a server trickling one byte every 4s holds the request open forever.

`fetchHttpImage` already uses the whole-call deadline (`44bf25f`); `graphQlCall` and `fetchImageOrMetadata` should too. Suggest reverting `f79a8d4` to `withDeadline(..., 5e3)` at both call sites and dropping `fetchHttpResponse` and the two `http.test.ts` cases. If a real slow-stream case ever shows up in metrics, that is the moment to add it.

tony8713:
I tested the requested `withDeadline(..., 5e3)` replacement against master's axios 0.25 transport using a scaled local stream: one chunk every 100 ms for 1.2 s with a 500 ms budget. Axios completed the 200 response (16 bytes) in 1277 ms, while `withDeadline(fetch + body read, 500)` raised `AbortError` at 514 ms. The replacement changes an inactivity timeout into a total wall-clock cap and would cut off a healthy active response, so I kept `withInactivityTimeout`, `fetchHttpResponse`, and the active-stream regression test. Is that behavior change intentional?

tony8713:
Applied the requested hard 5-second total deadline to GraphQL and Starknet metadata fetches, including full body consumption. Removed the inactivity timer, shared body reader, and its two dedicated tests while preserving BOM parsing, non-2xx rejection, and GraphQL error-body diagnostics. Rerequesting review.

tony8713:
Two follow-ups on the deadline change, from an independent pass over it.

Native fetch reports a peer reset as `TypeError` with `cause.code` set to `UND_ERR_SOCKET`, verified on this Node 22 both before headers and mid-body. That code was not in the silenced list, so the paths this PR moves off axios and node-fetch would have started reporting routine upstream disconnects. Added it, covered by a test that resets the socket.

The new deadline wiring also had no test that could fail: dropping the `signal` from either call site left everything green. Both `graphQlCall` and the Starknet metadata fetch now have a test that streams an incomplete body and expects the abort, and each goes red when its signal is removed.

tony8713:
Test CI on this head is red on Starknet only, and it is not from this branch. Nine cases across `test/integration/resolvers/image/starknet.test.ts` and `test/integration/resolvers/address/starknet.test.ts` fail with the upstream RPC answering `429: Monthly capacity limit exceeded` for `starknet_chainId` and `starknet_call`, so the resolver never gets a profile and returns false. The same commit range passed two hours ago, and I am running master itself through CI as a control to confirm it fails identically. Lint and typecheck are green.

tony8713:
Control confirms it: master `df5f1cf` pushed untouched to a throwaway branch fails the same two Starknet suites with the same nine cases and the same `Monthly capacity limit exceeded` responses, so the red Test run is the upstream RPC quota rather than this branch. That branch is deleted again. Everything outside those two suites passes, and Lint and typecheck are green.

tony8713:
Two more from a review pass over the deadline head, both about tests rather than behaviour.

The lens 503 mute matches on the substring `status code 503`, which since this PR is produced by `graphQlCall` itself, and nothing drove a 503 through it, so rewording that message would have turned every lens outage into a Sentry capture with CI green. There is now a test that drives the 503 through to the resolver and asserts against the resolver's own `MUTED_ERRORS`.

The space-sx envelope case handed one single-use `Response` to both subgraph calls, so the second read failed with `Body is unusable` and only one upstream was really asserted. It now returns a fresh response per call and asserts both are queried. Same shape as the earlier fix in `noData.test.ts`; this call site was missed.

src/resolvers/image/starknet.ts
```diff
@@ -27,13 +28,16 @@ async function getImage(domainOrAddress: string): Promise<string | null> {
...
-    ...axiosDefaultParams
-  });
-  const contentType: string = response.headers['content-type'] || '';
-  const data = Buffer.from(response.data);
+  const { response, data } = await withDeadline(async signal => {
+    const response = await fetch(url, { signal });
+    const data = Buffer.from(await response.arrayBuffer());
+
+    return { response, data };
+  }, 5e3);
```
wa0x6e: The tuple return here exists only because `response.ok` and `content-type` are read after `withDeadline` returns, while the body read has to stay inside it (`fetch()` resolves at headers, so a body read outside the closure would run after `deadline.ts`'s `finally` cleared the timer). Moving the branch inside the closure keeps the same deadline coverage without the two-value return:

```ts
async function fetchImageOrMetadata(url: string): Promise<Buffer | { image?: string }> {
  return withDeadline(async signal => {
    const response = await fetch(url, { signal });

    if (!response.ok) throw httpError(new URL(url).host, response.status, response.statusText);

    return response.headers.get('content-type')?.includes('application/json')
      ? await response.json()
      : Buffer.from(await response.arrayBuffer());
  }, 5e3);
}
```

Two side benefits: the `!response.ok` throw now happens *before* the body is read, so an upstream answering a 500 with a large HTML page no longer spends the 5s budget downloading bytes that are immediately discarded; and `response.json()` does the spec UTF-8 decode, replacing the `Buffer` → `toString('utf-8')` → `JSON.parse` round-trip.

Both existing tests in `test/unit/resolvers/image/starknet.test.ts` still hold by construction: the 404 case throws on `!ok` before any read, and the incomplete-body case sets `Content-Type: application/json`, so `response.json()` blocks on the stream until the deadline aborts it — same `AbortError`.

(Posted as a plain comment rather than a suggestion because the replacement spans lines 30-45 and this diff hunk ends at line 43.)

src/helpers/errors.ts
```diff
@@ -42,7 +42,7 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
         error.error?.message?.includes(m) ||
         error.cause?.message?.includes(m)
     ) ||
-    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET', 504, 429].some(c =>
+    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET', 'UND_ERR_SOCKET', 504, 429].some(c =>
```
wa0x6e: While this list is being touched: `ECONNABORTED` was axios's own timeout code (it threw `ECONNABORTED` with "timeout of 5000ms exceeded"). undici never emits it, and the only remaining emitters in `node_modules` are `superagent` (test-only, via supertest) and `express`/`raw-body`, which raise it for *inbound* request aborts on the server side — not for the upstream failures this classifier sees. Nothing in `src/` or `test/` asserts it either, so it can go with the dependency:

```suggestion
    ['TIMEOUT', 'ETIMEDOUT', 'ECONNRESET', 'UND_ERR_SOCKET', 504, 429].some(c =>
```

For reference, the entries that stay are all still reachable: `TIMEOUT` comes from ethers (`@ethersproject/web` rejects with `Logger.errors.TIMEOUT`) via every `getProvider()` call, `ETIMEDOUT`/`ECONNRESET` are OS errnos still surfaced through ethers' node http path and undici's `cause.code`, and `504`/`429` are the statuses `httpError()` sets.

src/helpers/graphql.ts
```diff
@@ -13,29 +13,55 @@ function graphQlEnvelopeError(url: string, status: number, message: string) {
...
+  }
+}
+
 export async function graphQlCall<T = any>(
   url: string,
   query: string,
   variables?: Record<string, any>,
   options: any = { headers: {} }
-): Promise<AxiosResponse<GraphQlResponse<T>>> {
+): Promise<{ data: GraphQlResponse<T>; status: number }> {
```
wa0x6e: Now that `AxiosResponse` is gone, the outer envelope has no reader. All nine call sites unwrap it identically and none touches `status`:

```ts
const {
  data: {
    data: { accountsBulk }
  }
} = await graphQlCall(API_URL, query, …);
```

(`src/resolvers/address/{ens,lens,snapshot}.ts`, `src/resolvers/image/{lens,snapshot×2,space-sx}.ts`, `src/resolvers/lookupDomains/ens.ts` — grepping `.status` across `src/resolvers/` turns up only `Promise.allSettled` results and native `Response` objects on the non-`graphQlCall` paths.)

Returning the body directly drops one level everywhere:

```ts
): Promise<GraphQlResponse<T>> {
  …
  return body;
}
```

and each call site becomes `const { data: { accountsBulk } } = await graphQlCall(…)`. `status` stays in use inside this file for `graphQlEnvelopeError`, it just isn't part of the contract.

Cost is nine call sites plus their tests, so it's a fair thing to defer — but this PR is where it's cheapest, since the return type is already changing and every call site is already being re-verified. Left as a comment rather than a suggestion because the change spans both the signature (line 30) and the return (line 75), which are in different hunks.

src/helpers/graphql.ts
```diff
@@ -13,29 +13,55 @@ function graphQlEnvelopeError(url: string, status: number, message: string) {
...
+        ...Object.fromEntries(
+          Object.entries(options.headers).filter(
+            ([, value]) => value !== undefined && value !== null
+          )
+        )
+      },
+      body: JSON.stringify(data),
+      signal
+    });
+    const responseBody = Buffer.from(await response.arrayBuffer());
```
wa0x6e: The `arrayBuffer()` → `Buffer` → `toString()` round-trip is what makes the BOM strip in `parseGraphQlResponse` necessary, and nothing here needs bytes — both consumers (`error.response.data` and `parseGraphQlResponse`) call `.toString()` immediately.

`Response.text()` runs the WHATWG decode, which removes a leading BOM; the Buffer path keeps it. Verified on Node 22:

```
JSON.parse(BOM + json)     -> SyntaxError: Unexpected token '﻿' ... is not valid JSON
Response.text() first char -> "{"
Buffer path first char     -> "﻿"
```

So switching to `text()` deletes the hand-rolled strip rather than replacing it:

```diff
 function parseGraphQlResponse(body: string): any {
-  const decoded = body.startsWith('﻿') ? body.slice(1) : body;
   try {
-    return JSON.parse(decoded);
+    return JSON.parse(body);
   } catch {
     return body;
   }
 }
@@
       body: JSON.stringify(data),
       signal
     });
-    const responseBody = Buffer.from(await response.arrayBuffer());
+    const responseBody = await response.text();
@@
     );
-    error.response.data = responseBody.toString();
+    error.response.data = responseBody;
     throw error;
   }
 
-  const body = parseGraphQlResponse(responseBody.toString());
+  const body = parseGraphQlResponse(responseBody);
```

The `JSON.parse` try/catch still earns its place (a non-JSON body has to fall through to the "no data envelope" error), it just no longer needs the BOM branch. The new BOM test in `graphql.test.ts` keeps passing.

One aside worth recording in the PR description either way: BOM tolerance is *new* behaviour, not preserved behaviour — under axios a BOM-prefixed body also failed to parse (its default `transformResponse` decodes to a string and `JSON.parse`s it the same way), so it would have surfaced as "no data envelope" too.

(Plain comment rather than a suggestion because the change touches four non-contiguous lines: 17, 48, 59 and 63.)

src/helpers/graphql.ts
```diff
@@ -13,29 +13,55 @@ function graphQlEnvelopeError(url: string, status: number, message: string) {
...
+    return { response, responseBody };
+  }, 5e3);
+
+  if (!response.ok) {
+    const error: any = graphQlEnvelopeError(
+      url,
+      response.status,
+      `status code ${response.status}: ${response.statusText}`
+    );
+    error.response.data = responseBody.toString();
```
wa0x6e: `error.response.data` is set only on the non-2xx path. The two envelope errors thrown further down (`body.errors` at line 66, `!body?.data` at line 73) come from the same helper and the same error type but carry no `data`, so a caller that reads `.response.data` gets a string on one failure and `undefined` on the other two.

Nothing reads it today — `isSilencedError` reads `error.response?.status`, `isRoutineMiss` reads `status`/`response.status`, and `@snapshot-labs/snapshot-sentry` initialises without `extraErrorDataIntegration`, so arbitrary own properties on the Error don't reach Sentry either. Only the new unit test asserts it. So there are two coherent endings, and it's worth picking one deliberately:

**Make it consistent** — thread `data` through the helper so all three throw sites attach it:

```diff
-function graphQlEnvelopeError(url: string, status: number, message: string) {
+function graphQlEnvelopeError(url: string, status: number, message: string, data?: string) {
@@
-  return httpError(source, status, message);
+  const error = httpError(source, status, message);
+  if (data !== undefined) error.response.data = data;
+  return error;
@@
   if (!response.ok) {
-    const error: any = graphQlEnvelopeError(
+    throw graphQlEnvelopeError(
       url,
       response.status,
-      `status code ${response.status}: ${response.statusText}`
+      `status code ${response.status}: ${response.statusText}`,
+      responseBody.toString()
     );
-    error.response.data = responseBody.toString();
-    throw error;
   }
```

(plus the same extra argument at the other two throw sites).

**Or drop it** — the PR description says the change "preserves non-2xx status and body context for callers", but no caller consumes that context, so the simpler ending is to delete the assignment and the test that pins it, and add the body to the Sentry `capture` call if the context is actually wanted.

I'd lean on the second unless you know of a consumer; the first is right if `.response.data` is meant to become part of the contract.

Note this touches the same lines as the `response.text()` comment above — if both land, `responseBody.toString()` here is just `responseBody`.

test/unit/api.test.ts
```diff
@@ -20,11 +20,37 @@ jest.mock('../../src/resolvers/getOwner', () => ({
...
+    const response = await request(app).get(`/space-cover/${ADDRESS}`);
+
+    expect(response.status).toBe(200);
+    expect(response.headers['content-type']).toMatch(/^image\/webp/);
+    expect(response.headers['cache-control']).toBe('public, max-age=3600');
+    expect(response.body.length).toBeGreaterThan(0);
+  });
+});
+
 describe('POST /', () => {
```
wa0x6e, line 54: Anchored here because the lines this is about sit just below the hunk, plus in a file this PR doesn't touch.

Now that `axios` is out of `package.json` (and out of `yarn.lock` entirely — nothing pulls it transitively), a few fixtures still describe it as a live transport:

**`test/unit/api.test.ts:58-60`** — `new Error('Request failed with status code 500')` as the `graphQlCall` rejection. `graphQlCall` now throws `httpError`, whose message is `[hub.snapshot.org] status code 500: Internal Server Error`. Behaviourally the test still passes (both strings contain `status code 500`, which is what the message matching keys on), but the fixture no longer resembles what the code under test actually produces.

**`test/integration/helpers/errors.test.ts:32-110`** — three cases named `axiosError`, carrying `code: 'ERR_BAD_RESPONSE'` / `'ERR_BAD_REQUEST'` and messages like `'Request failed with status code 504'`. Those codes are axios-specific and nothing in the tree emits them anymore. The cases do still cover a live branch — silencing there comes from `response.status`, which `httpError` sets — so they aren't dead tests, just tests documenting a transport that's gone. The `// Shape observed from Sentry STAMP-7` comment above the 504 case explicitly attributes the shape to axios.

Cheapest fix is a rename plus a message swap to the `httpError` shape, keeping the same assertions:

```ts
const upstreamError = {
  message: '[hub.snapshot.org] status code 504: Gateway Timeout',
  status: 504,
  response: { status: 504 }
};
```

Not blocking, and reasonable to leave for a follow-up — but if it's left, the STAMP-7 comment is worth updating so the next reader doesn't go looking for the axios call site that produced it.

wa0x6e (changes requested):
Migration itself looks sound — native `fetch` behind the existing `withDeadline` seam finishes a consolidation the codebase had already started, all nine `graphQlCall` call sites were migrated, and the lens `MUTED_ERRORS` phrasing was preserved and pinned by a test.

Comments are simplifications and leftovers rather than blocking defects: the axios-shaped `{ data, status }` return and `error.response.data` have no readers, `response.text()` would delete the hand-rolled BOM strip, `ECONNABORTED` is vestigial now that axios is out of the lockfile entirely, the starknet error body is downloaded before being discarded, and a few test fixtures still describe axios as a live transport.

One thing deliberately not raised as a change request: the 5s budget switching from socket-inactivity to total wall-clock semantics. Confirmed real, but `6808c0e` shows it was a considered decision, it matches `fetchHttpImage`, and the payloads involved are small JSON envelopes. Worth naming in the PR description, since "keeps request deadlines" reads as no change in kind.

src/resolvers/image/starknet.ts
tony8713: Applied as written. The tuple is gone, the `!ok` throw now happens before any body read, and `response.json()` replaces the decode round-trip. Both existing tests still hold for the reasons you gave, and I mutation-checked the guard again on the new shape: deleting the `!ok` throw fails the 404 test.

src/helpers/errors.ts
tony8713: Dropped. I confirmed the same thing independently before removing it: the only occurrence left in `src/` or `test/` was the list entry itself, and the remaining emitters under `node_modules` are `express`/`raw-body` for inbound aborts plus test-only `superagent`. Those never reach this classifier, since `isSilencedError` is only called on upstream resolver errors while `src/api.ts` captures inbound failures directly. Removing `504` from the list still fails the transient-status tests, so the rest of the list is still pinned.

src/helpers/graphql.ts
tony8713: Done, and it was nine call sites rather than the eight my own sweep found: `lookupDomains/ens.ts` calls it twice. Every one now destructures a single level. I mutation-checked it by re-wrapping the return, which fails the three `graphQlCall` shape tests plus the snapshot and ENS resolver tests, so the contract is pinned at the call sites and not just in the helper.

src/helpers/graphql.ts
tony8713: Switched to `response.text()` and deleted the strip. I reproduced your Node 22 check first: through `text()` the first code unit is `7b`, through the Buffer path it is `feff`, and only the former parses. Swapping `text()` back for the Buffer round-trip now fails the BOM test, so the decode is what that test pins. The PR description now says BOM tolerance is new behavior rather than preserved.

src/helpers/graphql.ts
tony8713: Took the second ending and dropped it, along with the assertion that pinned it. I checked the Sentry half myself rather than take it on trust: `initLogger` calls `Sentry.init` with no `integrations` key, so own properties on the error never leave the process. With the body no longer needed on the failure path, the `!ok` throw also moved inside the closure, so a non-2xx no longer spends the budget downloading a page that gets discarded.

test/unit/api.test.ts
tony8713: Reshaped both files rather than leaving them: the `graphQlCall` rejections in `api.test.ts` now carry the `httpError` message and shape, and the three cases in `errors.test.ts` are renamed off `axiosError` with matching messages. The STAMP-7 comment kept the part that is still true, which is why the `||` chain was wrong, and lost the axios attribution. I checked the reshaped fixtures still discriminate: removing `504` from the silenced list fails them. Left `address.test.ts` and `failureContract.test.ts` alone, since those carry the same axios-shaped strings but sit outside this diff.

src/helpers/graphql.ts
tony8713: Correcting one thing from my earlier reply here: the aside about axios is wrong, and I put it in the PR description before checking it, which was my mistake.

axios 0.25 strips a leading BOM itself, in `stripBOM` at `lib/utils.js:319`, and its default `transformResponse` runs that before `JSON.parse`. A live request against a BOM-prefixed body parses cleanly and yields the object, on the same axios version this PR removes.

So BOM tolerance is preserved behavior, not new. What actually happened is that the Buffer round-trip lost it and the hand-rolled strip put it back; `response.text()` is the tidier way to keep it, which is why the suggestion still stands. The description now says preserved rather than new.

tony8713:
Another review pass, and one of its findings was a test that could not fail.

`test/integration/resolvers/lookupDomains/shibarium.test.ts` had no `mockReset` between cases, so with `clearMocks` the queued `mockResolvedValueOnce` responses survived into the retries that `jest.retryTimes(3)` runs. Forcing `hasMore = false`, which removes pagination altogether, left the file at 14 of 14 green, including both cases that claim to check what happens when a later page fails. With the reset added, that same mutation reddens exactly those two. The file predates this PR but the PR rewrites its mocking, and `noData.test.ts` already had the reset.

Also deduped the stalled-body fixture the diff had added twice into `test/helpers/stalledResponse.ts`, and dropped the STAMP-7 comment I had trimmed rather than deleted last round: it narrated a past change and described a `||` chain that no longer exists.

One finding I am not acting on. It called the fourth `options` argument and its header filter dead, but `src/resolvers/address/snapshot.ts:19` and `src/resolvers/image/snapshot.ts:63` both pass `x-api-key` through it, and removing the parameter fails typecheck at exactly those two lines. The undefined filter earns its place too: `HUB_API_KEY` ships empty in `.env.example`, and without it an unset value goes out as the literal string `undefined` instead of being omitted.

test/integration/helpers/errors.test.ts
```diff
@@ -29,18 +29,14 @@ describe('isSilencedError', () => {
...
-    // The previous `||` chain short-circuited on error.code and never
-    // reached error.response.status.
-    const axiosError = {
-      message: 'Request failed with status code 504',
-      code: 'ERR_BAD_RESPONSE',
+  it('silences a 504 carried on error.response', () => {
+    const upstreamError = {
+      message: '[hub.snapshot.org] status code 504: Gateway Timeout',
+      status: 504,
       response: { status: 504 }
```
wa0x6e: **This fixture reshape deleted the only regression test for the `error.response.status` read.**

My suggestion last round handed over a fixture carrying both `status` and `response.status`. That was my mistake: with both present, nothing discriminates which one `isSilencedError` actually reads — and `src/helpers/errors.ts:1-2` calls both load-bearing, with STAMP-7 being specifically about the `||` chain never reaching `error.response.status`.

Measured by deleting one slot at a time from the `codes` array (`src/helpers/errors.ts:30-37`) and running `test/unit` + `test/integration/helpers` (243 tests):

| slot | red at this head | red at `df5f1cf` |
|---|---|---|
| `error.error?.code` | 0 | 0 (pre-existing) |
| `error.error?.status` | 1 | — |
| `error.code` | 0 | 0 (pre-existing) |
| `error.status` | 3 | — |
| **`error.response?.status`** | **0** | **2** |
| `error.cause?.code` | 3 | — |

The merge-base fixture (`code: 'ERR_BAD_RESPONSE'` + `response: { status: 504 }`, no top-level `status`) caught that mutation. This one does not.

Dropping the redundant field restores it — `error.status` stays pinned by `lookupDomains.test.ts` and the `httpError` path, so no case is lost:

```suggestion
      response: { status: 504 }
```

Same edit at lines 92-93 for the `silences a 429 carried on error.response` case.

While you are in here: `error.code` and `error.error?.code` have no regression test either. That is pre-existing (confirmed unpinned at `df5f1cf` too), so not a change request — but it is two more fixtures in the file you are already editing if you want the matrix closed in one pass.

src/helpers/graphql.ts
```diff
@@ -13,38 +13,59 @@ function graphQlEnvelopeError(url: string, status: number, message: string) {
...
+          Object.entries(options.headers).filter(
+            ([, value]) => value !== undefined && value !== null
+          )
+        )
+      },
+      body: JSON.stringify(data),
+      signal
+    });
+
+    if (!response.ok) {
```
wa0x6e, line 48: **Throwing before reading the body costs a fresh TCP connection per non-2xx, and strands the old socket until GC.**

Also my suggestion's doing — both round-9 comments asked for the `!ok` throw to move ahead of the body read, here and in `src/resolvers/image/starknet.ts:34`. At `df5f1cf` both paths went through axios, which buffers the whole response before rejecting, so the connection went back to the keep-alive agent. `fetchHttpImage`/`farcaster`/`coingecko` already had this shape on master; these two acquired it in `e48d9ad`.

Measured on Node 22.23.2 — 20 sequential requests to a local keep-alive server answering 404 with a 64 KB body, `withDeadline` reproduced verbatim:

| variant | sockets opened | alive immediately | alive after GC + 7s |
|---|---|---|---|
| this head (throw before read) | 40 | 20 | 1 |
| `await response.arrayBuffer()` first | **2** | 2 | 0 |
| `await response.body?.cancel()` first | 22 | 2 | 0 |
| `controller.abort()` in `withDeadline`'s `finally` | 22 | 2 | 0 |

Not an unbounded leak — GC reclaims them — but during an upstream outage every failed request pays a fresh TCP + TLS handshake and holds an extra fd until the next collection. A lens 503 storm is exactly the case `MUTED_ERRORS` exists for.

The current state is strictly worse than the cancel/abort rows: same loss of reuse, plus the lingering sockets. Fix belongs at the seam rather than at each call site, so all eight `fetch` sites in `src/` get it — `src/helpers/deadline.ts`, which this PR does not otherwise touch:

```diff
   } finally {
     clearTimeout(timeout);
+    // Cancels an un-consumed response body: every caller throws on !ok before
+    // reading it, which otherwise strands the socket until GC.
+    controller.abort();
   }
```

Safe on the success path — measured, 20 successful requests open 2 sockets with or without it, because the body is fully consumed before `finally` runs. It does not restore *reuse* on the error path; only draining does, which is what axios did. My read is take the `abort()` and leave reuse on the floor: error volume is low and the argument for not spending the 5s budget downloading a page you discard still holds. If you disagree, the alternative is putting the body read back ahead of the `!ok` check and undoing that half of my suggestion — either ending is fine, but the current one is neither.

src/resolvers/image/starknet.ts
```diff
@@ -27,17 +28,15 @@ async function getImage(domainOrAddress: string): Promise<string | null> {
...
-  return data;
+  return withDeadline(async signal => {
+    const response = await fetch(url, { signal });
+
+    if (!response.ok) throw httpError(new URL(url).host, response.status, response.statusText);
+
+    return response.headers.get('content-type')?.includes('application/json')
+      ? await response.json()
+      : Buffer.from(await response.arrayBuffer());
+  }, 5e3);
```
wa0x6e: The 5s budget now has two literals and one constant: this line, `src/helpers/graphql.ts:70`, and `IMAGE_FETCH_BUDGET` at `src/helpers/http.ts:20`.

This is the "three expressions of one timeout" nit from the June review, and it went slightly backwards here — at `df5f1cf` this call site read the value from the shared `axiosDefaultParams.timeout`, so the PR turned one shared constant into a second literal.

```diff
-const IMAGE_FETCH_BUDGET = 5e3;
+export const IMAGE_FETCH_BUDGET = 5e3;
```

then import it here and at `graphql.ts:70`. Or hoist it to `deadline.ts` next to `UPSTREAM_TIMEOUT`, which is where the other budget already lives — either is fine, one place is the point.

test/integration/resolvers/image/deadline.test.ts
```diff
@@ -38,7 +28,7 @@ beforeAll(async () => {
   });
   server.on('connection', socket => sockets.add(socket));
   await new Promise<void>(resolve => server.listen(0, '127.0.0.1', () => resolve()));
-  mockHangingUrl = `http://127.0.0.1:${(server.address() as AddressInfo).port}/`;
+  const mockHangingUrl = `http://127.0.0.1:${(server.address() as AddressInfo).port}/`;
```
wa0x6e, line 31: Anchored here because the lines this is about were deleted, just above the `describe` at line 52.

`c1619e4` removed the comment block explaining why these cases wait out the real deadline instead of reaching in and firing the timer by hand — that reaching in would abort the signal even where the code under test had already cleared it, which is exactly the case they need to be able to fail on.

Only the `node-fetch` sentence in it went stale; the rest is transport-neutral and it is the invariant that stops the next person "speeding this suite up" with fake timers and silently making it unable to fail. The sibling comment removed from `farcaster.ts` in `3cf3edd` was a duplicate — `coingecko.ts:27-29` still carries it — but this one has no other copy.

Worth restoring without the two-transports wording.

wa0x6e (changes requested):
Re-verified every item raised on this PR from the code at `570b61b` rather than from the replies: 22 distinct issues across the four rounds, **19 fixed, 1 fixed with a caveat, 1 nit not fixed, 1 that was never a real problem.** `tsc --noEmit` clean, 243/243 unit + helper-integration tests green, and each fix mutation-checked — reverting it in a scratch edit and confirming a test actually reddens.

Spot checks worth recording, since they answer things the replies asserted:

- **The BOM correction is right and my original aside was wrong.** axios 0.25 has `stripBOM` at `lib/utils.js:319`, applied in `lib/adapters/http.js:305` for any non-`arraybuffer` responseType — read at the pinned tag, not from memory. BOM tolerance on `graphQlCall` is preserved behaviour. The description now says so.
- **Nine `graphQlCall` call sites, not eight** — your count was the correct one; `lookupDomains/ens.ts:31` is the one my own grep would have missed under the generic.
- **The `!response.ok` class is closed repo-wide**, not just at the two reported sites: all eight `fetch(` call sites in `src/` now guard it.
- **The lens mute chain is pinned at both ends** — `graphqlEnvelope.test.ts:56` drives a real 503 through the resolver, and removing `r.MUTED_ERRORS` from `address/index.ts:72` reddens `address.test.ts`, so the wiring is covered too.
- **`error.response.data` was safe to drop** — there is no reader of `response.data` anywhere in `src/`, confirmed by grep.
- **Dropping the keep-alive agents costs nothing on the happy path** — measured, 20 sequential 200s open 2 sockets under undici's default pool. That June nit is resolved by measurement. The non-2xx path is a different story; see the `graphql.ts` comment.
- **The red Test check is not this branch.** Nine failures, all in the two Starknet integration suites, all `429: Monthly capacity limit exceeded` from the Alchemy RPC. Those go through `getProvider`, not any `fetch` path this PR touches. Your control run matches what the job log shows.

Four comments below. Both majors are consequences of my own round-9 suggestions rather than deviations from them — one traded away a regression test, the other traded away connection reuse — so treat them as corrections to the suggestions, not as new asks.

Two pre-existing things found while verifying, neither belonging in this PR:

- `image/index.ts`'s `withFailureContract` calls `isSilencedError(err)` with no second argument, while `address/index.ts:72` passes `r.MUTED_ERRORS`. So a lens 503 is muted on the address path and captured on the image path, from the same upstream. Unchanged here — neither message matched at `df5f1cf` either. I will open a follow-up issue.
- `error.code` and `error.error?.code` have no regression test in `isSilencedError`, confirmed unpinned at `df5f1cf` too. Cheap to fold into the `errors.test.ts` edit if you want the whole matrix closed in one pass, but not a change request.

test/integration/helpers/errors.test.ts
tony8713: Confirmed and fixed. I reproduced your table row before touching anything: with `error.response?.status` deleted from the `codes` array, `test/unit` plus `test/integration/helpers` was 243 of 243 green, so nothing pinned that slot. Dropping the redundant top-level `status` from both fixtures restores it, and the same deletion now reddens exactly those two cases.

I also took the optional half, since the file was open: two fixtures for `error.code` and `error.error?.code`. Each was mutation-checked on its own slot, and each reddens only its own case.

src/helpers/graphql.ts
tony8713: Reproduced, and took the `abort()`. My run on Node 22.23.1, 20 sequential 404s with a 64 KB body: throwing before the read opened 40 sockets and left 20 alive, `arrayBuffer()` first opened none beyond the pool, and both `body.cancel()` and `controller.abort()` in `finally` opened 20 and left 2. That matches your table, so it is at the seam in `deadline.ts` and all eight `fetch` sites get it.

On the comment: I tried the stronger version first, a test that fails when the `abort()` is deleted, and it would not hold. Four shapes, counting sockets opened, counting the alive delta, and tracking only the connections the loop itself opened, all drifted between runs because the undici pool is shared with the other cases in that file. One diagnostic run measured 8 stranded against 0, and the very next run of the same assertion passed with the abort removed. A test that passes when the code is broken is worse than no test, so I dropped it and kept a two-line comment, recorded through `comment-gate --allow` rather than slipped in.

src/resolvers/image/starknet.ts
tony8713: Hoisted rather than exported: `FETCH_BUDGET` now lives in `src/helpers/deadline.ts` next to `UPSTREAM_TIMEOUT`, and `http.ts`, `graphql.ts` and `image/starknet.ts` all import it. `IMAGE_FETCH_BUDGET` is gone, since the name had stopped being true once GraphQL used the same value. The only `5e3` left in `src/` is `DEFAULT_PROVIDER_TIMEOUT` in `provider.ts`, which is the RPC provider timeout and a different budget.

test/integration/resolvers/image/deadline.test.ts
tony8713: Restored above the `describe`, minus the two-transports wording. Kept the part that is still the invariant: why these wait out the real deadline instead of firing its timer by hand, and why the body case needs the deadline to cover more than the request. Also through `comment-gate --allow`, so it is a recorded decision rather than a comment that slipped past.

tony8713:
A review pass told me I was wrong about the abort not being pinnable, and it was right.

I had concluded a test could not hold it and kept a comment instead. Three things made my attempts drift, and only one was timing: undici pools per origin, so the other cases in the file were taking and leaving sockets; below about 32 KiB there is nothing to measure at all, because the body is already buffered and the socket already returned; and `jest.retryTimes(3)` gives every case four attempts, so a one-in-four flaky assertion reads as green. That last one is what made "the next run passed with the abort removed".

`test/integration/helpers/socketRelease.test.ts` now asserts on the server side of a single request, event-driven rather than sampled: its own server on its own port, a `Content-Length` far larger than what it writes, and `res.on(close)` checking `writableFinished`. Deleting `controller.abort()` reddens it, plus the two new `withDeadline` unit cases. The comment on that line is gone, which is the better ending.

Also closed a gap the same pass found: deleting the entire non-2xx branch in `graphQlCall` left every case in its own test file green, because every body those tests send is envelope-less too, so the second throw site produced the same status and host. A 500 carrying a well-formed `{ data: ... }` now has a case, and it reddens when that branch goes.

Not acting on the third one, the axios-shaped fixtures still in `failureContract.test.ts` and `address.test.ts`. Same call as last round: those files are outside this diff, and I would rather not widen it further.

tony8713:
Two more from a review pass, both about tests rather than behaviour, and one of them was my own mistake.

The socket test I added last round asserted something that could never be false. Its handler never calls `res.end()`, so `res.writableFinished` is permanently false and `!res.writableFinished` was a constant. The real discriminating half was always whether `close` fires inside the 2s cap, so that is all it asserts now. It still reddens when `controller.abort()` is deleted.

The larger one: `graphQlCall` had no test running against a real socket, and the `finally` abort quietly made the position of `await response.text()` load-bearing. Hoisting that read out of the `withDeadline` callback breaks every GraphQL call against a real server, and all 19 cases in its own unit file stayed green under that mutation, because a `Response` built from an in-memory string is not wired to the abort signal the way undici body is. There is now an integration test against a local upstream, and it is the only thing that catches the hoist.

Also took two tidies: collapsed a redundant ternary in the shared test helper, after checking the 204 case behaves identically either way, and inlined a single-caller helper in `abortClassification.test.ts` to the `.catch` idiom its sibling already used.

src/helpers/graphql.ts
```diff
@@ -13,38 +13,59 @@ function graphQlEnvelopeError(url: string, status: number, message: string) {
...
+        response.status,
+        body.errors[0]?.message || 'GraphQL request failed'
+      );
+    }
+    if (!body?.data) {
+      throw graphQlEnvelopeError(url, response.status, 'GraphQL response has no data envelope');
+    }
+
+    return body;
+  }, FETCH_BUDGET);
```
wa0x6e: **Correction to [my comment on `image/starknet.ts`](https://github.com/snapshot-labs/stamp/pull/457#discussion_r3829630179) from yesterday's review — do not do what it asked.** It told you to export `IMAGE_FETCH_BUDGET` and import it here and in `image/starknet.ts`. That is the wrong ending; take this one instead.

Two reasons.

**The constant is not carrying its weight even where it lives.** `IMAGE_FETCH_BUDGET` (`src/helpers/http.ts:20`) is used exactly once, at line 29, is not exported, and no test references it. The name it supplies is already supplied twice over — the enclosing function is `fetchHttpImage`, and `withDeadline`'s signature already calls the parameter `budget`. `}, 5e3);` inside `fetchHttpImage` reads no worse than `}, IMAGE_FETCH_BUDGET);`.

**The three 5s are not one value.** They are three that happen to be equal, and only for a historical reason: axios had a single `axiosDefaultParams.timeout = 5e3` applied to every request, and this migration carried that number to each site unchanged. Nobody derived 5s for a GraphQL POST. The workloads differ —

- `fetchHttpImage`: an arbitrary third-party image, possibly large, often through an IPFS gateway rewrite
- this line: a small JSON envelope POSTed to hub or a subgraph
- `image/starknet.ts:39`: a third-party image *or* a metadata JSON

Sharing one constant across those couples three independent knobs. Whoever later needs to raise the image budget for a slow gateway raises the GraphQL budget with it, silently and without noticing.

So: **delete `IMAGE_FETCH_BUDGET` and leave three bare `5e3`.** Smaller diff than what I originally asked for, and it stops pretending the three move together.

`UPSTREAM_TIMEOUT` (`src/helpers/deadline.ts:1`) stays as it is — also single-use, but it earns the name because it is the *default* in a signature, where a bare `10000` would be a magic number with no caller context to read it against. Worth keeping the two apart for a different reason too: `UPSTREAM_TIMEOUT` bounds a whole operation, which at `lookupDomains/shibarium.ts:24` and `lookupDomains/unstoppableDomains.ts:52` is a pagination loop of N requests, and at `getOwner/shibarium.ts:53` is not a fetch at all. These three bound one round trip. Collapsing all four into one number would either give `fetchHttpImage` a 10 s single-request budget or give the shibarium loop 5 s for all its pages.

What stands from the original comment is only the observation underneath it: one site named and two bare is the incoherent middle, since it implies the named one is special. Either ending fixes that; this is the cheaper one.

wa0x6e (commented):
Correcting one finding from my review yesterday rather than leaving it to be implemented.

**The comment on [`image/starknet.ts`](https://github.com/snapshot-labs/stamp/pull/457#discussion_r3829630179) — the 5 s budget one — was wrong.** It asked for the three expressions of that budget to be consolidated behind one exported constant. They should be three bare `5e3` literals and `IMAGE_FETCH_BUDGET` should be deleted — reasoning inline on `helpers/graphql.ts`. Shorter diff, and it avoids coupling three budgets that are only equal by accident of axios having had one `timeout` for everything.

Two corrections to this comment itself, both mine:

- An earlier version of it referred to "Minor 3", which is numbering from my own notes and appears nowhere on this PR. It meant the `image/starknet.ts` comment linked above.
- It also said the head was unchanged at `570b61b`. That was stale — `a35eef4`, `a566f2c` and `a77bdff` landed after my review, and `a35eef4` reads as a fix for the socket finding. I had not re-checked before writing it.

So treat the two majors from yesterday as **not yet re-verified against `a77bdff`** rather than as still standing. I am re-checking them now and will follow up with the verdicts. The correction above is independent of the head — `IMAGE_FETCH_BUDGET` is a master-era constant and none of the new commits touch it.

Not re-requesting changes on this one.

tony8713:
Rebased onto master to clear the conflict that #588 introduced when it landed the Starknet address validation. Head moved from `a77bdff` to `53e11b4`, so the two majors you are re-checking are pinned to a SHA that is no longer on the branch.

The conflict was confined to `src/resolvers/image/starknet.ts` and `test/unit/resolvers/image/starknet.test.ts`. The resolution keeps master's `isStarknetFelt` guard and all four of its test cases, and replays this PR's fetch migration on top. All 19 commits survived and no behaviour changed.

Your `IMAGE_FETCH_BUDGET` correction is not addressed here, since this was a conflict fix only.

src/helpers/graphql.ts
tony8713: Done. `FETCH_BUDGET` is deleted from `src/helpers/deadline.ts` and both call sites now carry a bare `5e3`: `graphql.ts:70` and `http.ts:30`, inside `fetchWithDeadline`.

Two sites rather than three. On this head `image/starknet.ts` fetches its image-or-metadata through `fetchWithDeadline`, so it reads `http.ts`'s literal and there is no third one to inline. The image and metadata budgets are therefore still a single number, but that is one function being one function, not a constant implying three independent knobs move together. `UPSTREAM_TIMEOUT` is untouched, for the reason you gave.

tony8713:
Before you finish that re-check: the head has moved twice since `a77bdff`. First to `53e11b4`, a rebase onto master to clear the conflict #588 introduced in `src/resolvers/image/starknet.ts`, and then to `6a96976`, which deletes `FETCH_BUDGET` from `src/helpers/deadline.ts` and inlines a bare `5e3` at both call sites, per your correction.

`a77bdff` is off the branch now, so pull `6a96976` rather than re-verifying the two majors against a SHA that no longer exists on it.

wa0x6e (commented):
utAck

wa0x6e (approved):

PR #459, opened 2026-06-10
chore: bump sharp to 0.34.5 (latest stable, binary as optional dep, no postinstall download)

tony8713:
sharp was pinned `^0.30.1`. On 0.30.x the native binary is fetched in a postinstall step, which times out flakily on CI and causes intermittent install failures.

sharp 0.33+ ships platform binaries as `@img/sharp-*` npm **optional dependencies**, so there is no postinstall download step at all - the right prebuilt binary is just resolved from the lockfile. Bumped to the current latest stable, `0.34.5` (0.35.x is still RC-only), per @wa0x6e's request for the latest version.

## Scope (3 files)

- `package.json`: sharp `^0.30.1` -> `^0.34.5`
- `yarn.lock`: now carries the `@img/sharp-*` optional deps at `0.34.5` (e.g. `@img/sharp-darwin-arm64`, the linux/musl/win32 variants, and `@img/sharp-libvips-*` at `1.2.4`)
- `src/utils.ts`: annotate `resize()` as `Promise<Buffer>`. sharp 0.34.5's types resolve cleanly to `Buffer`, so no `any` cast is needed - this just makes the inferred return type explicit.

## Base

Based on #456 (`test/resolver-image-snapshots`), not master. sharp 0.34.5's stricter `toBuffer()` types make the resolver return type `Buffer | false`; the resolver image-snapshot test changes on #456 are what keep `tsc` green under the new types. Bumping sharp directly on master (with the old-style `result.length` tests) does not typecheck. Merge #456 first, then this.

## Node compat

`engines.node` is `>=22.6.0`, CI runs on Node 22. sharp 0.34.5 requires Node `^18.17.0 || ^20.3.0 || >=21.0.0`, so Node 22 is well within range.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "node-fetch": "v2.7.0",
     "nodemon": "^2.0.7",
     "redis": "^4.6.10",
-    "sharp": "^0.30.1",
+    "sharp": "^0.34.5",
     "starknet": "^6.11.0",
     "ts-node": "^10.8.1",
     "typescript": "^4.7.3"
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -45,7 +45,7 @@ export function sha256(str) {
   return createHash('sha256').update(str).digest('hex');
 }
 
-export async function resize(input, w, h, options?) {
+export async function resize(input, w, h, options?): Promise<Buffer> {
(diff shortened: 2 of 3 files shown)


wa0x6e (approved):
utAck

PR #460, opened 2026-06-10
chore(deps): upgrade @snapshot-labs/snapshot-sentry to v2.1.0

tony8713:
Upgrades `@snapshot-labs/snapshot-sentry` from `^1.5.5` to `^2.1.0` as requested.

## Breaking change
snapshot-sentry v2 bumps `@sentry/node` to v10, which uses **OpenTelemetry-based auto-instrumentation**. Per the upgrade instructions in the snapshot-sentry README:

- `initLogger()` **must run before** any instrumented module (`http`, `express`, `pg`, ...) is imported. The recommended pattern is a dedicated init file loaded first.
- `initLogger()` **no longer accepts** an Express `app` argument (the old `initLogger(app)` signature is gone).
- `fallbackLogger(app)` and `capture()` signatures are unchanged.

## Changes
- Added `src/instrument.ts` that loads `dotenv/config` and calls `initLogger()` at the very top of the process.
- `src/index.ts` now imports `./instrument` as the **very first** import (before `express`), and calls `fallbackLogger(app)` as before. Removed the old `initLogger(app)` call.
- Bumped the dep in `package.json` + `yarn.lock`.

No changes needed to the many `capture(...)` call sites - that API is unchanged.

## Verification
- `yarn lint` ✅
- `yarn typecheck` (tsc --noEmit) ✅
- `yarn build` (tsc) ✅
- Smoke test: `node dist/src/index.js` boots, logs "Sentry is disabled (missing SENTRY_DSN)" before "Listening...", confirming init runs first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@metamask/jazzicon": "^2.0.0",
     "@self.id/core": "^0.3.0",
     "@snapshot-labs/snapshot-metrics": "^1.4.1",
-    "@snapshot-labs/snapshot-sentry": "^1.5.5",
+    "@snapshot-labs/snapshot-sentry": "^2.1.0",
     "@snapshot-labs/snapshot.js": "^0.14.21",
     "@unstoppabledomains/resolution": "^9.2.2",
     "@webinterop/dns-connect": "^0.3.1",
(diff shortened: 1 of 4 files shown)


wa0x6e (approved):
utAck

PR #480, opened 2026-07-05
feat: add support for .gwei domains

bonustrack:
## Summary

Adds a **Gwei Name Service (GNS)** address resolver, enabling `.gwei` name resolution in both directions.

`.gwei` names ([gwei.domains](https://gwei.domains)) are an ownerless, ENS-compatible namespace on Ethereum mainnet ([lucadonnoh/gwei-names](https://github.com/lucadonnoh/gwei-names)). Names are ERC-721 tokens whose id is the ENS namehash of the name — so token ids are computed off-chain and reads are batched through the existing `batchContractCalls` (Multicaller) helper, matching the `spaceId` resolver.

- `resolveNames` (name → address): `resolve(namehash(name))` on the GNS contract.
- `lookupAddresses` (address → name): `reverseResolve(address)`. The contract only returns a name that forward-resolves back to the address, so reverse records are spoofing-safe on-chain — no extra confirmation call needed.
- Registered `gweiResolver` in the resolver list (after `spaceId`; ENS keeps priority when an address has both an ENS and a `.gwei` primary name).

No new runtime dependency — the ~2 contract reads are done natively via the existing ethers/Multicaller stack.

## Test plan

- New integration test `test/integration/addressResolvers/gwei.test.ts` using `donnoh.gwei` ↔ `0xC04689227Fa24785609B1174698DBe481437f1A3` as a bidirectional fixture — 6/6 passing.
- `addressResolvers/index.test.ts` still passes (registration integrates cleanly).
- Verified live against mainnet that off-chain `namehash` matches the contract's on-chain `computeId`, and that forward/reverse round-trip.


src/addressResolvers/gwei.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function normalizeHandles(handles: Handle[]): Handle[] {
+  return handles.filter(h => h.endsWith(TLD));
+}
+
+// Token id is the ENS namehash of the name; '' when it can't be normalized.
+function tokenId(handle: Handle): string {
+  try {
+    return namehash(handle);
```
wa0x6e: Hash raw bytes like the contract's `computeNamehash` (keccak label chain, ASCII-lowercase only, no NFC) instead of ethers `namehash`. Verified to match the deployed contract's `computeId` for every name including both café twins. Needs these imports at the top of the file (remove the `namehash` import):

```ts
import { keccak256 } from '@ethersproject/keccak256';
import { toUtf8Bytes } from '@ethersproject/strings';
import { concat } from '@ethersproject/bytes';
```

```suggestion
    const labels = handle.replace(/\.gwei$/i, '').split('.').concat('gwei');
    let node = `0x${'00'.repeat(32)}`;
    for (let i = labels.length - 1; i >= 0; i--) {
      const label = labels[i].replace(/[A-Z]/g, c => c.toLowerCase());
      node = keccak256(concat([node, keccak256(toUtf8Bytes(label))]));
    }
    return node;
```

wa0x6e (changes requested):
### Blocker: `tokenId()` uses ethers `namehash`, which normalizes Unicode; the GNS contract does not — so lookups can resolve to the wrong owner

`tokenId()` (line 31) computes the token id with ethers' `namehash`, which applies **ENSIP-15 / Unicode NFC normalization** before hashing. The GNS contract does **not** normalize — `_validateAndNormalize` only lowercases ASCII A–Z and stores raw UTF-8 bytes (the source even says *"Unicode normalization should be done off-chain"*). The two hash functions therefore disagree on any name that isn't already normalized.

Concretely, `café.gwei` written composed (`é` = U+00E9) and decomposed (`e` + U+0301) are **two distinct, separately-registerable tokens on-chain**, but ethers `namehash` folds both to the composed token's id. Verified live against mainnet (`0x9D51D507BC7264d4fE8Ad1cf7Fe191933A0a81d6`, verified source on Sourcify):

| | contract tokenId | registerable |
|---|---|---|
| composed `café.gwei` | `0x66df2ee8…db7de9` | `isAvailable = true` |
| decomposed `café.gwei` | `0x16242b1a…f0bde2` | `isAvailable = true` |
| ethers `namehash` (both forms) | `0x66df2ee8…db7de9` (= composed) | — |

So if Alice registers the composed name and a user looks up the decomposed twin (e.g. pasted from an attacker), `namehash` silently rewrites it to the composed id and the resolver returns **Alice's address for a name she never registered**. It never throws, so the `try/catch` in `tokenId()` doesn't catch it — a wrong answer, not an error. Latent today (neither twin is registered yet), but the contract permits the setup.

**Fix: drop `namehash` and hash raw bytes exactly like the contract.** The contract's `computeNamehash` is a plain keccak label chain with ASCII-lowercasing and no NFC — trivial to reproduce, and it's the same shape ethers' own namehash uses minus the normalization step. See the inline suggestion.

Import changes at the top of the file:

```ts
// remove:
import { namehash } from '@ethersproject/hash';
// add:
import { keccak256 } from '@ethersproject/keccak256';
import { toUtf8Bytes } from '@ethersproject/strings';
import { concat } from '@ethersproject/bytes';
```

(The now-stale comment on line 28 — *"Token id is the ENS namehash…"* — should be updated too, and the `try/catch` becomes dead since keccak/utf8 don't throw here; fine to keep or drop.)

Verified the reimplementation against the deployed contract's `computeId` for `donnoh.gwei`, `ABC.gwei`, `sub.name.gwei`, `日本.gwei`, and **both** café twins — all match exactly, so each name resolves to its own true owner and the misdirection is closed at the source.

Why this over an `ens_normalize` filter (the ens.ts / basename.ts pattern): raw hashing serves the **full** GNS namespace instead of dropping non-normalized names, and it stays symmetric with this resolver's reverse path (`lookupAddresses` already returns raw on-chain names via `endsWith` only). A normalize-and-filter approach would make forward refuse names that reverse happily surfaces.

bonustrack:
Good catch, thanks @wa0x6e — fixed in 2589194.

`tokenId()` now reproduces the contract's namehash exactly instead of using ethers' `namehash`: a plain keccak label chain with ASCII-only lowercasing and raw UTF-8 bytes (no Unicode NFC). I verified the reimplementation matches the deployed contract's `computeId` for `donnoh.gwei`, `ABC.gwei`, `sub.name.gwei`, `日本.gwei`, and **both** café twins (composed U+00E9 and decomposed e+U+0301) — the composed/decomposed misdirection is closed, each name resolves to its own owner. The three `@ethersproject/{bytes,keccak256,strings}` subpackages are now declared in `package.json`.

wa0x6e (approved):
tAck

PR #483, opened 2026-07-10
Migrate ENS resolution to the Universal Resolver

tony8713:
## Why

Reverse and forward ENS resolution are spread over three sources today: the ReverseRecords helper contract for the reverse batch, an `ethers` `lookupAddress` pass to patch up the addresses that contract misses, and the ENS subgraph for forward resolution. So a reverse lookup runs two code paths with two failure modes stitched together, and forward resolution leans on a hosted subgraph we do not control.

## What

Both directions now go through the ENS Universal Resolver, via a viem client kept behind one small module in `src/addressResolvers`. viem follows the CCIP-Read `OffchainLookup` flow natively and the Universal Resolver forward-verifies the reverse claim on chain, so a single path covers on-chain and off-chain names alike. The ReverseRecords contract, the patch-up pass and the subgraph query all go away.

Each address or name is settled on its own, so a reverting or expired entry resolves to nothing instead of affecting the rest of the batch. The external API (`lookupAddresses`, `resolveNames`) is unchanged. The avatar resolver reads the `avatar` text record through the same client, keeping the `metadata.ens.domains` fallback. viem stays boxed inside that one module, plain values in and out, and ethers stays everywhere else.

## Tradeoff

The viem client has no multicall batching configured, so a reverse lookup makes one RPC call per address where it used to make a single batched contract call. At the batch size the resolver uses that is a real increase in calls per request, and I think it is worth turning on viem's multicall batching before this merges.

## Notes

The avatar text record change overlaps with snapshot-labs/snapshot.js#1227, which does the same thing upstream. If that lands first, this part can be dropped.

Checked against mainnet with a normal name, a name served by an off-chain resolver, and an address whose name has expired.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "sharp": "^0.34.5",
     "starknet": ">=6.21.0 <6.21.2",
     "ts-node": "^10.8.1",
-    "typescript": "^4.7.3",
+    "typescript": "5.4.5",
+    "viem": "2.37.6",
     "zod": "^3.25.76"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "sharp": "^0.34.5",
     "starknet": "^6.11.0",
     "ts-node": "^10.8.1",
-    "typescript": "^4.7.3",
+    "typescript": "5.4.5",
+    "viem": "2.37.6",
     "zod": "^3.25.76"
   },
   "devDependencies": {
--- a/src/addressResolvers/ens.ts
+++ b/src/addressResolvers/ens.ts
@@ -1,14 +1,11 @@
 import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
-import snapshot from '@snapshot-labs/snapshot.js';
-import { provider as getProvider, isEvmAddress, isSilencedError } from './utils';
-import constants from '../constants.json';
-import { Address, graphQlCall, Handle } from '../utils';
+import { forwardLookup, reverseLookup } from './universalResolver';
+import { isEvmAddress, isSilencedError } from './utils';
+import { Address, Handle } from '../utils';
 
 export const NAME = 'Ens';
-const NETWORK = '1';
-const provider = getProvider(NETWORK);
 
 function normalizeEns(names: Handle[]): Handle[] {
   return names.map(name => {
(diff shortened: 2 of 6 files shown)


tony8713:
Merged master in to clear the conflicts. Both were in `src/addressResolvers/ens.ts`, against #507, which deleted `FetchError` and moved resolver reporting into `index.ts`. A fully failed batch now rethrows the original error instead of the sentinel, and the test asserts that rather than `instanceof FetchError`.

One judgement call worth a look. When part of a batch resolves, the resolver still captures locally. `index.ts` only reports what a resolver *throws*, and a batch that returns never reaches it, so it is report here or not at all. #507 already carved out that same exception for `resolveNames` in this file and for `lookupDomains/shibarium.ts`, so I followed it, but it is two lines to drop if you would rather stay silent on partial failures. Added two cases covering it.

Separately, the Why section is stale and will over-credit the diff: off-chain reverse names were already fixed by #430 (`provider.lookupAddress`) and the reverting batch call by #478, both well before this branch was cut. What is actually left is one Universal Resolver call per address instead of the batch contract plus N fallbacks, with on-chain forward verification, and `resolveNames` off the ENS subgraph.

Last thing, the `avsa.eth` fixture is dead. Its CCIP gateway `gateway.namestone.xyz` answers `503 This service has been suspended by its owner`, so that test fails here and on master identically. I left it alone since it is master's test and master's failure. The Universal Resolver path itself is fine: `offchainexample.eth` and `jesse.base.eth` both resolve through it.

Gate: typecheck, lint and unit (13/13) all clean. Integration is 11 failed / 161 passed on this branch against 11 failed / 158 passed on a clean `master` checkout at 2358802, same machine and env, and the two failure sets are the same 11 tests (the ENS one is this PR's rename of the same avsa.eth case). The +3 is the cases this PR adds. No regressions.

package.json
```diff
@@ -48,7 +48,8 @@
     "sharp": "^0.34.5",
     "starknet": ">=6.21.0 <6.21.2",
     "ts-node": "^10.8.1",
-    "typescript": "^4.7.3",
+    "typescript": "5.4.5",
+    "viem": "2.37.6",
```
wa0x6e, line 52: This pin installs a second viem copy: `@snapshot-labs/snapshot.js` already pulls in `viem@^2.55.11`, and `2.37.6` does not satisfy that range, so yarn keeps both (two `viem@` entries in yarn.lock). Suggest `"viem": "^2.55.11"` so both dedupe to one copy — also picks up the newer Universal Resolver/CCIP implementation the sdk uses.

tony8713:
Closing this. The correctness problems it was written against were already fixed on master before it was opened: #430 resolves names set via off-chain resolvers through CCIP-Read, and #478 stops a reverting entry taking down the rest of the batch. What is left is the "one path instead of two" cleanup, which is a refactor rather than a fix.

Two of the three pieces also have nowhere to go now. snapshot.js has since taken over the avatar text record, but it exposes no reverse lookup and no forward `addr` resolution, so `lookupAddresses` and `resolveNames` would still be moving onto a viem client stamp owns rather than onto a shared one.

Worth picking up again if the single-path cleanup is still wanted and the fan-out cost is answered first, since neither client batches and `lookupAddresses` takes up to 50 addresses per request.

wa0x6e:
Reopened as a series with the batching answered: #558 (reverse), #559 (forward), #560 (avatar).

PR #491, opened 2026-08-02
fix(starknet): resolve names on-chain instead of via api.starknet.id

### LINKED ISSUE #503: Garbage `.stark` handles resolve to a real user's address
`a!b.stark` resolves to the address that owns `ab.stark`. Any character outside `a-z0-9-` is silently dropped before the lookup, so a handle nobody has registered comes back carrying a real person's address.

```
resolveNames(['a!b.stark', 'ab.stark'])
{
  'a!b.stark': '0x00480b575faf9d674f26ddcb15c47dfa10086a2378e8194e836af28116b09060',
  'ab.stark':  '0x00480b575faf9d674f26ddcb15c47dfa10086a2378e8194e836af28116b09060'
}

resolveNames(['!!!.stark'])
{ '!!!.stark': '0x061b6c0a78f9edf13cea17b50719f3344533fadd470b8cb29c2b4318014f52d3' }
// lookupAddresses() on that address says fricoben.stark
```

StarknetID's encoder works off the alphabet `abcdefghijklmnopqrstuvwxyz0123456789-` plus a two-character "big" alphabet. A character in neither hits the `indexOf === -1` branch and is skipped without advancing the multiplier, so it just disappears:

```js
const { starknetId } = require('starknet');
starknetId.useEncoded('a!b'); // 38n — same as 'ab'
starknetId.useEncoded('!!!'); // 0n  — same as '', and felt 0 maps to fricoben's address
```

`normalizeHandles` (`src/addressResolvers/starknet.ts:45`) only checks `h.endsWith('.stark')`, and the shared gate in `src/addressResolvers/utils.ts:46` only rejects whitespace, so this reaches the public `resolve_names`.

### Pre-existing or introduced by #491?

Both, in different places.

`src/resolvers/starknet.ts:11` has the same `.endsWith('.stark')` check and has always resolved on-chain, so avatars collide on master today — it returns byte-identical images for `ab.stark`, `a!b.stark` and `a!!b.stark`.

The address resolver on master goes through api.starknet.id, whose `/domain_to_addr` matches the domain string exactly in Mongo and 400s on a miss, so no encoding and no collision there. #491 moves that resolver on-chain, which is what brings the bug into `resolve_names`. Caveat: api.starknet.id is still refusing connections (the outage #491 fixes), so that last part is read off the deployed API source rather than tested against the live service.

### Fix

Tighten `normalizeHandles` to StarknetID's own `isStarkDomain` character class, which keeps subdomains like `notion.eth.stark`:

```ts
return handles.filter(h => /^[a-z0-9-]+(\.[a-z0-9-]+)*\.stark$/.test(h));
```

`src/resolvers/starknet.ts` wants the same. No other resolver has this: ens goes through ens-normalize, lens and unstoppable already have strict regexes, and spaceId/gwei/shibarium namehash or DNS-resolve the handle so garbage stays distinct.

tony8713:
Fixes #503.

Depends on #504 — #491 drops its own range check in favour of the shared one #504 adds to `normalizeAddresses`, and the batch-guard test here stays red until #504 lands.
Depends on release of snapshot-labs/snapshot.js#1225 — merging #1225 is not enough; the `@snapshot-labs/snapshot.js` floor in `package.json` has to be raised to a published release that carries `baseFetch`.

### What was broken

The Starknet address resolver read `https://api.starknet.id`, and that host stopped accepting connections. `.stark` names quietly stopped resolving in production as a result: `lookup_addresses` on a named address returned the Snapshot profile-name fallback instead of the handle.

Nothing reported it. `apiCall` fanned out with `Promise.allSettled` and read a value only from fulfilled responses, so a total outage produced `{}` — indistinguishable from "no such name". Nothing ever threw, so the `capture()` in `index.ts` never fired and no Sentry event was raised. A red `Test` workflow is what surfaced it.

This is not a stale fixture like #482. The name the test asserts is still registered and still maps to the address it expects; the API was down for every name, so no fixture swap could have fixed it.

### The fix

Resolve on-chain instead, reading the StarknetID naming contract through the brovider we control. Not a new dependency or a new pattern — the avatar resolver in `src/resolvers/starknet.ts` already reads that contract through the same `snapshot.utils.getProvider` — and it drops a third-party HTTP single point of failure that [StarknetID's own docs](https://docs.starknet.id/devs/starknetid-api) advise against depending on.

Both directions go through one `snapshot.utils.multicall`, so a batch is a single round trip rather than a call per name. The `@snapshot-labs/snapshot.js` bump is what makes that possible, and it is load-bearing rather than incidental: the reverse lookup returns a `Span::<felt252>`, which the older `parseStarknetResult` has no case for — it hands back the span's length felt and the resolver returns `{}` for every address, which is the same silent failure this PR exists to remove. If the bump is ever lost to a dedup or a downgrade, the bug comes back.

A miss is a value now rather than a throw — an empty span one way, the zero address the other — so no error-message matching is left in the resolver. Anything that does throw out of the multicall is a real RPC or network fault, and it propagates to `index.ts` rather than being flattened into an empty result. An outage of this kind is loud now instead of silent.

### Handle validation (#503)

Resolving on-chain also brings StarknetID's encoder into `resolveNames`, and that encoder *skips* characters outside its alphabet rather than rejecting them, without advancing the multiplier. So `a!b` encodes identically to `ab`, and `!!!` encodes to felt 0, which maps to a real account. Checking the `.stark` suffix is not a sufficient gate: a garbage handle comes back carrying somebody else's address. The avatar resolver had the same gap, where `ab.stark`, `a!b.stark` and `a!!b.stark` all resolved to a byte-identical image.

`isStarkDomain` now gates on the encoder's own alphabet, subdomains included, and both call sites share the one helper so they cannot drift. Parts of that gate look wrong at a glance:

- The class is starknet.js's `basicAlphabet` **plus** its `bigAlphabet` (`这来`), not just `[a-z0-9-]`. Those glyphs are safe to admit: unlike a skipped character, the encoder consumes them and advances the multiplier, so they round-trip and cannot collide with a shorter handle. They are also in live use, so rejecting them would have dropped an already-registered name to the blockie fallback.
- The felt bound is guarded directly rather than through upstream's character cap. A label whose encoded felt reaches the field prime is rejected by the node, and because resolution is one atomic multicall, that failure takes down every name batched alongside it — as an error `isSilencedError` does not silence. Character length is the wrong proxy for that bound; the felt is the actual invariant.

### DIFF
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -1,15 +1,12 @@
+import { validateAndParseAddress } from 'starknet';
 import { provider as getProvider, isStarknetAddress, withoutEmptyValues } from './utils';
 import { Address, Handle } from '../utils';
 
 export const NAME = 'Starknet';
 const NETWORK = '0x534e5f4d41494e';
 const NOT_FOUND_ERROR = 'Starkname not found';
-const EMPTY_STARKNET_ADDRESS = '0x0';
+const EMPTY_STARKNET_ADDRESS = `0x${'0'.repeat(64)}`;
 const provider = getProvider(NETWORK);
-
-function padAddress(address: Address): Address {
-  return `0x${address.replace(/^0x/, '').padStart(64, '0')}`;
-}
 
 function normalizeAddresses(addresses: Address[]): Address[] {
   return addresses.filter(isStarknetAddress);
@@ -53,6 +50,10 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
   return await resolveEach(normalizedHandles, async handle => {
     const address = await provider.getAddressFromStarkName(handle);
 
-    return !address || address === EMPTY_STARKNET_ADDRESS ? undefined : padAddress(address);
+    if (!address) return undefined;
+
+    const parsedAddress = validateAndParseAddress(address);
+
+    return parsedAddress === EMPTY_STARKNET_ADDRESS ? undefined : parsedAddress;
   });
 }
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -1,41 +1,14 @@
-import axios from 'axios';
-import { isStarknetAddress, withoutEmptyValues } from './utils';
+import { provider as getProvider, isStarknetAddress, withoutEmptyValues } from './utils';
 import { Address, Handle } from '../utils';
 
 export const NAME = 'Starknet';
-const BASE_URL = 'https://api.starknet.id';
+const NETWORK = '0x534e5f4d41494e';
+const NOT_FOUND_ERROR = 'Starkname not found';
+const EMPTY_STARKNET_ADDRESS = '0x0';
+const provider = getProvider(NETWORK);
 
-type RESOLVE_TYPE = 'addr_to_domain' | 'domain_to_addr';
-
-function buildApiUrl(resolve_type: RESOLVE_TYPE, needle: string): string {
-  return `${BASE_URL}/${resolve_type}?${
-    resolve_type === 'addr_to_domain' ? 'addr' : 'domain'
-  }=${needle}`;
-}
-
-async function apiCall(
-  resolve_type: RESOLVE_TYPE,
-  needles: string[]
-): Promise<Record<string, string>> {
-  const requests = needles.map(needle =>
-    axios.get(buildApiUrl(resolve_type, needle), { timeout: 5e3 })
-  );
-  const responses = await Promise.allSettled(requests);
-
-  return withoutEmptyValues(
-    Object.fromEntries(
-      needles.map((needle, i) => {
-        const response = responses[i];
-        let value: string | undefined;
-
-        if (response.status === 'fulfilled') {
-          value = response.value.data[resolve_type === 'addr_to_domain' ? 'domain' : 'addr'];
-        }
-
-        return [needle, value];
-      })
-    )
-  );
+function padAddress(address: Address): Address {
+  return `0x${address.replace(/^0x/, '').padStart(64, '0')}`;
 }
 
 function normalizeAddresses(addresses: Address[]): Address[] {
@@ -46,12 +19,30 @@ function normalizeHandles(handles: Handle[]): Handle[] {
   return handles.filter(h => h.endsWith('.stark'));
 }
 
+async function resolveEach(
+  needles: string[],
+  lookup: (needle: string) => Promise<string | undefined>
+): Promise<Record<string, string>> {
+  const values = await Promise.all(
+    needles.map(async needle => {
+      try {
+        return await lookup(needle);
+      } catch (err: any) {
+        if (err?.message?.includes(NOT_FOUND_ERROR)) return undefined;
+        throw err;
+      }
+    })
+  );
+
+  return withoutEmptyValues(Object.fromEntries(needles.map((needle, i) => [needle, values[i]])));
+}
+
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
   const normalizedAddresses = normalizeAddresses(addresses);
 
   if (normalizedAddresses.length === 0) return {};
 
-  return await apiCall('addr_to_domain', normalizedAddresses);
+  return await resolveEach(normalizedAddresses, address => provider.getStarkName(address));
 }
 
 export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Address>> {
@@ -59,5 +50,9 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
 
   if (normalizedHandles.length === 0) return {};
 
-  return await apiCall('domain_to_addr', normalizedHandles);
+  return await resolveEach(normalizedHandles, async handle => {
+    const address = await provider.getAddressFromStarkName(handle);
+
+    return !address || address === EMPTY_STARKNET_ADDRESS ? undefined : padAddress(address);
+  });
 }
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -8,8 +8,21 @@ const NOT_FOUND_ERROR = 'Starkname not found';
 const EMPTY_STARKNET_ADDRESS = `0x${'0'.repeat(64)}`;
 const provider = getProvider(NETWORK);
 
+// A 64 hex-digit string is not necessarily a valid Starknet address: anything above
+// the felt address bound makes the StarknetID call fail with an opaque
+// 'Could not get stark name', which addressResolvers/index would then report as an
+// outage. Drop those here, so that error only ever means a real failure.
+function isResolvableAddress(address: Address): boolean {
+  try {
+    validateAndParseAddress(address);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
 function normalizeAddresses(addresses: Address[]): Address[] {
-  return addresses.filter(isStarknetAddress);
+  return addresses.filter(a => isStarknetAddress(a) && isResolvableAddress(a));
 }
 
 function normalizeHandles(handles: Handle[]): Handle[] {
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -1,5 +1,5 @@
 import snapshot from '@snapshot-labs/snapshot.js';
-import { constants, starknetId, validateAndParseAddress } from 'starknet';
+import { CallData, constants, starknetId, validateAndParseAddress } from 'starknet';
 import {
   provider as getProvider,
   hasStarknetAddressShape,
@@ -56,21 +56,17 @@ function normalizeHandles(handles: Handle[]): Handle[] {
   return handles.filter(isStarkDomain);
 }
 
-// `CallData.compile({ address, hint: [] })`, byte for byte what starknet.js sends: the
-// address, then an empty span, which serialises to a bare length of 0.
 function addressToDomainCalldata(address: Address): string[] {
-  return [address, '0'];
+  return CallData.compile({ address, hint: [] });
 }
 
-// `CallData.compile({ domain: encodedDomain, hint: [] })`: the label count, one felt per
-// label leaf first, then the empty hint span.
 function domainToAddressCalldata(handle: Handle): string[] {
-  const labels = handle
+  const domain = handle
     .replace(/\.stark$/, '')
     .split('.')
-    .map(label => starknetId.useEncoded(label).toString(10));
+    .map(label => starknetId.useEncoded(label));
 
-  return [labels.length.toString(), ...labels, '0'];
+  return CallData.compile({ domain, hint: [] });
 }
 
 // An address with no domain answers with an empty span rather than an error, so a miss
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -2,8 +2,8 @@ import snapshot from '@snapshot-labs/snapshot.js';
 import { CallData, constants, starknetId, validateAndParseAddress } from 'starknet';
 import {
   provider as getProvider,
-  hasStarknetAddressShape,
   isStarkDomain,
+  isStarknetAddress,
   withoutEmptyValues
 } from './utils';
 import { Address, Handle } from '../utils';
@@ -37,12 +37,13 @@ const NAMING_ABI = [
   }
 ];
 
-// Shape only. The felt bound is guarded upstream, in the shared `normalizeAddresses` that
-// addressResolvers/index runs before any resolver sees the input; what is left here is the
-// part that guard does not do, rejecting EVM addresses, which are valid felts themselves.
-// The composition is pinned by test/unit/addressResolvers/starknet-batch-guard.test.ts.
+// Shape and felt range both. `addressResolvers/index` already applies the same guard upstream,
+// but resolution is one atomic multicall, so a single out-of-range 64 hex-digit value -- a
+// transaction hash, a proposal id -- would take down every name batched alongside it with a
+// `-32602` that `isSilencedError` does not silence. Cheap enough to not depend on the caller.
+// The upstream composition stays pinned by test/unit/addressResolvers/starknet-batch-guard.test.ts.
 function normalizeAddresses(addresses: Address[]): Address[] {
-  return addresses.filter(hasStarknetAddressShape);
+  return addresses.filter(isStarknetAddress);
 }
 
 function normalizeHandles(handles: Handle[]): Handle[] {
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -4,6 +4,7 @@ import {
   provider as getProvider,
   isStarkDomain,
   isStarknetAddress,
+  starkDomainLabels,
   withoutEmptyValues
 } from './utils';
 import { Address, Handle } from '../utils';
@@ -55,10 +56,7 @@ function addressToDomainCalldata(address: Address): string[] {
 }
 
 function domainToAddressCalldata(handle: Handle): string[] {
-  const domain = handle
-    .replace(/\.stark$/, '')
-    .split('.')
-    .map(label => starknetId.useEncoded(label));
+  const domain = starkDomainLabels(handle).map(label => starknetId.useEncoded(label));
 
   return CallData.compile({ domain, hint: [] });
 }
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -2,8 +2,8 @@ import snapshot from '@snapshot-labs/snapshot.js';
 import { constants, starknetId, validateAndParseAddress } from 'starknet';
 import {
   provider as getProvider,
+  hasStarknetAddressShape,
   isStarkDomain,
-  isStarknetAddress,
   withoutEmptyValues
 } from './utils';
 import { Address, Handle } from '../utils';
@@ -41,21 +41,15 @@ const NAMING_ABI = [
   }
 ];
 
-// A 64 hex-digit string is not necessarily a valid Starknet address: anything above
-// the felt address bound makes the StarknetID call fail with an opaque
-// 'Could not get stark name', which addressResolvers/index would then report as an
-// outage. Drop those here, so that error only ever means a real failure.
-function isResolvableAddress(address: Address): boolean {
-  try {
-    validateAndParseAddress(address);
-    return true;
-  } catch {
-    return false;
-  }
-}
-
+// Both halves are load-bearing. The shape check is what rejects EVM addresses, which
+// `snapshot.utils.isStarknetAddress` reads as perfectly valid felts. And a 64 hex-digit
+// string is in turn not necessarily a valid address: anything above the felt address
+// bound is rejected by the node, and since the lookup goes through a multicall, one such
+// value fails the *whole* batch with '-32602 ... maximum field value was exceeded', which
+// `isSilencedError` does not silence, so addressResolvers/index would report an outage
+// for every address sent alongside it. Drop them both here.
 function normalizeAddresses(addresses: Address[]): Address[] {
-  return addresses.filter(a => isStarknetAddress(a) && isResolvableAddress(a));
+  return addresses.filter(a => hasStarknetAddressShape(a) && snapshot.utils.isStarknetAddress(a));
 }
 
 function normalizeHandles(handles: Handle[]): Handle[] {
--- a/src/addressResolvers/starknet.ts
+++ b/src/addressResolvers/starknet.ts
@@ -1,4 +1,5 @@
-import { validateAndParseAddress } from 'starknet';
+import snapshot from '@snapshot-labs/snapshot.js';
+import { constants, starknetId, validateAndParseAddress } from 'starknet';
 import {
   provider as getProvider,
   isStarkDomain,
@@ -8,10 +9,37 @@ import {
 import { Address, Handle } from '../utils';
 
 export const NAME = 'Starknet';
-const NETWORK = '0x534e5f4d41494e';
-const NOT_FOUND_ERROR = 'Starkname not found';
+const NETWORK = constants.StarknetChainId.SN_MAIN;
 const EMPTY_STARKNET_ADDRESS = `0x${'0'.repeat(64)}`;
+const NAMING_CONTRACT = starknetId.getStarknetIdContract(NETWORK);
 const provider = getProvider(NETWORK);
+
+// The two entrypoints starknet.js calls one address at a time in `getStarkName` and
+// `getAddressFromStarkName`. Declared here so `snapshot.utils.multicall` can parse the
+// retdata: it keys off `name`, and needs `outputs` to know a domain comes back as a span
+// of felts while an address is a single one.
+const NAMING_ABI = [
+  {
+    name: 'address_to_domain',
+    type: 'function',
+    inputs: [
+      { name: 'address', type: 'core::starknet::contract_address::ContractAddress' },
+      { name: 'hint', type: 'core::array::Span::<core::felt252>' }
+    ],
+    outputs: [{ type: 'core::array::Span::<core::felt252>' }],
+    state_mutability: 'view'
+  },
+  {
+    name: 'domain_to_address',
+    type: 'function',
+    inputs: [
+      { name: 'domain', type: 'core::array::Span::<core::felt252>' },
+      { name: 'hint', type: 'core::array::Span::<core::felt252>' }
+    ],
+    outputs: [{ type: 'core::starknet::contract_address::ContractAddress' }],
+    state_mutability: 'view'
+  }
+];
 
 // A 64 hex-digit string is not necessarily a valid Starknet address: anything above
 // the felt address bound makes the StarknetID call fail with an opaque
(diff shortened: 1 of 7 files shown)


tony8713:
Good call, done in 9349326: `padAddress` is gone in favour of `validateAndParseAddress` from starknet.js, which pads to 64 hex digits and additionally lowercases and rejects anything that isn't a valid contract address (the helper silently returned the zero address for `''` and passed `0xzz` straight through). `num.toHex` / `encode.addHexPrefix` / `encode.padLeft` don't work here since they don't pad to 64, and `num.toHex64` only exists from starknet v7 (we're on 6.11.0). Since the value is now normalized, the empty-address check moved after the parse so a padded `0x000...0` is dropped too; added test/unit/addressResolvers/starknet.test.ts covering it.

tony8713:
Bumped to `@snapshot-labs/snapshot.js@0.15.1` and moved both directions onto `snapshot.utils.multicall`, so a batch of N names is one RPC round trip instead of N.

On 0.15.1 specifically, since it went out nine minutes after #1219 merged: it was cut from `a62c5d8`, which has the #1219 squash `53a41ba` as an ancestor, so it carries the felt252 padding fix rather than the revision that regressed low bytes. Worth checking on the published tarball rather than the repo, because the branch shas are not ancestors of master after the squash. Driven through the published parser, `0x55534409436f696e` decodes to `"USD\tCoin"` and `0xcafebabe` still comes back as a raw felt.

Same answers as the per-address version on every name I ran against mainnet, subdomains included: `checkpoint.stark`, `fricoben.stark`, `abwab.stark`, `th0rgal.stark`, `fricoben.braavos.stark`, `test.braavos.stark`, `alpha.braavos.stark`.

A miss is a value now, an empty span or the zero address, so the `Starkname not found` message match is gone. Anything that throws out of `multicall` is still an RPC fault and still reaches `capture()`.

The unit suite moves off the `getStarkName` mocks onto a `callContract` stub returning real `aggregate` retdata. Every case it pinned before is still pinned, plus the span decode, a truncated response, and one asserting a whole batch costs a single call.

`yarn typecheck` and `yarn lint` clean, all three starknet suites green. The `ens` CCIP-Read and `zapper` failures are unchanged from before this push and unrelated.

src/addressResolvers/utils.ts
```diff
@@ -12,6 +12,15 @@ export function isStarknetAddress(address: Address): boolean {
...
 }
 
+// StarknetID's encoder skips any character outside this class instead of rejecting it,
+// so `a!b.stark` encodes exactly like `ab.stark` and resolves to that owner, and
+// `!!!.stark` encodes to felt 0, which maps to a real account. Checking the suffix is
+// therefore not enough: the whole handle has to match the character class StarknetID
+// itself accepts (`isStarkDomain` in lfglabs-dev/starknetid.js), subdomains included.
+export function isStarkDomain(domain: Handle): boolean {
+  return /^[a-z0-9-]+(\.[a-z0-9-]+)*\.stark$/.test(domain);
+}
```
wa0x6e, line 40: **`isStarkDomain` needs a felt-bound guard on each label** — the character class alone still admits labels whose encoded felt exceeds the field prime, and the RPC then rejects the entire multicall.

Verified live against `rpc.snapshot.org/sn` with the exact calldata this code generates:

- `'a'.repeat(49) + '.stark'` → `-32602: Invalid params: {"reason":"More than 64 digits found: 65"}`
- `'a'.repeat(48) + '.stark'` (fits 64 hex digits, but `useEncoded` ≥ PRIME) → `-32602: Invalid params: {"reason":"The maximum field value was exceeded."}`

Both surface as a bare `LibraryError` matching nothing in `isSilencedError`, so unauthenticated `resolve_names` input raises a Sentry event — and because resolution is one atomic `aggregate` call, every valid name batched alongside it is lost too (reproduced with `['checkpoint.stark', 'a'.repeat(49)+'.stark']`: the whole call throws and `checkpoint.stark`'s result vanishes).

The PR body's rationale for skipping the 48-char cap was sound under the old per-name, error-swallowing path — but this PR's design is loud errors + atomic batching, and under that design the oversized label is exactly the input that pages like an outage. The right guard is the felt bound (the actual invariant), not a character count: `'a'.repeat(47)+'b'` is 48 chars and valid, `'a'.repeat(48)` is 48 chars and over-prime.

```suggestion
export function isStarkDomain(domain: Handle): boolean {
  if (!/^[a-z0-9-]+(\.[a-z0-9-]+)*\.stark$/.test(domain)) return false;

  return domain
    .replace(/\.stark$/, '')
    .split('.')
    .every(label => starknetId.useEncoded(label) < constants.PRIME);
}
```

Requires adding to this file's imports (outside this suggestion's range):

```ts
import { constants, starknetId } from 'starknet';
```

(`starknet` is already a direct dependency.) Checked against the installed package: drops `'a'.repeat(48)` and `'a'.repeat(49)`, keeps `checkpoint.stark`, `'a'.repeat(47)+'b'`, `fricoben.braavos.stark`. Worth adding boundary regression tests for 47/48/49-char labels and a mixed batch asserting the bad name is excluded while the valid one still resolves.

tony8713:
@wa0x6e Done in `2b96aab`: the shape check is now `hasStarknetAddressShape`, and the local `isResolvableAddress` is gone in favour of `snapshot.utils.isStarknetAddress`.

The filter still runs both, since the SDK function is a felt check and answers `true` for an EVM address.

src/addressResolvers/starknet.ts
```diff
@@ -1,63 +1,134 @@
...
+  withoutEmptyValues
+} from './utils';
 import { Address, Handle } from '../utils';
 
 export const NAME = 'Starknet';
-const BASE_URL = 'https://api.starknet.id';
+const NETWORK = constants.StarknetChainId.SN_MAIN;
+const EMPTY_STARKNET_ADDRESS = `0x${'0'.repeat(64)}`;
+const NAMING_CONTRACT = starknetId.getStarknetIdContract(NETWORK);
+const provider = getProvider(NETWORK);
```
wa0x6e, line 16: **The 5s timeout is silently lost on this path** — `provider()` passes `timeout: 5e3` like every other resolver, but snapshot.js's `getStarknetProvider` forwards only `nodeUrl`:

```js
// @snapshot-labs/snapshot.js 0.15.1, dist/snapshot.cjs.js:2268-2279
function getEvmProvider(networkId, options) {
    return new providers.StaticJsonRpcProvider({
        url: `${options.broviderUrl}/${networkId}`,
        timeout: options.timeout,   // <- forwarded
        allowGzip: true
    }, Number(networkId));
}
function getStarknetProvider(networkKey, options) {
    return new starknet$1.RpcProvider({
        nodeUrl: `${options.broviderUrl}/${networkKey}`  // <- timeout dropped
    });
}
```

starknet.js v6's `RpcChannel` has no timeout field and its `fetch()` carries no `AbortSignal`, so nothing downstream bounds the call either. The old axios path had a real socket-aborting 5s timeout. Net effect: a brovider `/sn` that accepts the connection but never responds (hung reverse proxy — the *other* outage mode from the connection-refused one this PR fixes) hangs `callNamingContract` until undici's ~300s ceiling, and since `index.ts` awaits all resolvers under one `Promise.all`, it stalls the entire `lookup_addresses`/`resolve_names` request — non-Starknet inputs included. Every other resolver keeps its 5s bound (EVM providers get it forwarded; lens/snapshot use axios with `timeout: 5e3`); this is the only path where the option is a silent no-op.

Not fixable properly in this repo: the right fix is upstream in `@snapshot-labs/snapshot.js` — `getStarknetProvider` should honor `options.timeout` (starknet.js's `RpcProvider` exposes no timeout option, so upstream needs to wrap the transport with an `AbortSignal`, or move to a starknet.js version that exposes a fetch/timeout hook). Flagging here so it's tracked; an interim app-level `Promise.race` around the multicall is possible but the upstream fix restores parity with `getEvmProvider` for every consumer.

wa0x6e:
**Minor: the "Fix" section of the PR body describes an implementation this PR no longer ships.**

The body's "Fix" section (and its error-handling/padding sub-bullets) documents the per-name `provider.getStarkName` / `getAddressFromStarkName` design from `b4b6d8c`, including a `Starkname not found` string-match catch. `1894f3a` replaced that with a batched `snapshot.utils.multicall` call the same day — a miss is now decoded from a value (empty span / zero address), not caught from a thrown error, and neither `getStarkName` nor `getAddressFromStarkName` is called anywhere in the shipped file anymore (only referenced in a comment as prior art).

More importantly, the body never mentions the `@snapshot-labs/snapshot.js` `0.14.26` → `0.15.1` bump (`ae00f1d`), and that bump is load-bearing for the new design: diffing `src/multicall/starknet.ts` between the two npm tarballs, 0.14.26's `parseStarknetResult` has no case for `Span::<felt252>` — the type `address_to_domain`'s output uses. On 0.14.26 it falls through to the default branch and returns the span's length felt instead of the decoded span, which makes `decodeDomain()` fail its `Array.isArray` guard and makes `lookupAddresses` silently return `{}` for every address — i.e. the exact silent-outage failure mode this PR exists to fix, reintroduced by the batching refactor if the version bump is ever lost (e.g. a future dependency dedup/downgrade). The new unit suite does catch this today via a mocked `callContract` that exercises the real `parseStarknetResult`, but that safety net isn't documented either.

Since this is purely a documentation gap (the code, tests, and CI are all correct as shipped), I'd suggest editing the PR body's "Fix" section to describe the multicall/batch design and call out why the snapshot.js bump is required, so a future reader investigating a `.stark` incident — or reviewing a dependency bump that touches `snapshot.js` — has the real picture instead of the superseded one. Draft replacement text available on request.

src/addressResolvers/starknet.ts
```diff
@@ -1,63 +1,134 @@
...
+  return [labels.length.toString(), ...labels, '0'];
 }
 
-function normalizeAddresses(addresses: Address[]): Address[] {
-  return addresses.filter(isStarknetAddress);
+// An address with no domain answers with an empty span rather than an error, so a miss
+// is a value here instead of a throw. Anything that does throw out of `multicall` is an
+// RPC or network fault and is left to propagate to `index.ts`, which captures it.
+function decodeDomain(span: unknown): Handle | undefined {
+  if (!Array.isArray(span) || span.length === 0) return undefined;
```
wa0x6e: Nit: these two builders hand-expand exactly what their comments say `CallData.compile` produces — and the library is already imported in this file. The commit message notes the expansion but not a reason for it. Risk of keeping the copy: if `hint` ever becomes non-empty (offchain-resolver domains use hints) or a starknet.js bump changes span framing, these arrays silently drift from what the library sends — and the unit test can't catch it, since it recomputes labels with the same `replace/split/useEncoded` chain (it asserts the encoding against a copy of itself).

```suggestion
function addressToDomainCalldata(address: Address): string[] {
  return CallData.compile({ address, hint: [] });
}

function domainToAddressCalldata(handle: Handle): string[] {
  const domain = handle
    .replace(/\.stark$/, '')
    .split('.')
    .map(label => starknetId.useEncoded(label));

  return CallData.compile({ domain, hint: [] });
}
```

Needs `CallData` added to the existing `starknet` import. Output is byte-identical for the current shapes (verified via `transaction.fromCallsToExecuteCalldata` against the hand-rolled arrays), so no behavior change today — this only removes the drift window.

src/addressResolvers/utils.ts
```diff
@@ -8,10 +8,22 @@ export function isEvmAddress(address: Address): boolean {
...
+// for an EVM address. This one is what tells the two address formats apart.
+export function hasStarknetAddressShape(address: Address): boolean {
   return /^0x[a-fA-F0-9]{64}$/.test(address);
 }
 
+// StarknetID's encoder skips any character outside this class instead of rejecting it,
+// so `a!b.stark` encodes exactly like `ab.stark` and resolves to that owner, and
+// `!!!.stark` encodes to felt 0, which maps to a real account. Checking the suffix is
+// therefore not enough: the whole handle has to match the character class StarknetID
+// itself accepts (`isStarkDomain` in lfglabs-dev/starknetid.js), subdomains included.
```
wa0x6e: **One more gap in the character class: StarknetID's two-glyph "bigAlphabet" (`这`, `来`).** This one is a live avatar regression, verified on mainnet: `来baba这.stark` is registered (`getStarkName('0x180f9e6c...86ae7')` returns it), resolved to a real identicon through the old `endsWith('.stark')` path, and now fails `isStarkDomain` → falls into `normalizeAddress` → throws → silent blockie fallback.

Admitting exactly these two glyphs is safe with respect to the collision bug this PR closes: the encoder *consumes* bigAlphabet characters (multiplier advances — `useEncoded('来baba这')` = `11885385095`, `useDecoded` round-trips it exactly), unlike `!` which is skipped. The encoder-skip collision only applies to characters outside *both* alphabets, which stay rejected. Upstream's `isStarkDomain` shares this gap, so the PR body's "matching upstream" claim is accurate — but upstream's validator gates registration UI, while this one gates resolution of already-registered names.

**This suggestion supersedes my earlier one on this function** (felt-bound guard) — it's both changes merged, apply this block only:

```suggestion
export function isStarkDomain(domain: Handle): boolean {
  if (!/^[a-z0-9-这来]+(\.[a-z0-9-这来]+)*\.stark$/.test(domain)) return false;

  return domain
    .replace(/\.stark$/, '')
    .split('.')
    .every(label => starknetId.useEncoded(label) < constants.PRIME);
}
```

(Still needs `import { constants, starknetId } from 'starknet';` per the earlier comment.) Suggested regression tests: `来baba这.stark` reaches the contract call; `a!b.stark`, underscores, whitespace and non-bigAlphabet Unicode remain rejected; 47/48/49-char labels for the felt bound.

src/addressResolvers/utils.ts
tony8713: Applied, in the merged form from your later comment — the felt bound is in `isStarkDomain` now.

Reproduced your result against `rpc.snapshot.org/sn`: `checkpoint.stark` resolves on its own, and batched with `'a'.repeat(48) + '.stark'` the whole call throws and its result is gone. `isSilencedError` returns `false` for both the 48- and 49-character errors, so it would have paged. Regression tests cover 47/48/49 and the mixed batch.

src/addressResolvers/starknet.ts
tony8713: Confirmed, verbatim — `getStarknetProvider` forwards only `nodeUrl` where `getEvmProvider` forwards `options.timeout`, in the installed 0.15.1 dist at the lines you quoted.

I've bounded it here rather than only tracking it, though: master's axios path had a real 5s timeout on this exact call, so shipping without one is a strict availability regression, and unlike a throw a hang isn't contained by the per-resolver `try/catch` — it stalls the whole `Promise.all`. It's `917baa3`, kept as its own commit, raising `ETIMEDOUT` so `isSilencedError` treats it like the other resolvers' timeouts. Happy to drop that commit if you'd rather carry the gap until the upstream fix lands; leaving this open for your call.

Left the avatar path in `resolvers/starknet.ts` alone — same gap, but a hang there only costs that one request.

src/addressResolvers/starknet.ts
tony8713: Done in `0326332`.

One correction: the two forms aren't byte-identical for `address_to_domain` — the hand-rolled version kept the padded hex string where `CallData.compile` emits the decimal felt (`"5"` vs `"0x00…05"`). Your conclusion still holds, since `multicall` serialises through `fromCallsToExecuteCalldata` and the resulting wire payload is identical; it's only the intermediate array that differs. Also took the second half of your point and moved the assertions onto pinned felts, so they no longer recompute the expected labels with the same chain the resolver uses.

src/addressResolvers/utils.ts
tony8713: Applied as suggested in `12dca32`.

Confirmed `来baba这.stark` against mainnet — registered, resolves to `0x180f9e6c6cc47e2db5d888df7fb9f904fd59cb1a559824ad4f6e597c7c86ae7`, and it was reaching the contract before this PR through `endsWith('.stark')`, so this was a live regression rather than a hypothetical. Also checked the round trip holds across the extended class (every single and pair, plus 40k random labels) so the two glyphs can't reintroduce the collision. Tests cover it alongside 47/48/49 and a mixed batch.

tony8713:
Body updated — you were right that it documented a design this PR stopped shipping.

The "Fix" section now describes the batched `multicall` design rather than the per-name `getStarkName` one, and drops the `Starkname not found` string-match (a miss is a value now, so there is no message match anywhere in the resolver). Added a paragraph on why the `0.15.1` bump is load-bearing: I confirmed 0.14.26's `parseStarknetResult` has no `Span::<felt252>` case and returns the span's length felt, so `lookupAddresses` would go silently empty exactly as you described, and noted that the mocked-`callContract` suite is what catches it.

"Handle validation" is rewritten for the gate as it now stands — the `bigAlphabet` glyphs and the felt bound. The 48-character cap is still not matched, but the reasoning is now that the felt bound is guarded directly instead, rather than that no guard is needed.

tony8713:
Rebased onto master. The only conflict was `package.json`/`yarn.lock`: master bumped `@snapshot-labs/snapshot.js` to `^0.14.27` while this branch had it at `^0.15.1`. I kept `^0.15.1` — it is the newer release and this branch needs the Starknet multicall span parsing from it — and took master's `@snapshot-labs/snapshot-metrics@^2.0.0` alongside. `starknet` stays resolved at 6.11.0. Flagging it because "keep the higher version" was a call, not a forced outcome.

**917baa3 is deliberately still here.** Today's testing suggests the `Promise.race` timeout can come out once snapshot.js #1225 ships, but #1225 is not released, so I preserved the commit rather than dropping it as part of the rebase.

Two overlaps for whoever merges: #509 also edits `package.json`/`yarn.lock` (the starknet pin), and #504 edits the same `normalizeAddresses` this branch renames the shape check in. Whichever of each pair lands second needs a small follow-up.

Worth noting this PR is currently fixing live breakage: `api.starknet.id` is refusing connections right now, and master's integration suite fails all four Starknet address-resolver cases because of it.

Gate after the rebase: typecheck, lint and 49 unit tests pass; integration goes from master's 11 failures to 7, the difference being exactly the four Starknet address-resolver cases this PR fixes. e2e is green (29 passed).

src/addressResolvers/utils.ts
```diff
@@ -8,9 +9,34 @@ export function isEvmAddress(address: Address): boolean {
...
+// `'a'.repeat(47) + 'b'` resolves and `'a'.repeat(48)` is over-prime, both 48 characters. One
+// over-prime label fails the *whole* multicall with a `-32602` that `isSilencedError` does not
+// silence, taking down every name batched alongside it.
+export function isStarkDomain(domain: Handle): boolean {
+  if (!/^[a-z0-9-这来]+(\.[a-z0-9-这来]+)*\.stark$/.test(domain)) return false;
+
+  return domain
+    .replace(/\.stark$/, '')
+    .split('.')
+    .every(label => starknetId.useEncoded(label) < constants.PRIME);
```
wa0x6e: **Unbounded label length lets one `resolve_names` request block the event loop for minutes.** `isStarkDomain` hands any regex-matching label to `starknetId.useEncoded`, and nothing upstream bounds label length — `resolve_names` is `z.array(z.string()).max(5)` with no per-string cap, the Express body limit is 4 MB, and the shared `normalizeHandles` only rejects whitespace. `useEncoded`'s BigInt loop grows worse than quadratically with label length (measured on the pinned starknet 6.21.1: 256k chars → 6.9 s, 512k → 38.8 s, **1M → 136 s**), and it runs synchronously before the first `await`, so a single unauthenticated request wedges the whole process for the duration — the endpoint has no auth and no rate limit.

A 48-character cap per label closes this at zero cost: every 49+-character label is over-prime regardless of content (brute-forced the minimum-value construction against `constants.PRIME`; consistent with the existing `OVER_PRIME_49` fixture), so the cap rejects only labels the felt check was going to reject anyway — it just refuses to pay for the encoder first. Bounded quantifiers can't backtrack pathologically here since the class excludes `.`, so label boundaries are deterministic — worst adversarial 4 MB input costs 63 ms in the regex, and the worst payload that still passes the gate (2M one-char labels) encodes in ~242 ms, versus 136 s today.

```suggestion
// Labels are also capped at 48 characters in the regex: every 49+ character label is over-prime
// whatever its content, so the cap rejects nothing the felt check would keep -- and it stops a
// multi-megabyte label from reaching `useEncoded`, whose worse-than-quadratic BigInt growth
// would otherwise block the event loop for minutes (measured ~136s at 1M characters) on a
// single unauthenticated `resolve_names` request.
export function isStarkDomain(domain: Handle): boolean {
  if (!/^[a-z0-9-这来]{1,48}(\.[a-z0-9-这来]{1,48})*\.stark$/.test(domain)) return false;

  return domain
    .replace(/\.stark$/, '')
    .split('.')
    .every(label => starknetId.useEncoded(label) < constants.PRIME);
```

Worth a regression test pinning that the encoder is never reached — the existing `OVER_PRIME_49` case passes via the felt check either way, so it doesn't prove the cheap path:

```ts
it('drops an oversized label without ever calling the encoder', async () => {
  const spy = jest.spyOn(starknetId, 'useEncoded');

  await expect(resolveNames([`${'a'.repeat(100_000)}.stark`])).resolves.toEqual({});

  expect(spy).not.toHaveBeenCalled();
  expect(mockCallContract).not.toHaveBeenCalled();
});
```

src/addressResolvers/starknet.ts
```diff
@@ -1,63 +1,120 @@
...
-  return `${BASE_URL}/${resolve_type}?${
-    resolve_type === 'addr_to_domain' ? 'addr' : 'domain'
-  }=${needle}`;
+// Shape only. The felt bound is guarded upstream, in the shared `normalizeAddresses` that
+// addressResolvers/index runs before any resolver sees the input; what is left here is the
+// part that guard does not do, rejecting EVM addresses, which are valid felts themselves.
+// The composition is pinned by test/unit/addressResolvers/starknet-batch-guard.test.ts.
+function normalizeAddresses(addresses: Address[]): Address[] {
+  return addresses.filter(hasStarknetAddressShape);
 }
```
wa0x6e: Defense in depth, take or leave — the shared hub guard makes this safe today, so this is about the resolver's self-containment, not a live bug. Any future direct caller of the exported `lookupAddresses` that passes a shape-valid but over-felt 64-hex value (a tx hash, a proposal id — about 97% of that shape) poisons the whole multicall with a `-32602` that `isSilencedError` doesn't silence, and the invariant preventing that lives two modules away in `addressResolvers/index.ts`. `isStarknetAddress` is exported from the same utils module already imported here and does both halves (shape + felt range) at negligible cost — pure computation, no I/O — so the resolver stays correct even if the hub composition ever changes:

```suggestion
function normalizeAddresses(addresses: Address[]): Address[] {
  return addresses.filter(isStarknetAddress);
}
```

(Import swap on line 5 posted as its own suggestion. If you take both, the comment's pointer to `starknet-batch-guard.test.ts` goes with it — that test keeps pinning the hub composition either way.)

src/addressResolvers/starknet.ts
```diff
@@ -1,63 +1,120 @@
-import axios from 'axios';
-import { isStarknetAddress, withoutEmptyValues } from './utils';
+import snapshot from '@snapshot-labs/snapshot.js';
+import { CallData, constants, starknetId, validateAndParseAddress } from 'starknet';
+import {
+  provider as getProvider,
+  hasStarknetAddressShape,
```
wa0x6e: Companion to the `normalizeAddresses` suggestion below — the import swap:

```suggestion
  isStarknetAddress,
```

src/addressResolvers/starknet.ts
```diff
@@ -1,63 +1,120 @@
...
+}
+
+async function callNamingContract(entrypoint: string, calldata: string[][]): Promise<any[]> {
+  return await snapshot.utils.multicall(
+    NETWORK,
+    provider,
+    NAMING_ABI,
+    calldata.map(args => [NAMING_CONTRACT, entrypoint, args])
+  );
 }
```
wa0x6e, line 84: **Flagging for a follow-up issue, not this PR: one on-chain panic fails the whole name-resolution batch.**

The naming contract dispatches to a per-domain, opt-in custom resolver when one is set on an ancestor label, and some of those resolver contracts panic when called with the empty `hint` this calldata sends (same calldata starknet.js's own `getAddressFromStarkName` sends — the CCIP hint-retry lives in wrapper libs, not starknet.js core). Because both directions go through one atomic `aggregate` call, that panic fails every name batched alongside, and `isSilencedError` doesn't match the resulting `LibraryError` (the JSON-RPC code is baked into `.message`; no structured `.code` survives), so each occurrence is also a Sentry capture.

Verified live against `rpc.snapshot.org/sn` with this PR's exact call path: `braavos.stark`, `xplorer.stark`, `sol.stark`, `gg.stark` and `ctrl.stark` all carry a non-zero resolver today (enumerable by anyone via `DomainResolverUpdate` events), and batching `vitalik.stark` with a *guessed* `b.gg.stark` — both pass `isStarkDomain` — throws `40: Contract error` for the whole batch. Plain unregistered domains are safe (they return `0x0` without dispatching); the trigger needs a subdomain under a resolver-bearing parent — but a legitimate Cartridge/controller-style off-chain name triggers it as innocently as an attacker's guess. The blanked batch then lands in Redis for the 12 h negative-cache TTL, so the good names stay blank well past the one request.

Suggested direction for the follow-up: keep the atomic multicall as the fast path, catch only errors matching `': Contract error:'` (a Cairo revert, never a transport fault — those must stay loud), and retry the batch per-call with `Promise.allSettled`, mapping a per-call revert to "not found" like an empty span. Trade-off to weigh there: a non-revert failure inside the fallback gets swallowed for that one name — narrower than the pre-PR silent-`{}` behavior, but worth stating in the implementation.

src/addressResolvers/starknet.ts
```diff
@@ -1,63 +1,120 @@
...
+    .replace(/\.stark$/, '')
+    .split('.')
+    .map(label => starknetId.useEncoded(label));
 
-        return [needle, value];
-      })
-    )
-  );
+  return CallData.compile({ domain, hint: [] });
 }
```
wa0x6e, line 61: The label decomposition here must stay byte-identical to the one in `isStarkDomain` — the guard validates the felts this builder later sends, so any drift between the two splitters (a `.trim()`, a separator change) lets a guard-approved handle decompose differently at send time, reintroducing the batch-killing over-prime `-32602` the guard exists to prevent. Worth extracting the shared step into `utils.ts` next to `isStarkDomain`:

```ts
// Shared with `domainToAddressCalldata` in addressResolvers/starknet.ts so the labels validated
// here and the labels later sent to the contract can never drift apart.
export function starkDomainLabels(domain: Handle): string[] {
  return domain.replace(/\.stark$/, '').split('.');
}
```

with `isStarkDomain` returning `starkDomainLabels(domain).every(label => starknetId.useEncoded(label) < constants.PRIME);` (composes with the label-length suggestion above — only the regex line differs), and this side becoming:

```suggestion
function domainToAddressCalldata(handle: Handle): string[] {
  const domain = starkDomainLabels(handle).map(label => starknetId.useEncoded(label));

  return CallData.compile({ domain, hint: [] });
}
```

(plus `starkDomainLabels` joining the `./utils` import.)

wa0x6e:
**Design note for the follow-up queue (pre-existing mechanism, amplified by this PR's batching):** the negative cache multiplies any batch-wide failure. `_call` writes `''` with `ttl: 43200` for every input a resolver run leaves unresolved, and `getCache` serves `''` as a hit — under master's per-name fan-out a transient failure blanked one name for 12 h; with resolution now a single atomic multicall, a batch-wide failure that surfaces as an empty result (the resolver-panic case in the review thread on `callNamingContract` is a measured example) blanks every name in the batch for 12 h. "Loud instead of silent" holds for the Sentry event but not for the user-visible result, which no longer self-heals on the next request.

The fix belongs in `addressResolvers/index.ts`/`cache.ts` (skip the negative-cache write when a resolver threw), so it's out of scope here — flagging so it rides the same follow-up as the resolver-panic issue rather than being rediscovered from a 12-hour blank-name report.

src/addressResolvers/starknet.ts
wa0x6e: Resolved by the dependency route instead of the app-level bound: the rebased branch drops `917baa3` and raises the `@snapshot-labs/snapshot.js` floor to 0.15.2, whose `getStarknetProvider` now wraps the transport with a `withTimeout`-bound `baseFetch` honoring `options.timeout` and raising `code: 'TIMEOUT'` — which `isSilencedError` already treats like the other resolvers' timeouts. Verified against the published 0.15.2 dist: the timeout is forwarded (with NaN/overflow guards), restoring parity with `getEvmProvider` for every consumer. That's the upstream fix this thread asked for; closing.

wa0x6e (approved):
tAck

PR #497, opened 2026-08-03
fix(lookupDomains): stop reporting Unstoppable Domains rate limits

### LINKED ISSUE #494: Sentry STAMP-6X: Unstoppable Domains HTTP 429 floods Sentry
https://snapshot-labs.sentry.io/issues/STAMP-6X — 251 events in ~6s on
2026-08-03, escalating.

`fetchDomains` in `src/lookupDomains/unstoppableDomains.ts` throws on
`!response.ok` (added in #489). During a traffic burst the UD API key
gets rate-limited and every chain-146 lookup throws HTTP 429.

429 is an expected upstream condition (there is no throttle, cache, or
retry at this layer, and the pagination loop can issue several calls per
address), but it is not in `isSilencedError`
(`src/addressResolvers/utils.ts`) — so every rate-limited request is
captured. One burst produced 251 events.

### Proposed fix

Stop capturing 429s: add to `isSilencedError` or skip `capture` when
status is 429. `FetchError` still propagates, behavior unchanged.

Optional: honor `Retry-After` with one retry if 429s are frequent enough
to hurt results.

Related: #493 (lookupDomains not best effort — same incident made every
lookup_domains request 500).

tony8713:
## Summary

Reproduced STAMP-6X through the real resolver with `fetch` mocked to 429: one Sentry capture per request, all under a single signature, which is the 251 events in ~6s.

Fixes #494

## Root cause

Not that 429 is missing from the list. The error `fetchDomains` throws carries its status only inside the message text, and `isSilencedError` reads status as data (`error.code`, `error.response?.status`), never as prose. So no HTTP status from this resolver could be silenced, and a 504 was captured too, although 504 is already silenced by policy everywhere else.

Before (on `master`):

```
mode=mock429  requests=50   Sentry captures: 50 / 50
mode=mock504  requests=10   Sentry captures: 10 / 10
mode=live401  requests=5    Sentry captures: 5 / 5
```

After:

```
mode=mock429  requests=50   Sentry captures: 0 / 50
mode=mock504  requests=10   Sentry captures: 0 / 10
mode=live401  requests=5    Sentry captures: 5 / 5
```

`live401` is a real request to `api.unstoppabledomains.com` with an invalid key. Matching the message string instead would have fixed the 429 and left the identical 504 flood in place.

## Changes

- `fetchDomains` attaches `status` to the error it throws.
- `isSilencedError` reads `error.status`, and treats 429 as transient alongside 504. This also silences axios-shaped 429s (`error.response.status`) for every other resolver.

401 and 500 stay visible. HTTP responses are unchanged: a failing resolver still rejects the whole `lookup_domains` request, which is #493 / #448.

## Tests

5 tests fail on `master` and pass here: 3 in `isSilencedError` (429 and 504 on `error.status`, axios 429) and 2 in the resolver suite (429 and 504 no longer captured). Two more assert the fix does not over-silence: a 401 stays unsilenced and a 500 is still reported with its descriptive message.

Full suite is unchanged against `master`: same 10 failures in the same 4 suites (4 starknet `api.starknet.id`, 1 ENS CCIP-Read, 5 shibarium / Unstoppable from missing local API keys), 177 -> 183 passing.

## Adjacent finding, not changed here

`src/lookupDomains/shibarium.ts:63` captures with no `isSilencedError` filter at all, so a D3 rate limit still floods the same way. Left alone as a separate behaviour decision.

### DIFF
--- a/src/addressResolvers/utils.ts
+++ b/src/addressResolvers/utils.ts
@@ -67,6 +67,7 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
     error.error?.code,
     error.error?.status,
     error.code,
+    error.status,
     error.response?.status,
     error.cause?.code
   ];
@@ -77,7 +78,7 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
         error.error?.message?.includes(m) ||
         error.cause?.message?.includes(m)
     ) ||
-    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET', 504].some(c =>
+    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET', 504, 429].some(c =>
       codes.some(v => String(v ?? '').includes(String(c)))
     )
   );
--- a/src/lookupDomains/shibarium.ts
+++ b/src/lookupDomains/shibarium.ts
@@ -1,5 +1,6 @@
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import fetch from 'node-fetch';
+import { isSilencedError } from '../addressResolvers/utils';
 import constants from '../constants.json';
 import { Address, Handle } from '../utils';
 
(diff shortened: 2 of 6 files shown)


wa0x6e (approved):
utAck

Issue #496, opened 2026-08-03
refactor: centralize resolver error capture in index.ts, remove per-resolver try/catch and the FetchError sentinel


wa0x6e:
## Roadmap context

Step **3 of 3** in cleaning up resolver error handling. End goal: resolvers throw raw errors, orchestrators own capture/silence policy, `FetchError` sentinel deleted, api.ts catch-all is a pure safety net for unexpected errors, and `lookup_domains` is best-effort.

1. **#492 (merged)** — stopgap: api.ts skips re-capturing `FetchError`, class gets a name/message. Fixed the Sentry noise (#488, STAMP-6V / STAMP-35) at the boundary, changed no behavior.
2. **#448** — behavior change, isolated on purpose: `lookupDomains` absorbs per-resolver failures, one failing chain returns `[]` instead of rejecting the whole request (partial 200 instead of 500). **Blocks the `lookupDomains` half of this issue.**
3. **This issue** — refactor, no behavior change once #448 has landed: move capture into the orchestrators, delete per-resolver try/catch and the `FetchError` sentinel entirely.

## Context

Currently every resolver repeats the same catch block (~16 sites):

```ts
} catch (err) {
  if (!isSilencedError(err)) capture(err, { input });
  throw new FetchError();
}
```

`FetchError` is a "handled, give up on this resolver" sentinel whose only purpose is to cross the `index.ts` and `api.ts` boundaries without being re-reported (#492 added the `instanceof` skip in api.ts for exactly that). Once the orchestrators own the catch, the sentinel has no job left.

## Proposed shape

**`addressResolvers/index.ts`** — independent of #448, no behavior change, can ship any time. `_call` already swallows per-resolver errors with a bare `catch {}`, so capture just moves into it:

```ts
try {
  result = await r[fnName](_input);
  status = 1;
} catch (err) {
  if (!isSilencedError(err, r.MUTED_ERRORS)) capture(err, { input: { [fnName]: _input } });
}
```

- Resolvers stop catching entirely; ~14 catch blocks deleted.
- `lens.ts` passes an extra muted-errors list to `isSilencedError` — resolver exports `MUTED_ERRORS`, index reads it.
- Context granularity: resolvers currently capture *normalized* input; index has pre-normalization `_input`. Acceptable — normalization is deterministic.

[...]

PR #498, opened 2026-08-03
fix(lookupDomains): stop reporting Shibarium rate limits

tony8713:
## Summary

`src/lookupDomains/shibarium.ts` captured every pagination failure with no `isSilencedError` filter, so a D3 rate limit or gateway timeout floods Sentry exactly like STAMP-6X did for Unstoppable Domains.

## Root cause

Two halves, and only both together fix it. The guard was missing, and the thrown error carried its status only inside the message text, so `error.status` was undefined and `HTTP 504: Gateway Timeout` never matched the existing `status=504` message. Adding the guard on its own changes nothing:

```
guard only, no status attached
mode=mock429  requests=50  Sentry captures: 50 / 50
mode=mock504  requests=10  Sentry captures: 10 / 10
```

## Verification

Real resolver, D3 endpoint mocked, counting `capture()` calls.

Before:

```
mode=mock429    requests=50   Sentry captures: 50 / 50
mode=mock504    requests=10   Sentry captures: 10 / 10
mode=mock401    requests=5    Sentry captures: 5 / 5
mode=mockdns    requests=5    Sentry captures: 5 / 5
mode=live401    requests=5    Sentry captures: 5 / 5
```

After:

```
mode=mock429    requests=50   Sentry captures: 0 / 50
mode=mock504    requests=10   Sentry captures: 0 / 10
mode=mock401    requests=5    Sentry captures: 5 / 5
mode=mock500    requests=5    Sentry captures: 5 / 5
mode=mockdns    requests=5    Sentry captures: 5 / 5
mode=live401    requests=5    Sentry captures: 5 / 5
```

`live401` is a real request to `api-public.interstellar.xyz` with an invalid key, `mockdns` an `ENOTFOUND` with no HTTP status. Both stay visible, so this does not over-silence.

## The outer catch

It gets the same guard. It is near dead today, only the `AbortController` and the `setTimeout` sit outside the inner `try`, but it is one line, it changes nothing for a genuine error, and leaving one guarded and one unguarded `capture` in the same file is how this flood comes back the moment anything moves above the inner `try`.

## Depends on #497

Stacked on `fix/494-silence-unstoppable-domains-rate-limit`. The guard only works because #497 makes `isSilencedError` read `error.status` and treat 429 as transient. `utils.ts` is untouched here. Merge #497 first.

## Scope

Behaviour fix only. Shibarium still swallows and returns partial results instead of rethrowing (#448), and nothing is centralised (#496).

## Tests

8 new tests in `test/integration/lookupDomains/shibarium.test.ts`, 5 of which fail without the `src` change. CI is unchanged against the base branch: the same 5 pre-existing failures (4 starknet, 1 ENS CCIP-Read), 188 -> 196 passing.

### DIFF
--- a/src/lookupDomains/shibarium.ts
+++ b/src/lookupDomains/shibarium.ts
@@ -1,5 +1,6 @@
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import fetch from 'node-fetch';
+import { isSilencedError } from '../addressResolvers/utils';
 import constants from '../constants.json';
 import { Address, Handle } from '../utils';
 
@@ -44,7 +45,9 @@ export default async function lookupDomains(
         }
 
         if (!response.ok) {
-          throw new Error(`HTTP ${response.status}: ${response.statusText}`);
+          throw Object.assign(new Error(`HTTP ${response.status}: ${response.statusText}`), {
+            status: response.status
+          });
         }
 
         let data: { pageItems?: Array<{ sld: string; tld: string }> };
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

PR #500, opened 2026-08-04
refactor(addressResolvers): capture resolver errors in index.ts

tony8713:
First half of #496. No behaviour change on the JSON-RPC path, and independent of #448 as the issue says.

Every resolver repeated the same block: catch, decide whether to report, then throw a `FetchError` purely so the failure could cross the `index.ts` boundary without being reported twice. `_call` already swallows those rejections with a bare `catch {}`, so the sentinel carried an "already handled" flag the orchestrator never read.

The capture moves into `_call`'s catch and the resolvers throw their original errors. 14 catch blocks go, and with them 14 of the 16 `throw new FetchError()` sites. Silence policy stays with the resolver: `lens.ts` exports `MUTED_ERRORS` and `index.ts` passes it to `isSilencedError`. `RESOLVERS` picks up an explicit `Resolver` type so `MUTED_ERRORS` is a real optional field rather than an untyped lookup.

### Three catch blocks stay

They are partial-failure handling, not the sentinel pattern. They capture and continue instead of giving up, so `index.ts` never sees the error and could not report it:

- `ens.ts` `resolveNames`, subgraph path and provider path — either can fail without losing the other's results
- `unstoppableDomains.ts` `resolveNames`, per handle — one unresolvable handle must not drop the batch

Same reasoning the issue already applies to `lookupDomains/shibarium.ts`'s inner capture.

### One behaviour change, worth a look

`basename.ts` `getAvatar` reaches `collect()` from outside the orchestrator, and its only caller `resolvers/basename.ts` swallows with a bare `catch {}`. So Basename avatar failures stop being reported. The sibling `call('text', ...)` on that same path already reported nothing, and every other `src/resolvers/*` swallows the same way, so this makes Basename consistent rather than odd. Say the word if you would rather keep a capture there.

### FetchError

Still needed by `lookupDomains`' 2 remaining sites, so the class stays here. It goes in the follow-up.

### Tests

One new file, `test/unit/addressResolvers.test.ts`, on the one concern this PR moves: who reports a resolver failure. Five cases — captured once with the input as context, silenced errors not captured, `MUTED_ERRORS` honoured, `MUTED_ERRORS` scoped to the resolver that exports it, and one failing resolver not taking the others down. Both halves are mutation-checked: reverting the capture fails 2 of them, dropping `r.MUTED_ERRORS` from the `isSilencedError` call fails a third.

`test/integration/addressResolvers/ens.test.ts` had one case asserting the old contract (`rejects.toBeInstanceOf(FetchError)` plus a resolver-level capture). It now asserts the raw error is rethrown untouched and nothing is captured at the resolver. That is the only edit to that file — its unrelated live-network failure is untouched.

### Gate

- `yarn typecheck` clean
- `yarn lint` clean
- `yarn jest test/unit` — 10/10, 3 suites
- `test/integration/addressResolvers/utils.test.ts` green

CI will still show `addressResolvers/ens`, `addressResolvers/starknet` and `resolvers/zapper` red. Those are the live-network suites that have been failing since 2026-07-07, untouched here.

### Base

Stacked on #499 so this PR's CI is not red for a reason that has nothing to do with it — master's `test/unit/api.test.ts` is currently broken by the #448 merge and #499 is the two-line fix. Retarget to master once #499 lands. Nothing in this diff depends on it otherwise.

Refs #496

### DIFF
--- a/src/addressResolvers/basename.ts
+++ b/src/addressResolvers/basename.ts
@@ -1,9 +1,8 @@
 import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
 import { namehash } from '@ethersproject/hash';
-import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
-import { FetchError, provider as getProvider, isEvmAddress, isSilencedError } from './utils';
+import { provider as getProvider, isEvmAddress } from './utils';
 import { Address, EMPTY_ADDRESS, getUrl, Handle } from '../utils';
 
 export const NAME = 'Basename';
@@ -44,13 +43,8 @@ async function collect(
   items: string[],
   query: (item: string) => Promise<[string, string]>
 ): Promise<Record<string, string>> {
-  try {
-    const entries = await Promise.all(items.map(query));
-    return Object.fromEntries(entries.filter(([, value]) => value));
-  } catch (err) {
-    if (!isSilencedError(err)) capture(err, { input: items });
-    throw new FetchError();
-  }
+  const entries = await Promise.all(items.map(query));
+  return Object.fromEntries(entries.filter(([, value]) => value));
 }
 
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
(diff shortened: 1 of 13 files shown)


wa0x6e (approved):
utAck

tony8713:
Only the naming, and that's #502 now. #500 never touched `api.test.ts` (that hunk was #499's two lines showing through a stale merge base), and the assertions themselves hold up: mutation-checked on master, dropping the capture in `lookupDomains/index.ts`, doubling it, dropping the `isSilencedError` guard, or letting the rejection reach `api.ts` each fail one or both of those cases. What was stale is that the describes still said the resolver reports and silences its own error, which #501 moved into `index.ts`.

PR #501, opened 2026-08-04
refactor(lookupDomains): capture in index.ts, drop the FetchError sentinel

### LINKED ISSUE #496: refactor: centralize resolver error capture in index.ts, remove per-resolver try/catch and the FetchError sentinel

tony8713:
Second half of #496, on top of #500. Closes the issue.

#448 made `lookupDomains` catch each resolver rejection and return `[]`, so capture + swallow is the contract now. This moves the capture into that same per-promise catch: the resolvers throw their original errors and `index.ts` decides what to report, with `{ address, chainId }` as context, exactly as the issue proposes.

That empties the last 2 of the 16 `throw new FetchError()` sites, so the sentinel goes: the class, the `api.ts` `instanceof` skip #492 added, and the import. `api.ts`'s catch-all is a plain safety net for unexpected errors again.

### Response bodies do not change

`FetchError` set `this.name`, an own enumerable property, so `rpcError`'s `data: e` serialized it into 500 bodies. That was real, but #448 already removed the only path that put one there. Verified by running the same two requests against master and this branch:

| request | master | this branch |
|---|---|---|
| `lookup_domains`, resolver throws | `200 {"jsonrpc":"2.0","result":[],"id":null}` | identical |
| `get_owner`, unexpected error | `500 ... "data":{}` | identical |

For contrast, the same `lookup_domains` request with #448 reverted gives `500 ... "data":{"name":"FetchError"}`. So the leak the issue wants gone was already gone; deleting the class removes dead code rather than changing output.

### Two things this drops, both worth a look

- **`lookupDomains/ens.ts` loses its response context.** Its capture carried `response.status` and `response.data` alongside the input; the index capture is `{ address, chainId }` per the issue. That is a real downgrade for debugging subgraph 5xx. One line to put back if you want it.
- A stray `console.log(err)` in that same catch goes with it.

### `lookupDomains/shibarium.ts` is left alone

It never rejects — it captures and returns the pages it already has — so `index.ts` would never see its error and could not report it. Its inner capture also carries the pagination `skip`, which `index.ts` has no access to. This is the "keep that one inner capture" option from the issue.

### Tests, split by concern

- `test/unit/lookupDomains.test.ts` gains a second describe for reporting: captured once with `{ address, chainId }`, silenced errors not captured, one event per failing chain. The #448 cases stay as they were, with plain `Error`s instead of `FetchError`.
- `test/integration/lookupDomains/unstoppableDomains.test.ts` now asserts what the resolver actually owns: that the thrown error carries its HTTP status **as data**, which is the whole point of 455b86d — `isSilencedError` reads the status off the error and never out of the message text. Which statuses are silenced is already covered in `addressResolvers/utils.test.ts`, and whether a capture happens is now the index's business. Net 81 lines down to 45.

### Gate

- `yarn typecheck` clean
- `yarn lint` clean
- `yarn jest test/unit test/integration/lookupDomains test/integration/addressResolvers/utils.test.ts` — 6 suites green, 55 tests
- `test/integration/lookupDomains.test.ts` has 3 failures locally, byte-identical on master, from missing `D3_API_KEY_MAINNET` / `UNSTOPPABLE_DOMAINS_API_KEY` in my environment. It passes on CI.

CI will still show the three live-network suites red, as on master.

Closes #496

### DIFF
--- a/src/addressResolvers/basename.ts
+++ b/src/addressResolvers/basename.ts
@@ -1,9 +1,8 @@
 import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
 import { namehash } from '@ethersproject/hash';
-import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
-import { FetchError, provider as getProvider, isEvmAddress, isSilencedError } from './utils';
+import { provider as getProvider, isEvmAddress } from './utils';
 import { Address, EMPTY_ADDRESS, getUrl, Handle } from '../utils';
 
 export const NAME = 'Basename';
@@ -44,13 +43,8 @@ async function collect(
   items: string[],
   query: (item: string) => Promise<[string, string]>
 ): Promise<Record<string, string>> {
-  try {
-    const entries = await Promise.all(items.map(query));
-    return Object.fromEntries(entries.filter(([, value]) => value));
-  } catch (err) {
-    if (!isSilencedError(err)) capture(err, { input: items });
-    throw new FetchError();
-  }
+  const entries = await Promise.all(items.map(query));
+  return Object.fromEntries(entries.filter(([, value]) => value));
 }
 
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
--- a/src/addressResolvers/ens.ts
+++ b/src/addressResolvers/ens.ts
@@ -2,7 +2,7 @@ import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
-import { FetchError, provider as getProvider, isEvmAddress, isSilencedError } from './utils';
+import { provider as getProvider, isEvmAddress, isSilencedError } from './utils';
 import constants from '../constants.json';
 import { Address, graphQlCall, Handle } from '../utils';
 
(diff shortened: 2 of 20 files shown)


wa0x6e (approved):
utAck

PR #504, opened 2026-08-05
fix(addressResolvers): keep out-of-range 64-hex values out of the hub batch

tony8713:
Fixes STAMP-6W.

A 64 hex-digit value is only a Starknet address when it is below the felt bound, and a transaction hash or a proposal id has exactly the same shape. `normalizeAddresses` was forwarding those to the hub, which rejects the whole `id_in` batch when one arrives — so a single bad value nulls the hub names of every address queried alongside it, and they are then cached as "no name".

`isStarknetAddress` range-checks as well as shape-checks now, so they never reach the hub.

Merge order: #491 delegates its own felt-range guard to this one, so it needs this to land first.

### DIFF
--- a/src/addressResolvers/utils.ts
+++ b/src/addressResolvers/utils.ts
@@ -8,20 +8,9 @@ export function isEvmAddress(address: Address): boolean {
   return /^0x[a-fA-F0-9]{40}$/.test(address);
 }
 
+// `snapshot.utils.isStarknetAddress` is a felt-range check: `true` for an EVM address too.
 export function isStarknetAddress(address: Address): boolean {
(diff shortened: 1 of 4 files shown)


tony8713:

Heads-up that #491 renames `isStarknetAddress` to `hasStarknetAddressShape` in this same file, so whichever of the two lands second needs a one-line touch-up. The two fixes are complementary rather than duplicates: this one guards the hub batch in the shared `normalizeAddresses`, #491 guards the StarknetID call in its own local one.

wa0x6e (approved):
utAck

PR #506, opened 2026-08-05
feat(utils): validate the GraphQL response envelope in graphQlCall

### LINKED ISSUE #272: feat: validate GraphQL response envelopes (9 blind data.data destructures)
Nine sites destructure `data.data.X` straight off a GraphQL response with no check that the envelope is there. An upstream that answers HTTP 200 with `{"errors":[…],"data":null}` — or with no top-level `data` key at all — does not make axios throw, so the destructure throws a TypeError naming the field we wanted. The stack trace points at our code and says nothing about the upstream failure that caused it.

All nine go through `graphQlCall` (`src/utils.ts:186`), and those nine are its only callers, so a single check closes the whole class.

## Sites (master `b19d9c2`)

- `src/addressResolvers/snapshot.ts:8` — `users`
- `src/addressResolvers/ens.ts:78` — `domains`
- `src/addressResolvers/lens.ts:23` — `accountsBulk`
- `src/lookupDomains/ens.ts:18` — `data` (optional-chained downstream, so a missing envelope here returns the un-decoded name instead of throwing: same hole, failing quietly)
- `src/lookupDomains/ens.ts:46` — `account`
- `src/resolvers/snapshot.ts:54` — `entry`
- `src/resolvers/snapshot.ts:85` — `spaces`
- `src/resolvers/space-sx.ts:16` — `spaces`
- `src/resolvers/lens.ts:37` — `account`

## Observed, not hypothetical

- **STAMP-6W** — `Cannot read properties of null (reading 'filter')` in `addressResolvers/snapshot.ts`. The hub answers an invalid `id_in` batch with HTTP 200 and `{"errors":[{"message":"Invalid addresses in id"}],"data":{"users":null}}`. Fixed in #504.
- **STAMP-49** — `Cannot read properties of undefined (reading 'account')` in `lookupDomains/ens.ts`. One event, 2026-05-16, dormant since. Tracked here, there is no separate GitHub issue for it. Note that #438 added a diagnostic to capture the response body on the next occurrence, and `145a440` removed it when capture moved to `lookupDomains/index.ts` — waiting for a recurrence will no longer tell us the shape.
- **STAMP-19** / #148 — `Cannot read properties of null (reading 'ownedBy')` in lens. Same shape, patched at the one site in 2023.

## The obvious fix is the wrong one

Adding `?.` and returning an empty result makes the TypeError go away and makes the outage invisible with it. Every user quietly loses their ENS name, nothing is reported, nothing is silenced, and `cache.ts` stores the empty value for 12h. An absent envelope is an upstream failure and has to surface as one.

Two cases that look alike and are not:

- `data.account === null` — legitimate, the address has no record. Already handled by the `account?.` reads downstream. Leave it alone.
- `data` itself absent — the upstream failed. Throw.

So what is needed is an envelope check that throws, not a null guard that swallows.

## What to build

One shared helper that validates the envelope and throws with the upstream's own `errors[]` message, carrying the HTTP status **on the error object**:

```ts
throw Object.assign(new Error(`[${NAME}] ${body.errors[0].message}`), {
  status,
  response: { status }
});
```

Not interpolated into the message string. `isSilencedError` (`src/addressResolvers/utils.ts:47`) reads `error.status` and `error.response.status`, so a status that exists only in prose is unreachable: a 429 or 504 from a GraphQL upstream would stay noisy while the identical axios error is silenced. That is the STAMP-6X lesson.

`graphQlCall` is the natural home. It would have to become `async`; all nine callers already `await` it.

## Already done

#504 does exactly this at `addressResolvers/snapshot.ts` and is the pattern to copy for the other eight.

<details>
<summary>Original issue text</summary>

As discussed here: https://github.com/stamp-labs/stamp/pull/270#discussion_r1696725496

All responses from APIs, REST or GraphQL, should be validated. I suggest using [Zod schemas](https://zod.dev/?id=basic-usage).

```ts
import { z } from 'zod';

async function graphqlCall<T extends z.ZodType>(
  schema: T,
  // ...
): Promise<z.infer<T>> {
  try {
    const jsonResponse = await fetch(endpoint, { ... });

    // Validate the data against the provided schema
    const validatedData = schema.parse(jsonResponse);

    return validatedData;
  } catch (error) {
    // ...
  }
}

// Use
const responseSchema = z.object({
  user: z.object({
    id: z.string(),
    imgUrl: z.string(),
  }),
});

const typedResponse = await graphqlCall(
  responseSchema,
  // ...
)
```

Full per-query schema validation is a bigger job than the envelope check above. The envelope check is the part that is throwing in production today, and it can land first.

</details>

tony8713:
Closes #272.

Every GraphQL call site destructured `data.data.X` straight off the response with no check that the envelope was there. A GraphQL upstream reports its own failures inside an HTTP **200** body, so axios does not throw, and the destructure raised a TypeError naming the field *we* wanted while carrying none of the upstream's evidence. STAMP-6W, STAMP-49 and STAMP-19 are all that one failure.

Those sites all go through `graphQlCall`, and they are its only callers, so the check goes there once. It throws when the body carries `errors`, or when the `data` envelope is absent.

Three things worth pointing at:

**The status goes on the error object, not into the message.** `isSilencedError` reads `error.status` and `error.response.status`. A status that exists only in prose is unreachable to it, so a 429 or 504 from a GraphQL upstream would stay noisy where the identical axios error is silenced. In practice axios rejects anything outside 2xx before this check runs, so the status carried here will be 200. It goes on the object anyway, because that is the only place downstream can read it and because the alternative, a status in the message string, is the exact mistake STAMP-6X was about.

**`source` is the upstream host, not a resolver name.** Most of the call sites have no name to give. The host is uniform across all of them, it names the upstream that actually failed rather than the caller that noticed, and it groups stably in Sentry.

**A null field is not an envelope failure.** `data.account === null` is the upstream saying "no record", it is legitimate, and it still passes straight through to the caller's existing `account?.` reads. Only a missing envelope, or an explicit `errors[]`, throws. Optional chaining was deliberately not used as the fix: returning empty on a missing envelope converts an upstream outage into a silent empty result, every user quietly loses their ENS name, and `cache.ts` then stores that emptiness.

## Behaviour change in `lookupDomains/ens.ts`

Everywhere else this is pure hardening. A TypeError becomes a meaningful error, on the same path, with the same outcome.

`src/lookupDomains/ens.ts` was the wrong version of this fix already. It destructured the envelope and then read it back with `?.`, so a subgraph failure silently returned the raw `[7f3a].eth` label, indistinguishable from a label the subgraph genuinely has not indexed. It now throws like the rest, which means a subgraph outage during label decoding drops that chain's ENS domains instead of serving an undecoded name. Those failures are reported through `lookupDomains/index.ts`, where every resolver failure is already caught and captured. The second query hits the same subgraph as the first, so if it fails the answer we were about to serve was not trustworthy anyway.

A `registration` that is genuinely `null` still falls through to the unchanged name, so the two cases stay separated.

### DIFF
--- a/src/lookupDomains/ens.ts
+++ b/src/lookupDomains/ens.ts
@@ -28,7 +28,12 @@ async function fetchDomainData(domain: Domain, chainId: string): Promise<Domain>
     }`,
     { id: `0x${hash}` }
   );
-  const labelName = data?.registration?.domain?.labelName;
+  // `data?.` used to swallow a missing envelope here and return the raw
+  // `[hash]` label as if the subgraph had simply not known it. graphQlCall now
+  // throws on that, so a subgraph failure is reported rather than served as an
+  // undecoded name. A registration that genuinely has no labelName still falls
+  // through to the unchanged name.
+  const labelName = data.registration?.domain?.labelName;
 
   return {
     ...domain,
--- a/src/lookupDomains/ens.ts
+++ b/src/lookupDomains/ens.ts
@@ -28,9 +28,8 @@ async function fetchDomainData(domain: Domain, chainId: string): Promise<Domain>
     }`,
     { id: `0x${hash}` }
   );
-  // Not `data?.`: a missing envelope must not fall through to the raw `[hash]`
-  // label and make a subgraph outage look like an unindexed name. A `registration`
-  // that is genuinely null still does fall through, to the unchanged name.
+  // Not `data?.`: swallowing a missing envelope here serves the raw `[hash]`
+  // label as though the subgraph had simply not indexed it.
   const labelName = data.registration?.domain?.labelName;
 
   return {
--- a/src/lookupDomains/ens.ts
+++ b/src/lookupDomains/ens.ts
@@ -28,11 +28,9 @@ async function fetchDomainData(domain: Domain, chainId: string): Promise<Domain>
     }`,
     { id: `0x${hash}` }
   );
-  // `data?.` used to swallow a missing envelope here and return the raw
-  // `[hash]` label as if the subgraph had simply not known it. graphQlCall now
-  // throws on that, so a subgraph failure is reported rather than served as an
-  // undecoded name. A registration that genuinely has no labelName still falls
-  // through to the unchanged name.
+  // Not `data?.`: a missing envelope must not fall through to the raw `[hash]`
+  // label and make a subgraph outage look like an unindexed name. A `registration`
+  // that is genuinely null still does fall through, to the unchanged name.
   const labelName = data.registration?.domain?.labelName;
 
   return {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -188,44 +188,22 @@ export type GraphQlResponse<T = any> = {
   errors?: { message?: string }[];
 };
 
-/**
- * Builds the error a failed GraphQL envelope describes.
- *
- * The HTTP status goes *on* the error, at both `status` and `response.status`,
- * because those are the two places isSilencedError() reads. A status
- * interpolated into the message string is unreachable to it, so a rate limit or
- * a gateway timeout would stay noisy where the equivalent axios error is
- * silenced (STAMP-6X).
- */
 function graphQlEnvelopeError(url: string, status: number, message: string) {
   let source = url;
   try {
     source = new URL(url).host;
   } catch {
-    // A non-absolute url is unexpected; naming it in full still beats no source.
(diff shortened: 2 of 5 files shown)


tony8713:

Worth flagging that #457 rewrites `graphQlCall` to native fetch and drops the `AxiosResponse` return type this PR builds on, so those two will conflict properly whichever order they merge in.

wa0x6e (approved):
utAck

Issue #495, opened 2026-08-03
fix(api): image request hangs with no 500 when the cached image read or resize fails


wa0x6e:
When the S3 cache returns a broken image, or the S3 read fails partway, an avatar/space/token image request gets no response. The client waits until it times out instead of getting a 500, and nothing reaches Sentry. A failed read on an already-cached response can also take the process down.

## Reproduction

1. Request any image once, e.g. `GET /avatar/0x…?s=64`, so its base image is cached.
2. Overwrite that base cache object (`public/stamp-7/<key>/<key>` with `<key>` for size `constants.max`) with non-image bytes.
3. Request the same image at a size not yet cached, e.g. `?s=65`.

Expected: `500` with an error body, Sentry event.
Actual: the request hangs; the process logs an unhandled promise rejection from sharp.

## Cause

Express is pinned to 4 (`"express": "^4.17.1"`), which does not forward a rejected async handler promise to the error middleware. The `GET /:type/:id` handler (`src/api.ts:77`) has no try/catch anywhere in its body, so any throw is an unhandled rejection and `res` is never written:

- `src/api.ts:111`: `streamToBuffer(base)` rejects on an S3 stream error.
- `src/api.ts:148`: `resize(baseImage, …)` throws when the cached base image cannot be decoded (the reproduction above).
- `src/api.ts:140`: fallback (blockie/jazzicon) render and resize. Deterministic, low risk.
- `src/api.ts:104`: `cache.pipe(res)` attaches no `'error'` listener to the S3 stream; a mid-stream failure emits an unhandled `'error'`, which crashes the process rather than hanging the request.

Resolver calls are not in this list: each resolver is wrapped to catch, report, and return `false`.

Not a regression. The resize and cache-read path has run unguarded since [b3b6cd5](https://github.com/snapshot-labs/stamp/commit/b3b6cd51fb4fff34d2e8ac0ab153b3aa7739909d).

## Fix

Wrap the handler body in try/catch: `capture(err)` and, if headers are not yet sent, reply `500` in the same shape as the other routes (`{ status: 'error', error: … }`). Add an `'error'` listener on the cached stream before `cache.pipe(res)` that captures and ends the response.

PR #507, opened 2026-08-06
refactor(resolvers): capture resolver errors in index.ts

tony8713:
Third leg of the resolver error-capture refactor tracked in #496. `addressResolvers` (#500) and `lookupDomains` (#501) are on master; #496 scoped itself to those two orchestrators and left the image resolvers in `src/resolvers/` out. This brings them under the same shape — without that history it reads as a duplicate of a change already merged.

### What was broken

Every image resolver ended in `catch { return false }`. A resolver dying and an address simply having no avatar were the same event: blockie fallback, nothing reported. A resolver could be dead upstream indefinitely and the only symptom would be that people get blockies.

### The fix

The resolvers throw, and the capture moves into `src/resolvers/index.ts`, which reports and then answers `false` on the resolver's behalf.

`src/resolvers/*` export bare default functions rather than modules with a `NAME`, so unlike `addressResolvers` there is nothing for index.ts to read a name off. It keeps a small `{ name, fn, mutedErrors }` registry instead and wraps each entry. The default export is still a name → callable map that answers `false` when a resolver cannot answer, so `api.ts` and its callers did not have to change, and a resolver still never throws into the unguarded image route (#495 is neither fixed nor made worse here).

The catches that answer "this input is not an address" stay where they are: that is no data, not a failure.

### "No data" that used to reach index.ts as a rejection

These had to be fixed first, or the wrapper would page someone on a perfectly normal request.

**space-sx** is the one to understand before judging the wrapper over-broad. `getSpaceProperty` rejected with a literal `false` to mean "not on this API", and the two APIs were aggregated with `Promise.any`, so under a plain wrapper every space without an avatar reports an `AggregateError` of `false` values. It returns `null` now, and the resolver takes the first non-null of an `allSettled` fan-out. A rejection is rethrown only when no API produced a value, so one API being down still cannot cost us an avatar the other one has: the resilience `Promise.any` gave us is kept.

The rest:

- **starknet** — `normalizeAddress` threw on anything that is not a felt252, i.e. on bad user input. Returns `null`.
- **coingecko** — fetching the image URL of a coin that has no image threw `Provided config url is not valid`, as if the request itself had failed. Guarded. A 404 is read as "no such coin" (the API's own `{"error":"coin not found"}`), while any other non-2xx is raised rather than read as an empty body — otherwise a rejected API key would be invisible.
- **onchain space ids** — `getAddress()` threw on any id that is not an address, so a junk id in the URL reported. A shared `spaceIds()` helper returns `null` instead.
- **space-logo on an onchain network** — `SpaceMetadataItem` has no `logo` field, so asking the onchain API for one is a GraphQL validation failure on every single request rather than a missing logo. There is nothing to ask for, so it no longer asks. Same `false` as before, minus the doomed round trip.

### Muting

Per resolver rather than global, through a `MUTED_ERRORS` the resolver exports and the registry reads:

- `trustwallet` — the assets repo 404s for every token it has no logo for, which is most of them.
- `ens` — a name with no avatar record falls through to `metadata.ens.domains`, which answers 404 as well.

Deliberately not muted: zapper's 403 and a rejected coingecko key. Those are a resolver being down, which is exactly what should be visible.

### Tests

A unit suite covers the wiring itself: pass-through, `false` not reported, an error captured with the resolver and its arguments as context, silenced and muted errors, and muting scoped to the resolver that exports it.

The shared integration helper's no-avatar cases now also assert that nothing was reported. That is the guarantee this PR exists for, and it is asserted against the live upstreams rather than a mock.

### CI is not green

`Test` is already red on master, on live-network suites this PR does not touch, and it is red here too. This branch does add a failure of its own, and I would rather not hide it: zapper's no-avatar case.

`storage.googleapis.com/zapper-fi-assets` answers `403 UserProjectAccountProblem` — the owning billing account is absent — for every token, so with zapper there is no longer any such thing as "no logo". Every request is a failure, and the stricter assertion now says so. The same 403 is why zapper's image-snapshot case is already red on master. If we want that suite green, the honest fix is to retire the resolver (or `skip` it the way selfid is), not to mute a 403.

### DIFF
--- a/src/resolvers/basename.ts
+++ b/src/resolvers/basename.ts
@@ -4,12 +4,8 @@ import { resize } from '../utils';
 import { fetchHttpImage } from './utils';
 
 export default async function resolve(nameOrAddress: string) {
-  try {
-    const url = await getAvatar(nameOrAddress);
-    if (!url) return false;
+  const url = await getAvatar(nameOrAddress);
+  if (!url) return false;
 
-    return await resize(await fetchHttpImage(url), max, max);
-  } catch {
-    return false;
-  }
+  return await resize(await fetchHttpImage(url), max, max);
 }
--- a/src/resolvers/coingecko.ts
+++ b/src/resolvers/coingecko.ts
@@ -1,7 +1,6 @@
-import { getAddress } from '@ethersproject/address';
 import { max } from '../constants.json';
 import { resize } from '../utils';
-import { fetchHttpImage } from './utils';
+import { fetchHttpImage, toChecksumAddress } from './utils';
 
 const API_KEY = process.env.COINGECKO_API_KEY;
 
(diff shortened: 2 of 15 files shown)


tony8713:

One thing to know before this goes green: the new `zapper › returns false when no avatar is set, without reporting an error` case fails right now, and it is not the rebase. The `zapper-fi-assets` GCS bucket has started answering 403 to everything (`curl` on the raw asset URL returns 403 for both a missing token and the base asset), so the resolver throws instead of 404-ing, and the new wrapper reports it. I confirmed it fails identically on this PR's pre-rebase tip 896deca. The same outage is why master's `zapper › matches the image snapshot` case is red.

If the bucket is not coming back, zapper probably wants a `MUTED_ERRORS` entry the way ens and trustwallet have one — but that is a call for this PR, not something I wanted to slip into a rebase.

wa0x6e:
See https://github.com/snapshot-labs/stamp/issues/511

PR #509, opened 2026-08-10
fix(deps): pin starknet below the ASCII-only .stark domain guard

tony8713:
starknet.js added an ASCII-only regex in front of `getAddressFromStarkName`. StarknetID's bigAlphabet glyphs are valid on chain but do not match it, so a `.stark` name spelled with them throws, the avatar resolver's catch turns that into "no avatar", and the request falls through to a blockie. No fixture used a bigAlphabet name, so the whole thing goes silent with every gate green.

Pinning stamp's own dependency does not fix it. The Starknet provider is constructed inside snapshot.js, so the copy that actually runs is the one snapshot.js resolves, which is a nested one as soon as the two ranges diverge. The pin therefore lives in `resolutions`, which keeps every copy on one version. The lower bound stays at the release snapshot.js needs for its `baseFetch` hook.

Also adds a bigAlphabet name to the avatar resolver's image snapshots, so this cannot go quiet again.


wa0x6e (approved):
utAck

Issue #511, opened 2026-08-13
refactor: provider registries for resolvers and lookupDomains (plan and tracking)


wa0x6e:
Tracking issue for the provider refactoring. This began as the plan for the `src/resolvers` series and now covers both provider tracks, with each step specified in its own sub-issue. The reasoning below is the original plan's, restructured rather than replaced. What has actually been overtaken is called out under "What changed since this was written", so anyone who read the earlier version can see what moved.

## What is being changed, and why

Stamp has several families of providers and each family wires its providers up by hand. `src/resolvers` exports a map where every entry carries a decision made at the call site. `src/lookupDomains` keeps two per-provider lists that have to be kept in step. Adding or changing a provider means editing code in more than one place and re-deriving decisions that are written down in comments rather than carried by the entries.

Two consequences are the point of the work.

**Errors are invisible.** Every image resolver ends in a bare `catch { return false }`, so a resolver dying and an address simply having no avatar are the same event: blockie fallback, nothing reported. `src/addressResolvers` already solved this class of problem (#496), and the image resolvers should follow.

**Policy is scattered.** What is silenced and what pages is an ops decision. It is currently invisible and re-decided per file, and it should live in one place rather than in each resolver.

A registry is what makes both fixable: each provider gets one entry to carry its name, its properties and its error policy, and the index gets one place to own the pipeline contract.

## The end state

Two registry shapes, not one, and the second one is forced rather than chosen.

`src/addressResolvers` and `src/lookupDomains` register one module per provider: the entry *is* the module, and the name and properties are exports of it. `src/resolvers` cannot do that, because a module there is not a provider. `snapshot.ts` produces five registry entries from one factory and `space-sx.ts` produces two, so that registry has to be an array of entries with the function named in each.

What does converge is what an entry carries: a name, its properties and its error policy, written as data on the entry rather than re-derived at the call site from a comment. The planned steps in both tracks below are steps toward that, not separate cleanups that happen to look alike.

Where a family genuinely needs a property the others do not, that stays a deliberate difference rather than a reason to keep the shapes apart. The chains a domain provider serves is one: address resolvers do not fan out across chains at all.

## Tracks

Two tracks. They touch different files and are independent, so either can go first.

**Resolvers**, `src/resolvers`

- #522, build the resolver map from one registry array. Done.
- Moving each individually written resolver's `catch { return false }` into the wrapper: done, for the eight that had one.
[...]

PR #512, opened 2026-08-13
refactor(resolvers): resize in the resolver map instead of each resolver

tony8713:
Every image resolver ends with the same `return await resize(input, max, max)`, so the max by max webp normalization is copied across the resolver files along with its `resize` and `max` imports. That is pipeline work rather than resolver work: a resolver's job is to find the bytes, and normalizing them belongs to the layer that assembles the map.

This adds a `withResize()` wrapper in `src/resolvers/index.ts` and applies it to the avatar entries, dropping the resize call and its imports from those resolvers. Their contract is unchanged: a Buffer or `false`, never a throw.

The cover and logo entries (`user-cover`, `space-cover`, `space-logo`, `space-cover-sx`) are deliberately left unwrapped, and that is the part worth a second look. The shared `resize(input, max, max)` passes no fit option, so sharp falls back to `fit: 'cover'` and center-crops to a square instead of fitting inside one. Banners are wide, so wrapping those entries would bake a square crop into the base cache, and a base cached as a square can never serve a wide banner again. `api.ts` resizes every response to the requested `w`x`h` regardless, so this is about what the cache holds rather than the dimensions we serve. `blockie` and `jazzicon` stay unwrapped too, since `api.ts` feeds their result into `resize()` itself and they sit outside the false-on-failure contract.

A design note on the `catch` in the wrapper: today `resize` runs inside each resolver's own catch, so bytes sharp refuses (an ENS avatar record can point at anything) answer `false`. Keeping a catch preserves that. Without it those failures would throw into the unguarded `Promise.all` in `api.ts` and turn a per-resolver no-data answer into a failed request. It now captures as well, since one wrapper swallowing every resolver's sharp failures hides more than each resolver swallowing its own.

Part of #511, the resize extraction step of that plan.

### DIFF
--- a/src/resolvers/basename.ts
+++ b/src/resolvers/basename.ts
@@ -1,14 +1,12 @@
-import { getAvatar } from '../addressResolvers/basename';
-import { max } from '../constants.json';
-import { resize } from '../utils';
 import { fetchHttpImage } from './utils';
+import { getAvatar } from '../addressResolvers/basename';
 
 export default async function resolve(nameOrAddress: string) {
   try {
     const url = await getAvatar(nameOrAddress);
     if (!url) return false;
 
-    return await resize(await fetchHttpImage(url), max, max);
+    return await fetchHttpImage(url);
   } catch {
     return false;
   }
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -20,12 +20,10 @@ import { resolveAvatar as sxResolveAvatar, resolveCover as sxResolveCover } from
 import starknet from './starknet';
 import trustwallet from './trustwallet';
 
-// Normalizing to a max x max webp is pipeline work, not resolver work: a
-// resolver finds the bytes, the map normalizes them. The resize keeps the catch
-// it had inside each wrapped resolver, so hostile bytes from those entries
-// answer false rather than throwing into the unguarded image route in api.ts.
-// The cover/logo entries below skip this wrapper and remain exposed to that
-// same gap (pre-existing, tracked in #495).
+// The catch keeps what each resolver had around its own resize call: bytes
+// sharp refuses answer false rather than throwing into the image route, which
+// has no guard of its own (#495). The unwrapped entries below hand their bytes
+// to the bare resize() in api.ts and stay exposed to that gap.
 function withResize<T extends unknown[]>(
   resolve: (...args: T) => Promise<Buffer | false>
 ): (...args: T) => Promise<Buffer | false> {
@@ -42,19 +40,12 @@ function withResize<T extends unknown[]>(
   };
 }
 
-// Two groups of entries deliberately opt out of the resize:
-//   - covers and logos ('user-cover', 'space-cover', 'space-logo',
-//     'space-cover-sx') skip the shared resize so the base cache keeps its
-//     upstream aspect ratio; api.ts still resizes every response to the
-//     request's w x h (default 64x64) before serving it. resize() passes
-//     no options, so sharp falls back to fit: 'cover', which would center-crop a
-//     1500x500 banner into a 500x500 square; parseQuery gives the '-cover'
-//     types (user-cover, space-cover, space-cover-sx) their own, larger
-//     ceiling (constants.maxCover) for the same reason. space-logo keeps the
-//     default constants.max ceiling but is still excluded here so its aspect
-//     ratio isn't force-cropped to a square.
-//   - blockie and jazzicon are the fallbacks api.ts feeds straight into resize(),
-//     so they stay out of the false-on-failure contract and resize themselves.
+// Covers and logos opt out so the base cache keeps its upstream aspect ratio:
+// resize() passes no fit option, so sharp center-crops, and a banner cached as
+// a square can never be served wide again (api.ts resizes every response to the
+// requested w x h regardless).
+// blockie and jazzicon opt out because api.ts feeds their result straight into
+// resize(): they resize themselves and must never answer false.
 export default {
   blockie,
   jazzicon,
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -1,3 +1,4 @@
+import { capture } from '@snapshot-labs/snapshot-sentry';
 import { max } from '../constants.json';
 import { resize } from '../utils';
 import basename from './basename';
@@ -21,8 +22,10 @@ import trustwallet from './trustwallet';
 
 // Normalizing to a max x max webp is pipeline work, not resolver work: a
 // resolver finds the bytes, the map normalizes them. The resize keeps the catch
-// it had inside each resolver, so hostile bytes still answer false rather than
-// throwing into the unguarded image route in api.ts.
+// it had inside each wrapped resolver, so hostile bytes from those entries
+// answer false rather than throwing into the unguarded image route in api.ts.
+// The cover/logo entries below skip this wrapper and remain exposed to that
+// same gap (pre-existing, tracked in #495).
 function withResize<T extends unknown[]>(
   resolve: (...args: T) => Promise<Buffer | false>
 ): (...args: T) => Promise<Buffer | false> {
@@ -32,7 +35,8 @@ function withResize<T extends unknown[]>(
 
     try {
       return await resize(input, max, max);
-    } catch {
+    } catch (err) {
+      capture(err);
       return false;
     }
   };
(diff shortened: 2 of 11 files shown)


src/resolvers/index.ts
```diff
@@ -18,23 +20,50 @@ import starknet from './starknet';
...
+    }
+  };
+}
+
+// Two groups of entries deliberately opt out of the resize:
+//   - covers and logos ('user-cover', 'space-cover', 'space-logo',
+//     'space-cover-sx') are served at their upstream dimensions. resize() passes
+//     no options, so sharp falls back to fit: 'cover', which would center-crop a
+//     1500x500 banner into a 500x500 square; parseQuery gives those types their
+//     own, larger ceiling (constants.maxCover) for the same reason.
```
wa0x6e: The `parseQuery` ceiling claim here is false for `space-logo`: `utils.ts:96` gates the larger ceiling on `type.includes('-cover')`, which doesn't match `'space-logo'` — logos are capped at `constants.max` (500), not `maxCover` (1500). The exclusion itself is still right (behavior unchanged vs. the old `snapshot.ts` branch), only the recorded rationale is off, and it's what a future reader will trust when deciding whether this entry can be wrapped.

```suggestion
//     1500x500 banner into a 500x500 square; parseQuery gives the '-cover'
//     types (user-cover, space-cover, space-cover-sx) their own, larger
//     ceiling (constants.maxCover) for the same reason. space-logo keeps the
//     default constants.max ceiling but is still excluded here so its aspect
//     ratio isn't force-cropped to a square.
```

src/resolvers/index.ts
```diff
@@ -18,23 +20,50 @@ import starknet from './starknet';
...
+      return await resize(input, max, max);
+    } catch {
+      return false;
+    }
+  };
+}
+
+// Two groups of entries deliberately opt out of the resize:
+//   - covers and logos ('user-cover', 'space-cover', 'space-logo',
+//     'space-cover-sx') are served at their upstream dimensions. resize() passes
```
wa0x6e: "Served at their upstream dimensions" isn't quite right: `api.ts:152` unconditionally resizes every response to the request's `w`×`h` (default 64×64 via `parseQuery`, with sharp's `fit: 'cover'` default) before sending, cover types included. The opt-out only preserves the *base cache* aspect ratio — which is what makes it matter (a square-cropped base can never serve a wide banner later), but the comment as written sends anyone debugging a cropped `GET /space-cover/<id>` response down the wrong path.

```suggestion
//   - covers and logos ('user-cover', 'space-cover', 'space-logo',
//     'space-cover-sx') skip the shared resize so the base cache keeps its
//     upstream aspect ratio; api.ts still resizes every response to the
//     request's w x h (default 64x64) before serving it. resize() passes
```

src/resolvers/index.ts
```diff
@@ -18,23 +20,50 @@ import starknet from './starknet';
...
+): (...args: T) => Promise<Buffer | false> {
+  return async (...args: T) => {
+    const input = await resolve(...args);
+    if (!input) return false;
+
+    try {
+      return await resize(input, max, max);
+    } catch {
+      return false;
+    }
```
wa0x6e, line 39: Now that the 12 per-resolver catches collapsed into this single choke point, a silent swallow here hides every sharp-level failure at once — e.g. a `sharp`/libvips bump changing the pixel limit or dropping a codec would silently degrade every affected avatar to the blockie fallback with no signal. The sibling map already captures at this layer (`addressResolvers/index.ts:73`), so this is a one-liner now:

```suggestion
    try {
      return await resize(input, max, max);
    } catch (err) {
      capture(err);
      return false;
    }
```

(needs `import { capture } from '@snapshot-labs/snapshot-sentry';` — suggested on line 1. `capture()` is a no-op under `NODE_ENV=test` and degrades to `console.error` without a DSN.)

src/resolvers/index.ts
```diff
@@ -1,3 +1,5 @@
+import { max } from '../constants.json';
```
wa0x6e, line 2: Import for the `capture()` suggestion below (matches the scoped-before-relative convention in `api.ts:1`):

```suggestion
import { capture } from '@snapshot-labs/snapshot-sentry';
import { max } from '../constants.json';
```

src/resolvers/index.ts
```diff
@@ -18,23 +20,50 @@ import starknet from './starknet';
 import trustwallet from './trustwallet';
 import zapper from './zapper';
 
+// Normalizing to a max x max webp is pipeline work, not resolver work: a
+// resolver finds the bytes, the map normalizes them. The resize keeps the catch
+// it had inside each resolver, so hostile bytes still answer false rather than
+// throwing into the unguarded image route in api.ts.
```
wa0x6e: This safety claim only holds for the wrapped entries — the four unwrapped cover/logo entries still hand raw upstream bytes to the bare `resize()` at `api.ts:152` (no try/catch in the Express 4 handler, so a sharp rejection hangs the request per #495). Worth scoping the wording so the comment doesn't read as a map-wide guarantee:

```suggestion
// it had inside each wrapped resolver, so hostile bytes from those entries
// answer false rather than throwing into the unguarded image route in api.ts.
// The cover/logo entries below skip this wrapper and remain exposed to that
// same gap (pre-existing, tracked in #495).
```

src/resolvers/basename.ts
```diff
@@ -1,14 +1,12 @@
-import { getAvatar } from '../addressResolvers/basename';
-import { max } from '../constants.json';
-import { resize } from '../utils';
 import { fetchHttpImage } from './utils';
+import { getAvatar } from '../addressResolvers/basename';
```
wa0x6e, line 2: Nit: the import reorder is unrelated churn — sibling resolvers (`ens.ts`, `lens.ts`, `selfid.ts`) keep `fetchHttpImage` last.

```suggestion
import { getAvatar } from '../addressResolvers/basename';
import { fetchHttpImage } from './utils';
```

src/resolvers/snapshot.ts
```diff
@@ -130,11 +130,8 @@ function createPropertyResolver(entity: Entity, property: Property) {
       if (!value) return false;
 
       const url = getUrl(value);
-      const input = await fetchHttpImage(url);
 
-      if (['cover', 'logo'].includes(property)) return input;
-
-      return await resize(input, max, max);
+      return await fetchHttpImage(url);
```
wa0x6e: Nit: stray blank line left where the removed `const input` block sat.

```suggestion
      return await fetchHttpImage(url);
```

src/resolvers/space-sx.ts
```diff
@@ -41,10 +40,8 @@ function createPropertyResolver(property: 'avatar' | 'cover') {
       );
 
       const url = getUrl(value);
-      const input = await fetchHttpImage(url);
 
-      if (property === 'cover') return input;
-      return await resize(input, max, max);
+      return await fetchHttpImage(url);
```
wa0x6e: Same nit as in `snapshot.ts`: stray blank line left where the removed lines sat.

```suggestion
      return await fetchHttpImage(url);
```

src/resolvers/index.ts
tony8713: Correct, `'space-logo'.includes('-cover')` is false, so logos sit at `constants.max`. Took your wording in 37b0719, then the comment trim in a94e3e8 cut the ceiling clause altogether: what decides whether one of these entries can be wrapped is the aspect ratio, and the `maxCover` detail belongs to `parseQuery`. Say the word if you'd rather have the corrected sentence recorded here.

src/resolvers/index.ts
tony8713: Good catch, `api.ts:152` resizes on the way out either way. Applied in 37b0719; the trim in a94e3e8 rewords it but keeps both halves, that the opt-out buys the base cache's aspect ratio and that every response is resized to the requested `w`x`h` regardless.

That is the one claim here I did measure, since it decides whether these four can ever be wrapped: `sharp(banner).resize(500, 500)` on a 1500x500 returns mean RGB `0,255,1` for a red/green/blue-thirds banner, i.e. only the middle third survives, and re-expanding that base to 1500x500 still gives `0,255,1`. The outer thirds are gone for good, so the comment stays.

src/resolvers/index.ts
tony8713: Agreed, one choke point swallowing every resolver's sharp failures is worse than each resolver swallowing its own. Applied in 37b0719.

src/resolvers/index.ts
tony8713: Applied in 37b0719.

src/resolvers/index.ts
tony8713: Fair, the guarantee only covers the wrapped entries. Applied in 37b0719 and kept through the trim in a94e3e8, shorter but still scoped and still pointing at #495.

src/resolvers/basename.ts
tony8713: Can't take this one: the reorder is the linter's, not mine. With `../constants.json` and `../utils` deleted from this file, `import-x/order` wants `./utils` first, and your version fails `yarn lint`:

```
src/resolvers/basename.ts
  2:1  error  `./utils` import should occur before import of `../addressResolvers/basename`  import-x/order
```

The siblings keep `fetchHttpImage` last because they all still import `../utils`, which the rule ranks equal to `./utils`. `basename.ts` is the only one left whose neighbour is `../addressResolvers/*`, and there the rule has an opinion. Same reason it was fine on master: `../utils` used to sit between the two.

src/resolvers/snapshot.ts
tony8713: Applied in 37b0719.

src/resolvers/space-sx.ts
tony8713: Applied in 37b0719.

wa0x6e (approved):
tAck

PR #514, opened 2026-08-14
chore(resolvers): remove the zapper resolver

tony8713:
Zapper has shut down, and the asset bucket this resolver reads has had its Google Cloud billing disabled. Every object under `storage.googleapis.com/zapper-fi-assets` now answers `403 UserProjectAccountProblem`, for everyone, not only for us. The resolver was unauthenticated and never had a key, so there is nothing to renew and no replacement host. It cannot return an image again.

It catches the 403 and returns false, so the token chain falls through and nothing is user-visible today. What is left is a guaranteed-failing outbound request on every token lookup that reaches it, and an integration test pinned to a bucket that no longer serves anything.

This removes the resolver, its entry in the token chain, its integration test, the committed image snapshot and its fixture input. The token chain keeps trustwallet and coingecko.


wa0x6e (approved):
utAck

Issue #519, opened 2026-08-14
Integration tests that assert against live third-party services are the main source of red master


tony8713:
Three red master CI stretches in two weeks all have the same shape: an integration test whose pass or fail depends on a third-party service staying alive. api.starknet.id stopped accepting connections and took the `.stark` name tests with it, which is part of what #491 fixes by reading the naming contract on chain instead. Zapper shut down and its asset bucket started answering 403 for everyone, so the resolver and its test came out in #514. Now the ENS off-chain case is red because the gateway serving `avsa.eth` returns 503, which is #518.

Each of those assertions was reasonable when it was written, and each one now reports the same fact: someone else's service changed. Red master stops meaning our code broke, and `jest.retryTimes(3)` cannot tell the two apart, so a dead dependency burns four attempts and still lands as a test failure. The real cost is that a genuinely broken master looks exactly like the noise.

The line worth agreeing on: an integration test may hit a third party to prove that our request shape and our parsing are right, but master should not go red because one particular name, asset or gateway went away. Where the subject of the test is our own glue, stub the third-party half. Where a live call really is the point, it belongs somewhere that can go red without blocking merges. Also worth a pass over the rest of the integration suite to find which cases carry the same fuse, rather than meeting them one outage at a time.

Issue #518, opened 2026-08-14
ENS CCIP-Read test asserts against an off-chain gateway that is now down


tony8713:
`test/integration/addressResolvers/ens.test.ts`, case `resolves names that the batch contract misses`, has failed every master `Test` run since 2026-08-03. It is broken rather than flaky: `test/setup-jest.ts` sets `jest.retryTimes(3)`, so each red run is four failed attempts.

The cause is outside our code. The case expects `lookupAddresses` on `0x809FA673fe2ab515FaA168259cB14E2BeDeBF68e` to return `avsa.eth`, and gets `{}`. That name's forward resolver `0xd17347fA0a6eeC89a226c96a9ae354F785e94241` returns `addr() == 0x0` on chain and serves the real record off chain via CCIP-Read, and the gateway it points ethers at, `gateway.namestone.xyz`, answers `503 Service Suspended - This service has been suspended by its owner`. The batch path cannot cover for it either: with `addr()` empty on chain the ReverseRecords forward check fails, so the batch returns `""` and the per-address fallback throws. The resolver contract is alive, the gateway behind it is gone, and while it stays gone that name is unresolvable by any route. It is the name owner's to fix, not ours.

The fix is to keep the case live and point it at a name with more behind its gateway. Proposed fixture: address `0x2211d1D0020DAEA8039E46Cf1367962070d77DA9`, expected name `jesse.base.eth`. It has the same shape as the one it replaces. ReverseRecords returns `""` for that address, so the per-address `provider.lookupAddress` fallback still has to run, and that fallback still has to complete a real CCIP-Read round trip before the case can pass.

The gateway it reaches is `api.coinbase.com`, and that is the whole argument for it. It serves L1 resolution for every name under `base.eth`, not for one name, so it cannot be switched off without every basename going dark on mainnet at the same time. `gateway.namestone.xyz` could be, and was, switched off by one owner with nobody else affected.

`offchainexample.eth` was the obvious first choice and does not work here. It is forward only: the address it resolves to has an on-chain primary name, so the batch hits and the fallback never runs, and no address has a `*.offchainexample.eth` primary name to reverse.

This does not remove the dependency, it moves it somewhere heavier. The case still asserts against someone else's live service over the public internet, so it still carries the risk described in #519, and it can still go red for reasons that are not ours: the name owner re-points their primary name, or Coinbase changes the gateway. If we want that risk out of `Test`, the answer is still a job that can go red without blocking merges. This issue does not add one.

[...]

PR #520, opened 2026-08-14
test(ens): point the CCIP-Read case at a live off-chain gateway

tony8713:
The CCIP-Read fallback case in `test/integration/addressResolvers/ens.test.ts` asserts that `avsa.eth` resolves, and it cannot pass any more. That name's forward resolver returns an empty address on chain and serves the real record off chain, and the gateway it points ethers at, `gateway.namestone.xyz`, answers `503 Service Suspended`. The resolver contract is alive, the gateway behind it is gone, and while it stays gone the name is unresolvable by any route. It is the name owner's to fix, not ours.

This keeps the case and swaps the fixture to `jesse.base.eth`. Same shape as the one it replaces: ReverseRecords returns an empty string for that address, so the batch still misses and the per-address `provider.lookupAddress` fallback still has to complete a real CCIP-Read round trip before the case can pass. Nothing else in the case changes.

The argument for this gateway is what sits behind it. `api.coinbase.com` serves L1 resolution for every name under `base.eth` rather than for one name, so it cannot be switched off without taking the whole namespace with it. That is exactly what one owner was able to do to `gateway.namestone.xyz`.

This does not remove the dependency, it moves it somewhere heavier. The case still asserts against someone else's live service, so the risk in #519 stands: it can go red because the name owner re-points their primary name, or because Coinbase changes the gateway.

Ref #518

### DIFF
--- a/test/integration/addressResolvers/ens.test.ts
+++ b/test/integration/addressResolvers/ens.test.ts
@@ -18,13 +18,13 @@ testAddressResolver({
 });
 
 describe('ENS address resolver: CCIP-Read fallback', () => {
-  // avsa.eth's primary name is set via an off-chain resolver that the batch
+  // jesse.base.eth's primary name is set via an off-chain resolver that the batch
   // getNames contract doesn't follow, so the fallback to provider.lookupAddress
   // is required.
   it('resolves names that the batch contract misses', async () => {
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
utAck

PR #523, opened 2026-08-15
refactor(lookupDomains): register each provider in one place

### LINKED ISSUE #521: refactor(lookupDomains): register each provider in one place
A proposal, not a decision. This is the first PR of a longer cleanup, not the cleanup.

## What is wrong

`src/lookupDomains/index.ts` registers each provider twice. `RESOLVERS` holds the lookup functions, `DEFAULT_CHAIN_IDS` holds one renamed `DEFAULT_CHAIN_ID` import per provider, and the two lists are kept in step by hand. Add a provider to the first and not the second and it compiles, passes review, and is then never called on the default path, because the default `chains` argument is built from the second list. Nothing fails, so nothing tells you.

Nothing catches it either. The default argument is exercised by the integration suite, but the only case there with a positive assertion looks for a `.eth` name, so it shows ENS was reached and nothing more. The unit suite is the one that mocks the providers and could assert which of them ran, and it passes an explicit chain list in every case. A provider missing from the default list leaves the suite passing.

## The change

One array, entries mapped in place:

```ts
const PROVIDERS = [
  { lookup: ens, defaultChainId: ENS_DEFAULT_CHAIN_ID },
  { lookup: shibarium, defaultChainId: SHIBARIUM_DEFAULT_CHAIN_ID },
  { lookup: unstoppableDomains, defaultChainId: UNSTOPPABLE_DOMAINS_DEFAULT_CHAIN_ID }
];

const DEFAULT_CHAIN_IDS = PROVIDERS.map(p => p.defaultChainId);
```

The fan-out loop reads `lookup` off the entry. Nothing else moves. A provider is registered in one place and cannot be half registered.

## Why it cannot change behaviour

The entries keep the order the two lists have today, and that order is observable: the result is a flattened, deduplicated array built provider by provider and chain by chain, so reordering entries would reorder the response. Mapping in place preserves both orders.

`DEFAULT_CHAIN_ID` still comes from each provider module, so the existing unit suite, which mocks exactly that module shape and asserts the fan-out, the best effort result and one report per failing chain, passes with no edits. Passing it unchanged is what shows the behaviour is identical.

One test is added for the case nothing covers: called with no chain argument, every registered provider is invoked.

## Left for later

Sketches of where this could go, not commitments. None of them is proposed here, only the change above is.

- `refactor(lookupDomains): name each provider and put the name in the error context`, so a report can say which provider failed.
- `refactor(lookupDomains): report Shibarium failures through index.ts like the others`. It catches, classifies and reports inside itself and returns partial results, so it never reaches the handler the other two go through.
- `refactor(lookupDomains): declare each provider's chains in the registry`. Needs care, since a provider's real chain support is wider than its `DEFAULT_CHAIN_ID`: ENS serves every chain in `constants.ensSubgraph`.
- `feat(lookupDomains): give each provider an optional mute list`, matching `MUTED_ERRORS` in `src/addressResolvers/index.ts`.

The same lift for `src/resolvers` is planned separately as PR-1 in #511. The two do not overlap and either can go first.

tony8713:
Two commits, and they are worth reading separately. The first is provably inert, the second is the behavioural one.

**1. `refactor(lookupDomains): register each provider in one place`**

`src/lookupDomains/index.ts` registered each provider twice: `RESOLVERS` for the lookup function and `DEFAULT_CHAIN_IDS` for the chain it serves by default. Add a provider to the first and miss the second and it compiles, passes review, and is then never called on the default path, because the default `chains` argument is built from the second list. Nothing fails, so nothing tells you.

That collapses into one `PROVIDERS` array with the chain ids derived from it by mapping in place, so a provider cannot be half registered. Order is preserved because the result is flattened and deduplicated provider by provider and chain by chain. The existing unit suite passes with no edits at that commit, which is what shows the behaviour is identical.

**2. `refactor(lookupDomains): declare each provider's chains in the registry`**

The fan-out asked every provider about every requested chain and let each one decide for itself. On the default path that was nine calls, six of which returned empty from a guard before making any request. The explicit path did the same: `lookupDomains(address, ['1'])` asked all three providers and two returned empty.

Each provider now exports the chain ids it serves, derived from the same constant its own guard already reads, so there is still one declaration per provider. The registry entry carries that list and `index.ts` calls a provider only for the requested chains it declares, in the requested order.

Two things worth pointing at:

**The guards stay in the providers.** They are not dead code: each provider is imported and called directly, and both `lookupDomains/shibarium` and `lookupDomains/unstoppableDomains` have a test asserting they do not call their API on an unsupported chain. The registry is the routing policy, the guard is the provider's own precondition, and both read the same declaration.

**The shibarium API key check stays inside the provider.** Its guard is not purely a chain list, it also requires `D3_API_KEY_*` to be present. A static list says which chains the provider can serve; whether the key is configured is a runtime question the registry cannot answer, and hoisting it would freeze the answer at import time.

Results are unchanged on both paths. Two tests were added for the routing, and one existing test, `captures one event per failing chain`, now asks ENS about the two chains ENS declares rather than about three chains only one of which it ever served. Its assertion is otherwise unchanged.

Closes #521, including the "declare each provider's chains in the registry" step from its follow-on list.

### DIFF
--- a/src/lookupDomains/ens.ts
+++ b/src/lookupDomains/ens.ts
@@ -2,6 +2,7 @@ import constants from '../constants.json';
 import { Address, graphQlCall, Handle } from '../utils';
 
 export const DEFAULT_CHAIN_ID = '1';
+export const CHAIN_IDS = Object.keys(constants.ensSubgraph);
 
 type Domain = {
   name: string;
--- a/src/lookupDomains/index.ts
+++ b/src/lookupDomains/index.ts
@@ -2,16 +2,24 @@ import { isAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import { isSilencedError } from '../addressResolvers/utils';
 import { Address, Handle } from '../utils';
-import ens, { DEFAULT_CHAIN_ID as ENS_DEFAULT_CHAIN_ID } from './ens';
-import shibarium, { DEFAULT_CHAIN_ID as SHIBARIUM_DEFAULT_CHAIN_ID } from './shibarium';
+import ens, { CHAIN_IDS as ENS_CHAIN_IDS, DEFAULT_CHAIN_ID as ENS_DEFAULT_CHAIN_ID } from './ens';
+import shibarium, {
+  CHAIN_IDS as SHIBARIUM_CHAIN_IDS,
+  DEFAULT_CHAIN_ID as SHIBARIUM_DEFAULT_CHAIN_ID
+} from './shibarium';
 import unstoppableDomains, {
+  CHAIN_IDS as UNSTOPPABLE_DOMAINS_CHAIN_IDS,
   DEFAULT_CHAIN_ID as UNSTOPPABLE_DOMAINS_DEFAULT_CHAIN_ID
 } from './unstoppableDomains';
 
 const PROVIDERS = [
-  { lookup: ens, defaultChainId: ENS_DEFAULT_CHAIN_ID },
-  { lookup: shibarium, defaultChainId: SHIBARIUM_DEFAULT_CHAIN_ID },
-  { lookup: unstoppableDomains, defaultChainId: UNSTOPPABLE_DOMAINS_DEFAULT_CHAIN_ID }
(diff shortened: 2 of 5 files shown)


wa0x6e (approved):
utAck

PR #524, opened 2026-08-15
refactor(resolvers): build the map from one registry array

### LINKED ISSUE #522: refactor(resolvers): build the resolver map from one registry array
A proposal, not a decision. #511 already describes this series, and this is the concrete first step of it rather than a competing plan.

## What is wrong

`src/resolvers/index.ts` exports a hand-written map, and every entry carries a decision made at the call site: wrapped in `withResize()`, or passed through bare. Which entries are bare, and why, lives in a comment above the literal rather than in the entries themselves, and the two reasons for being bare are not the same reason. Covers and logos opt out so the base cache keeps its upstream aspect ratio. Blockie and jazzicon opt out because `api.ts` hands their result straight to `resize()` and they must never answer `false`. Adding a resolver means adding an import, adding a line to the literal, and re-deriving that decision from prose.

## Why this part first

#511 sequences the registry lift ahead of the resize move. #512 landed them the other way round, and that changes what is safe to do next.

The wrapper #512 introduced does two jobs at once, resizing and catching, and it is applied to only some entries. So moving each resolver's own `catch { return false }` into the wrapper, which is the other half of #511's PR-1, is no longer safe as one step. `user-cover`, `space-cover`, `space-logo` and `space-cover-sx` are bare in the map, and their catch lives in the shared factories in `src/resolvers/snapshot.ts` and `src/resolvers/space-sx.ts`. Deleting it would let them throw into the image route, which has no guard of its own (#495). Because those catches sit inside a factory, they also cannot be removed for the wrapped entries without being removed for the bare ones at the same time.

Separating the wrapper's two jobs needs the entries to carry their properties as data. So the registry goes first, on its own, and the catch move follows it.

## The change

One array in `src/resolvers/index.ts`, with the exported map built from it:

```ts
const RESOLVERS = [
  { name: 'blockie', fn: blockie, resize: false },
  { name: 'ens', fn: ens, resize: true },
  { name: 'user-cover', fn: sResolveUserCover, resize: false },
  // ...
] as const satisfies readonly ResolverEntry[];
```

Nothing outside `index.ts` moves. No resolver file is touched and no catch changes hands.

## Why it cannot change behaviour

The entries keep the current keys, functions and wrapping, so the exported map stays the object it is today.

The real trap is the type. `keyof typeof resolvers` is a literal union and the export has to stay one. Building the map with a bare `Object.fromEntries` widens it to `string`, and nothing fails to compile when it does. `test/integration/resolvers/snapshot.test.ts` passes `resolver: 'user-cover'`, `'space'` and `'space-cover'` into `helper.ts`, where that field is typed against the union, so the widening would quietly stop checking them. Keeping the union is part of the change rather than a nicety, and it is worth a type-level assertion in the suite so it cannot be lost later.

The resolver integration suite gives most of the runtime proof for free, since it calls entries through the exported map by name and asserts each one's committed image bytes, so a key that moved or a wrapping that changed surfaces there. It does not cover everything: `space-logo` and `space-cover-sx` have no case of their own and `selfid` is skipped, so those want reading rather than running.

## Left for later

None of these is proposed here.

- Splitting `resize` into the two properties it currently conflates. One flag reproduces today exactly, and two will be needed once the wrapper also catches.
- Moving each resolver's `catch { return false }` into the wrapper, which needs every entry covered by one first.
- The resolver name in the error context, the per-resolver mute lists, and the capture itself, which is the rest of #511.

This change touches no resolver file, so it does not move #483's conflict surface. That is the catch move, later in the series.

tony8713:
One commit, and it is inert. There is no behavioural half to separate out: the exported map ends up the same object it is today, so splitting it the way #523 was split would give a second commit with nothing in it.

`src/resolvers/index.ts` exported a hand-written literal where every entry carried a decision made at the call site, wrapped in `withResize()` or passed through bare. Which entries were bare, and why, lived in a comment above the literal rather than in the entry. The two reasons are not even the same reason: covers and logos opt out to keep their upstream aspect ratio, blockie and jazzicon opt out because `api.ts` hands their result straight to `resize()` and they must never answer `false`.

That collapses into one `RESOLVERS` array carrying the name, the function and whether the entry resizes, with the map derived from it. It follows the shape #523 settled on for `lookupDomains`: object literal entries with camelCase fields, module constants mapped in.

Two things worth pointing at:

**The literal key union is the whole risk.** `Object.fromEntries` types its result as `Record<string, ...>`, and `keyof` that is `string | number`, not the seventeen names. Nothing fails to compile at the point where it widens. `api.ts` indexes the map by name and `test/integration/resolvers/helper.ts` types its `resolver` field as `keyof typeof resolvers`, so the widening quietly stops checking every resolver name that `snapshot.test.ts` passes. The entries are therefore `as const` and the map is cast back to `Record<ResolverName, ResolverFn>`. `test/unit/resolvers.test.ts` carries an annotation that stops compiling if the union widens, which is the one failure that would otherwise be silent.

**The catches stay where they are.** Nothing outside `index.ts` is touched. Moving each resolver's `catch { return false }` into the wrapper is the next step, not this one, and it cannot be done in the same pass: `snapshot.ts` supplies five entries and `space-sx.ts` supplies two from shared factories, so removing the catch for the wrapped entries removes it for the bare ones in the same edit, and the bare ones would then throw into the image route, which has no guard of its own (#495).

`withResize` loses its generic and takes `ResolverFn` instead. Inferring a single type parameter across seventeen different resolver signatures does not work, and the map's values are all `ResolverFn` after the cast anyway, so nothing is lost.

Splitting the `resize` flag into the two properties it conflates is also left for later. One flag reproduces today exactly.

Closes #522.

### DIFF
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -20,14 +20,20 @@ import { resolveAvatar as sxResolveAvatar, resolveCover as sxResolveCover } from
 import starknet from './starknet';
 import trustwallet from './trustwallet';
 
+type ResolverFn = (...args: any[]) => Promise<Buffer | false>;
+
+type Resolver = {
+  name: string;
+  fn: ResolverFn;
+  resize: boolean;
+};
+
 // The catch keeps what each resolver had around its own resize call: bytes
 // sharp refuses answer false rather than throwing into the image route, which
-// has no guard of its own (#495). The unwrapped entries below hand their bytes
+// has no guard of its own (#495). Entries with resize: false hand their bytes
 // to the bare resize() in api.ts and stay exposed to that gap.
-function withResize<T extends unknown[]>(
-  resolve: (...args: T) => Promise<Buffer | false>
-): (...args: T) => Promise<Buffer | false> {
-  return async (...args: T) => {
+function withResize(resolve: ResolverFn): ResolverFn {
+  return async (...args) => {
     const input = await resolve(...args);
     if (!input) return false;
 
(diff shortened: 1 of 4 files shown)


src/resolvers/index.ts
```diff
@@ -40,28 +46,38 @@ function withResize<T extends unknown[]>(
...
+  { name: 'space', resolve: sResolveSpaceAvatar, resize: true },
+  { name: 'space-cover', resolve: sResolveSpaceCover, resize: false },
+  { name: 'space-logo', resolve: sResolveSpaceLogo, resize: false },
+  { name: 'space-sx', resolve: sxResolveAvatar, resize: true },
+  { name: 'space-cover-sx', resolve: sxResolveCover, resize: false },
+  { name: 'selfid', resolve: selfid, resize: true },
+  { name: 'lens', resolve: lens, resize: true },
+  { name: 'starknet', resolve: starknet, resize: true },
+  { name: 'farcaster', resolve: farcaster, resize: true }
+] as const satisfies readonly Resolver[];
```
wa0x6e, line 74: The `satisfies` here requires TypeScript ≥ 4.9, but `package.json` still declares `"typescript": "^4.7.3"` — a range that textually permits 4.7/4.8, where this file fails to parse.

No realistic install path actually breaks: `yarn.lock` pins 4.9.5, and 4.9.5 is the last 4.x ever published, so every resolution of `^4.7.3` lands on it (the next release, 5.0.0, is outside the caret). Purely housekeeping, but bumping the declared floor would make the manifest reflect the code's actual minimum:

```
"typescript": "^4.9.5",
```

(Can't attach as a suggestion — `package.json` isn't part of this diff.)

tony8713: Right on every point, and it is this PR that creates the gap: `satisfies` appears nowhere else in `src/` or `test/` on master, so the registry is the first use and the manifest floor is mine to move. Bumped to `^4.9.5`.

One thing your suggestion could not see from the diff view. The lockfile key has to move with it. `yarn.lock` keys the entry as `typescript@^4.7.3`, so changing only the manifest fails the install:

```
$ yarn install --frozen-lockfile
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```

That is exactly what `test.yml` runs, so the one-line version would have reddened CI. It is one line either side and the resolved version, tarball and integrity are all unchanged, which is the proof it changes no installed bytes:

```diff
-typescript@^4.7.3:
+typescript@^4.9.5:
   version "4.9.5"
```

Worth noting `^4.9.5` is now effectively an exact pin, since 4.9.5 is the last stable 4.x and the next release is 5.0.0, outside the caret. That is the honest floor rather than a tighter one.

Kept as its own commit so the registry change stays readable on its own.

wa0x6e (approved):
utAck

PR #526, opened 2026-08-15
refactor(lookupDomains): name each provider and put the name in the error context

tony8713:
The capture in the `lookupDomains` fan-out carried the address and the chain id but not which provider failed. A report could say a lookup broke on chain 1 without saying whether that was ENS or something else, and with three providers on the default path that is the first thing you want to know.

Each provider now exports a `NAME` and the registry entry carries it as `name`, which the fan-out puts into the captured context. The names come from the provider modules rather than from literals in `index.ts`, so they cannot drift from the module, and they reuse the spelling `src/addressResolvers` already uses for the same three upstreams.

The function field is `fn`, following the rule that an entry with one function calls it `fn` and an entry with several suffixes each one. That is why it is not `lookup`.

Two things worth pointing at.

**The name goes inside `input`, not in a tag.** `snapshot-sentry` only wraps the context object into `contexts` when none of its top level keys are Sentry's own (`tags`, `extra`, `contexts`, `user`, `level`, `fingerprint`). A top level `tags: { provider }` would stop that wrap and silently drop the address and chain id the report carries today. That is invisible at the call site, so it is the one comment in the diff.

**Shibarium will still be missing from these reports, and now visibly so.** It catches, classifies and reports inside itself and returns partial results, so its failures never reach this capture and never get a name. Naming the others does not fix that, it makes the asymmetry legible in Sentry: two providers that report with a name and one that reports without one from somewhere else. That is the argument for doing the Shibarium reporting step next rather than later.

Scope is the name and its use in the capture. The chain routing, the mute lists and the Shibarium reporting are untouched and keep their places in #511.

On the tests: the mocked provider modules gain `NAME`, and the two assertions that pin the captured payload exactly now include `provider`. That is the change itself rather than a fix to accommodate it. Worth knowing that without the mock change the suite passed anyway, because the mocks had no `NAME`, so `provider` was `undefined` and `toHaveBeenCalledWith` uses `toEqual` semantics that ignore undefined keys. The mocks have to mirror the module shape or the assertion proves nothing.

Part of #511.

### DIFF
--- a/src/lookupDomains/index.ts
+++ b/src/lookupDomains/index.ts
@@ -2,21 +2,45 @@ import { isAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import { isSilencedError } from '../addressResolvers/utils';
 import { Address, Handle } from '../utils';
-import ens, { CHAIN_IDS as ENS_CHAIN_IDS, DEFAULT_CHAIN_ID as ENS_DEFAULT_CHAIN_ID } from './ens';
+import ens, {
+  CHAIN_IDS as ENS_CHAIN_IDS,
+  DEFAULT_CHAIN_ID as ENS_DEFAULT_CHAIN_ID,
+  NAME as ENS_NAME
+} from './ens';
 import shibarium, {
   CHAIN_IDS as SHIBARIUM_CHAIN_IDS,
-  DEFAULT_CHAIN_ID as SHIBARIUM_DEFAULT_CHAIN_ID
+  DEFAULT_CHAIN_ID as SHIBARIUM_DEFAULT_CHAIN_ID,
+  NAME as SHIBARIUM_NAME
 } from './shibarium';
 import unstoppableDomains, {
   CHAIN_IDS as UNSTOPPABLE_DOMAINS_CHAIN_IDS,
-  DEFAULT_CHAIN_ID as UNSTOPPABLE_DOMAINS_DEFAULT_CHAIN_ID
+  DEFAULT_CHAIN_ID as UNSTOPPABLE_DOMAINS_DEFAULT_CHAIN_ID,
+  NAME as UNSTOPPABLE_DOMAINS_NAME
 } from './unstoppableDomains';
 
-const PROVIDERS = [
-  { lookup: ens, defaultChainId: ENS_DEFAULT_CHAIN_ID, chainIds: ENS_CHAIN_IDS },
-  { lookup: shibarium, defaultChainId: SHIBARIUM_DEFAULT_CHAIN_ID, chainIds: SHIBARIUM_CHAIN_IDS },
+type Provider = {
+  name: string;
+  fn: (address: Address, chainId: string) => Promise<Handle[]>;
+  defaultChainId: string;
+  chainIds: string[];
+};
+
+const PROVIDERS: Provider[] = [
   {
-    lookup: unstoppableDomains,
+    name: ENS_NAME,
+    fn: ens,
+    defaultChainId: ENS_DEFAULT_CHAIN_ID,
+    chainIds: ENS_CHAIN_IDS
+  },
+  {
+    name: SHIBARIUM_NAME,
+    fn: shibarium,
+    defaultChainId: SHIBARIUM_DEFAULT_CHAIN_ID,
+    chainIds: SHIBARIUM_CHAIN_IDS
+  },
+  {
+    name: UNSTOPPABLE_DOMAINS_NAME,
+    fn: unstoppableDomains,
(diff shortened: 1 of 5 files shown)


src/lookupDomains/index.ts
```diff
@@ -34,13 +51,16 @@ export default async function lookupDomains(
...
         promises.push(
-          lookup(address, chainId).catch(err => {
-            if (!isSilencedError(err)) capture(err, { input: { address, chainId } });
+          fn(address, chainId).catch(err => {
+            // Adding tags or extra beside input would drop the rest of this context.
+            if (!isSilencedError(err)) {
+              capture(err, { input: { address, chainId, provider: name } });
+            }
             return [];
           })
```
wa0x6e: `provider` lands in a Sentry *context* here, which is readable on an individual event but not searchable or groupable — contexts aren't indexed; tags are. So an on-call engineer still can't filter the issue stream by `provider:Ens` during a spike; they'd open events one by one.

The comment's either/or is narrower than the library allows: `normalizedCaptureContext` only skips the `contexts` wrap when you *rely on the auto-wrap* while adding `tags` (`{ tags, input }` would indeed drop `input`). Supplying `contexts` explicitly alongside `tags` passes through unchanged — both survive:

```suggestion
            // `provider` is a tag so it's searchable/filterable in Sentry;
            // address/chainId stay in `contexts.input` for per-event detail.
            if (!isSilencedError(err)) {
              capture(err, {
                tags: { provider: name },
                contexts: { input: { address, chainId } }
              });
            }
```

This also needs the 5 `toHaveBeenCalledWith` assertions in `test/unit/lookupDomains.test.ts` (lines 108-110, 131-133, 134-136, 137-139, 148-150) moved to the `{ tags: { provider }, contexts: { input: { address, chainId } } }` shape — `capture` is fully mocked there, so they assert the raw call argument.

(Caveat: tags-vs-contexts searchability is documented Sentry platform behavior but wasn't verified against this org's dashboard.)

src/lookupDomains/index.ts
```diff
@@ -2,21 +2,38 @@ import { isAddress } from '@ethersproject/address';
...
+  NAME as SHIBARIUM_NAME
 } from './shibarium';
 import unstoppableDomains, {
   CHAIN_IDS as UNSTOPPABLE_DOMAINS_CHAIN_IDS,
-  DEFAULT_CHAIN_ID as UNSTOPPABLE_DOMAINS_DEFAULT_CHAIN_ID
+  DEFAULT_CHAIN_ID as UNSTOPPABLE_DOMAINS_DEFAULT_CHAIN_ID,
+  NAME as UNSTOPPABLE_DOMAINS_NAME
 } from './unstoppableDomains';
 
 const PROVIDERS = [
```
wa0x6e: `PROVIDERS` has no element type, so the registry doesn't enforce `name`: if a future provider is added without one, array-literal widening makes `name: string | undefined` across all entries, and `capture`'s `any`-typed context means `provider: undefined` compiles clean under `strict: true` and ships to Sentry — the anonymity this PR removes, back silently. The sibling registry already guards this (`type Resolver = { NAME: string; ... }` in `src/addressResolvers/index.ts:25-30`):

```suggestion
type Provider = {
  name: string;
  lookup: (address: Address, chainId: string) => Promise<Handle[]>;
  defaultChainId: string;
  chainIds: string[];
};

const PROVIDERS: Provider[] = [
```

src/lookupDomains/index.ts
tony8713: You're right, applied. I had only checked `{ input, tags }`, which does drop it; supplying `contexts` explicitly does not. Ran both against the installed lib:

```
{ input, tags }    -> {"input":{...},"tags":{...}}               // input left at top level, Sentry ignores it
{ tags, contexts } -> {"tags":{...},"contexts":{"input":{...}}}  // both survive
```

`contexts.input` comes out identical to what the auto-wrap produces today, so address and chainId are unchanged and provider is now filterable. Took your comment and moved all 5 assertions.

src/lookupDomains/index.ts
tony8713: Right, and it is worse than not enforcing: it compiles clean. Dropped `name` from the unstoppableDomains entry and ran tsc both ways:

```
without the annotation:      exit 0, no diagnostic
with PROVIDERS: Provider[]:  error TS2741: Property 'name' is missing in type
                             '{ fn: ...; defaultChainId: string; chainIds: string[]; }'
                             but required in type 'Provider'.
```

The destructured `name` widens to `string | undefined` and `capture`'s context is `any`, so it would have gone out as `provider: undefined` exactly as you describe. Added the type, with `fn` rather than `lookup` since that is the field name on this branch.

wa0x6e (approved):
utAck

PR #527, opened 2026-08-15
fix(utils): render SVGs that use HTML named entities

### LINKED ISSUE #525: SVG avatars with HTML named entities never render, now escalating in Sentry (STAMP-70)
Sentry [STAMP-70](https://snapshot-labs.sentry.io/issues/STAMP-70):

```
Error: Input buffer has corrupt header: glib: XML parse error:
Error domain 1 code 26 on line 1 column 20620 of data: Entity 'nbsp' not defined
```

First seen 2026-08-14 17:24 UTC, escalating. 11 events at 10% sample rate, so ~110 real requests.

## Root cause

The ENS avatar of `0x6b07d4ace5bc9a690fda9ddeb5a0f131fda0b257` is an on-chain-art SVG whose `<text>` runs contain `&nbsp;`:

```
<tspan dy="16" x="203">&nbsp;&nbsp;&nbsp;…
```

XML defines five named entities (`&amp; &lt; &gt; &quot; &apos;`); everything else (`&nbsp;`, `&copy;`, `&mdash;`, …) needs an HTML DTD. Browsers render these SVGs fine, but libvips' XML parser refuses them, so `resize()` throws. Reproduced against sharp 0.34.5 with the downloaded avatar bytes — same error, same column.

## Not a regression, new visibility

These SVGs have **never** rendered through stamp — the owner always got the blockie fallback. Pre-#512 the ens resolver's own `catch { return false }` swallowed the throw silently. #512 moved resize into `withResize()`, whose catch does `capture(err)` — same failure, now reported. sharp hasn't changed since the initial commit, and the fallback behavior is untouched.

The avatar also never reaches the base cache (the resolver answers `false`), so every request for that address re-fetches and re-fails. That's the escalation.

## Proposed fix

When resize throws the entity error, rewrite named HTML entities to their numeric equivalents (`&nbsp;` → `&#160;`) and retry once:

```ts
try {
  return await resize(input, max, max);
} catch (e) {
  if (/Entity '\w+' not defined/.test(e.message)) {
    try {
      return await resize(decodeHtmlEntities(input), max, max);
    } catch {} // still garbage → fall through
  }
  capture(e);
  return false;
}
```

- These avatars render for the first time instead of falling to blockie, and the successful result populates the base cache, which also ends the Sentry noise.
- Capture the **original** error only when the retry also fails: fixable-entity SVGs stop reporting, true garbage still does.
- Inputs that are not actually SVG can't slip through: the retry still terminates in sharp, which rejects anything it can't rasterize, and nothing is ever served that sharp didn't decode.
- Needs a small named→numeric entity map (or an already-installed decoder), not just `nbsp` — any HTML-entity SVG hits the same wall.

Fixture for the test: the 21KB avatar SVG above.

tony8713:
Closes #525.

`resize()` throws on any SVG whose text uses a named entity beyond the five XML defines. libvips' XML parser loads no HTML DTD, so `&nbsp;`, `&copy;` and the rest are undefined entities and the whole document is rejected. Browsers render these files, so an address whose ENS avatar is one has only ever been served the blockie fallback, and since the resolver then answers `false`, nothing reaches the base cache and every request repeats the work.

On that error, rewrite the named entities to their numeric form and retry once. Decoding goes through `entities`, so the names and code points are the ones browsers accept rather than a list maintained here. Left alone: the five names XML defines, anything inside a comment or CDATA section, and any name the document declares for itself, since none of those is what the parser rejected. A name that cannot be resolved rethrows the original error rather than the retry's, so what gets reported still describes the bytes the caller sent, with the retry's own failure attached as `cause`.

The buffer is round-tripped through `latin1` rather than `utf8`, which is lossless for any byte and so leaves a document in a non-UTF-8 encoding intact for libxml2 to decode by its prolog.

The retry lives in `resize()` rather than in the `withResize()` wrapper in `src/resolvers/index.ts`. `api.ts` resizes the cover and logo entries itself and those entries are unwrapped, so an entity SVG on a cover throws at a call site with no guard around it. That is the worse path, since covers come from a user-set URL, and putting the retry in `resize()` covers both.

This narrows #495 rather than fixing it. The image handler is still unguarded and any other sharp failure there is still an unhandled rejection.

The fixture is the avatar from the report, unmodified. It is the first non-`.ts` file in `test/fixtures/`.


tony8713:
Both right, thanks, and both fixed.

On the CDATA point, `numericizeHtmlEntities` now consumes comments and CDATA sections whole so a name inside one stays the literal text it is, and there is a test that fails if that skip is removed. I stopped short of tokenizing XML, since the same regex can skip those two constructs in one alternation.

On coverage, the table is now the whole HTML 4.01 set, 248 names once the ones XML defines are dropped, generated from the W3C DTDs and checked against them. I kept it at HTML 4.01 rather than HTML5: a name still outside the table rethrows the original error and falls back exactly as it did before this PR, so the remaining gap costs nothing.

src/utils.ts
```diff
@@ -45,10 +45,106 @@ export function sha256(str) {
...
+    return await toWebp(input, w, h, options);
+  } catch (err) {
+    if (!isUndefinedEntityError(err)) throw err;
+
+    const svg = input.toString('utf8');
+    const rewritten = numericizeHtmlEntities(svg);
+    if (rewritten === svg) throw err;
+
+    try {
+      return await toWebp(Buffer.from(rewritten, 'utf8'), w, h, options);
```
wa0x6e: The `utf8` round trip is lossy for any SVG that isn't UTF-8. An `<?xml encoding="ISO-8859-1"?>` document with a raw high byte and `&nbsp;` gets each invalid byte turned into U+FFFD → `EF BF BD` here, and the retry then *parses successfully*, so a mojibake avatar is rendered and cached where the old code failed loudly (reproduced on sharp 0.34.5: `café` → `cafï¿½`, differing WebP output). libxml2 does honour the prolog's declared encoding, so feeding it the original bytes is what we want.

`latin1` is a lossless byte↔char map, and the regex only ever touches ASCII (`&name;`, `<!--`, `<![CDATA[`; UTF-8 continuation bytes are all ≥ 0x80), so UTF-8 documents come out byte-identical — checked against the fixture, which has 33 non-ASCII bytes.

```suggestion
    const svg = input.toString('latin1');
    const rewritten = numericizeHtmlEntities(svg);
    if (rewritten === svg) throw err;

    try {
      return await toWebp(Buffer.from(rewritten, 'latin1'), w, h, options);
```

Worth pinning with a small test: a `Buffer.from('…<text>caf\xE9&nbsp;</text>…', 'latin1')` input with an ISO-8859-1 prolog should render byte-equal to the same bytes with `&#160;`.

src/utils.ts
```diff
@@ -45,10 +45,106 @@ export function sha256(str) {
...
+  return svg.replace(
+    /<!--[\s\S]*?-->|<!\[CDATA\[[\s\S]*?\]\]>|&([A-Za-z][A-Za-z0-9]{0,31});/g,
+    (match, name?: string) => {
+      if (name === undefined) return match;
+
+      const codePoint = HTML_ENTITY_CODE_POINTS.get(name);
+      return codePoint === undefined ? match : `&#${codePoint};`;
+    }
+  );
+}
```
wa0x6e: This also rewrites names the document defines itself. librsvg does honour an internal DTD subset, so `<!DOCTYPE svg [<!ENTITY copy "CUSTOM">]>` + `&copy;` renders `CUSTOM` on its own — but once the same file also has an `&nbsp;` somewhere, the retry turns `&copy;` into `&#169;` and the author's definition is silently replaced by ©. Reproduced end to end (pixel comparison against a literal control). Narrow trigger — needs a name colliding with the table *and* a separately undefined entity — but it's silent, and covers/avatars are user-authored.

Skipping names the document declares closes it and leaves everything else as is:

```suggestion
export function numericizeHtmlEntities(svg: string): string {
  const declared = new Set([...svg.matchAll(/<!ENTITY\s+([A-Za-z][A-Za-z0-9]*)\s/g)].map((m) => m[1]));
  return svg.replace(
    /<!--[\s\S]*?-->|<!\[CDATA\[[\s\S]*?\]\]>|&([A-Za-z][A-Za-z0-9]{0,31});/g,
    (match, name?: string) => {
      if (name === undefined || declared.has(name)) return match;

      const codePoint = HTML_ENTITY_CODE_POINTS.get(name);
      return codePoint === undefined ? match : `&#${codePoint};`;
    }
  );
}
```

Pin with a string-level test next to `leaves the names XML defines alone`:

```ts
it('leaves entities the document declares itself alone', () => {
  const doc = '<!DOCTYPE svg [<!ENTITY copy "CUSTOM">]><svg>&copy;&nbsp;</svg>';
  expect(numericizeHtmlEntities(doc)).toBe('<!DOCTYPE svg [<!ENTITY copy "CUSTOM">]><svg>&copy;&#160;</svg>');
});
```

src/utils.ts
```diff
@@ -45,10 +45,106 @@ export function sha256(str) {
...
+    Scaron:352 scaron:353 Yuml:376 circ:710 tilde:732 ensp:8194 emsp:8195
+    thinsp:8201 zwnj:8204 zwj:8205 lrm:8206 rlm:8207 ndash:8211 mdash:8212
+    lsquo:8216 rsquo:8217 sbquo:8218 ldquo:8220 rdquo:8221 bdquo:8222
+    dagger:8224 Dagger:8225 permil:8240 lsaquo:8249 rsaquo:8250 euro:8364
+  `
+    .split(/\s+/)
+    .filter(Boolean)
+    .map((pair): [string, number] => {
+      const [name, codePoint] = pair.split(':');
+      return [name, Number(codePoint)];
```
wa0x6e: The table is clean today (checked: 248 entries, no duplicates, every value an integer), but the parse has no guard, and a future hand-edit typo (`copy:l69`, or a dropped `:`) puts `NaN` in the map. `NaN !== undefined`, so the lookup below doesn't skip it, the rewrite emits `&#NaN;`, sharp rejects the retry, and the outer catch rethrows the original entity error — that name silently falls back to blockie again with nothing failing, since the tests sample five of the names. Cheapest way to make that loud is at import:

```suggestion
      const [name, codePoint] = pair.split(':');
      if (name === undefined || !/^[0-9]+$/.test(codePoint ?? '')) {
        throw new Error(`HTML_ENTITY_CODE_POINTS: malformed entry "${pair}"`);
      }
      return [name, Number(codePoint)];
```

(Alternatively, since the test comment says the table is generated from the W3C DTDs: committing it as JSON the way `chains.json` is removes the parse step entirely, and the generator alongside it makes the "regeneration could pull `amp`/`lt`/`gt`/`quot` back in" worry checkable.)

src/utils.ts
```diff
@@ -45,10 +45,106 @@ export function sha256(str) {
...
+    const rewritten = numericizeHtmlEntities(svg);
+    if (rewritten === svg) throw err;
+
+    try {
+      return await toWebp(Buffer.from(rewritten, 'utf8'), w, h, options);
+    } catch {
+      // The retry parses a document we synthesised, so its error names an
+      // entity the caller never sent. Report the one describing their input.
+      throw err;
+    }
```
wa0x6e: Rethrowing the original is right, but this drops the retry's error on the floor. When the rewrite makes the document parse and the second pass then fails for another reason, the caller — and `withResize()`'s `capture()` — only ever sees `Entity 'nbsp' not defined`, which the retry has already shown is not the blocker. Reproducible with `&nbsp;` plus `width="100000" height="100000"`: retry parses, then dies on `Input image exceeds pixel limit`, and Sentry reports the entity.

Attaching it as `cause` keeps the reported message (and grouping) unchanged; Sentry's default `LinkedErrors` integration walks `.cause` and shows it as a chained exception. The `rethrows the original error when the retry also fails` test still passes since it asserts on `.message`.

```suggestion
    } catch (retryErr) {
      // The retry parses a document we synthesised, so its error names an
      // entity the caller never sent. Report the one describing their input,
      // with the retry's failure attached as `cause` so it isn't lost.
      (err as Error).cause = retryErr;
      throw err;
    }
```

test/unit/resize.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      expect(metadata.format).toBe('webp');
+      expect(metadata.width).toBe(MAX);
+      expect(metadata.height).toBe(MAX);
+    });
+
+    // The table is one generated blob, so these sample it rather than prove it:
+    // copy and yuml from the start and end of the Latin-1 group, alpha from the
+    // symbol and Greek group, mdash from the special group.
+    it('rewrites entities from every HTML 4.01 group', async () => {
+      const output = await resize(
```
wa0x6e: These byte-equality checks only tell a wrong code point apart when a text font is installed. Without one, `&#160;`, `&#169;`, `&#255;` and an empty `<text>` all render to the same bytes (checked with the repo's sharp; the CI runner image ships only `fonts-noto-color-emoji`), so a transposed digit in the table passes this test and the CDATA one — they prove "parses after the rewrite", not "picked the right code point". Cheapest way to make the sample bite regardless of fonts is to pin it at the string level too:

```suggestion
    it('rewrites entities from every HTML 4.01 group', async () => {
      // Pinned at the string level: without a font (none on the CI runner
      // image) a wrong code point renders identically to the right one, so
      // the byte comparison below can't catch a transposed table entry.
      expect(numericizeHtmlEntities('&copy; &yuml; &alpha; &mdash;')).toBe(
        '&#169; &#255; &#945; &#8212;'
      );

      const output = await resize(
```

src/utils.ts
```diff
@@ -45,10 +45,106 @@ export function sha256(str) {
...
+      return codePoint === undefined ? match : `&#${codePoint};`;
+    }
+  );
+}
+
+function toWebp(input, w, h, options?): Promise<Buffer> {
   return sharp(input).resize(w, h, options).webp().toBuffer();
 }
 
+export async function resize(input, w, h, options?): Promise<Buffer> {
```
wa0x6e: Pre-existing signature, but `input.toString(...)` below is the first place `resize()` actually needs a `Buffer` — sharp also takes a path string, and for one of those the retry would quietly no-op (the regex finds nothing in the path) and rethrow the entity error. Every current caller passes a Buffer, so stating it costs nothing (`tsc --noEmit` is clean either way):

```suggestion
export async function resize(input: Buffer, w, h, options?): Promise<Buffer> {
```

src/utils.ts
```diff
@@ -45,10 +45,106 @@ export function sha256(str) {
   return createHash('sha256').update(str).digest('hex');
 }
```
wa0x6e, line 48: This is the HTML 4.01 set, and browsers accept HTML5's superset (~2,200 names, some expanding to two code points), so `&colon;`/`&excl;` still end at `Entity 'colon' not defined` → blockie + Sentry, and `&lang;`/`&rang;` here carry the 4.01 code points (U+2329/232A) where browsers draw U+27E8/27E9. Rather than maintain the table by hand, [`entities`](https://www.npmjs.com/package/entities) (fb55, what htmlparser2/cheerio use; zero deps, ~250 KB unpacked) already ships the WHATWG table, and its strict decoder is exactly the shape we need — known name → text, unknown name or missing semicolon → input unchanged (checked on 8.0.0: `&colon;` → `:`, `&NotEqualTilde;` → two code points, `&notarealentity;` and `&copy` untouched):

```ts
import { decodeHTMLStrict } from 'entities';

// XML defines these itself; decoding them would turn escaped text into markup.
const XML_ENTITY_NAMES = new Set(['amp', 'lt', 'gt', 'quot', 'apos']);

export function numericizeHtmlEntities(svg: string): string {
  const declared = new Set([...svg.matchAll(/<!ENTITY\s+([A-Za-z][A-Za-z0-9]*)\s/g)].map((m) => m[1]));
  return svg.replace(
    /<!--[\s\S]*?-->|<!\[CDATA\[[\s\S]*?\]\]>|&([A-Za-z][A-Za-z0-9]{0,31});/g,
    (match, name?: string) => {
      if (name === undefined || XML_ENTITY_NAMES.has(name) || declared.has(name)) return match;

      const decoded = decodeHTMLStrict(match);
      if (decoded === match) return match;
      return [...decoded].map((c) => `&#${c.codePointAt(0)};`).join('');
    }
  );
}
```

That drops the whole table (and with it the typo/validation and 4.01-vs-5 code point questions), covers HTML5-only and multi-code-point names, and keeps the same skip rules. One new runtime dependency on a retry path is the trade — jsdom is already installed and also carries the table via parse5, but its decoding is the lenient browser kind (`&notarealentity;` → `¬arealentity;`), so it would need a guard nobody will trust.

src/utils.ts
```diff
@@ -45,10 +45,106 @@ export function sha256(str) {
...
+
+function isUndefinedEntityError(err: unknown): boolean {
+  return err instanceof Error && /Entity '[^']+' not defined/.test(err.message);
+}
+
+// The comment and CDATA alternatives match nothing worth rewriting. They are
+// there to consume those sections whole, so an entity name inside one stays
+// the literal text it is rather than becoming a character reference.
+//
+// Exported for the tests: which names this leaves alone is not observable
```
wa0x6e: Optional: this export exists only so the tests can see which names are left alone, and `utils.ts` is imported by nearly every resolver. Now that ~80 lines of static data plus two helpers live here, they'd sit more naturally in their own module (say `src/helpers/htmlEntities.ts`, exporting `numericizeHtmlEntities` and `isUndefinedEntityError`) with `resize()` importing from it — the export is then scoped to the thing being tested, and it's also the obvious home if the table ever moves to JSON with a generator alongside, the way `chains.json` is kept. Not a blocker; fine to leave for a follow-up.

src/utils.ts
tony8713: Right, and this one was mine: on your ISO-8859-1 case the retry rendered mojibake and cached it where the old code failed loudly. Reproduced it before changing anything, the 0xE9 came back as EF BF BD and the output no longer matched the correctly encoded document. Switched to latin1 and pinned it twice, a byte level round trip that holds anywhere and the render comparison you suggested.

src/utils.ts
tony8713: Taken. Reproduced with your pixel limit case: the retry parses, dies on `Input image exceeds pixel limit`, and the entity is what reached `capture()`. Attaching it as `cause` leaves the message alone and there is now a test asserting both halves.

test/unit/resize.test.ts
tony8713: Taken, the samples are pinned on the string now. One thing I could not confirm: I have 252 fonts here and even two code points with no glyph anywhere still render differently, and my attempt to simulate an empty fontconfig turned out non-deterministic, so I have no evidence either way about the runner. The string level pin is the better shape regardless, and switching to `entities` removes the transposed-entry risk entirely since there is no table left to transpose.

src/utils.ts
tony8713: Done. Every caller passes a Buffer and `tsc --noEmit` stays clean with it.

src/utils.ts
tony8713: Adopted, the table is gone. One correction worth knowing: `entities@8` is pure ESM with no CommonJS build, and jest 28 fails to parse it, so this pins `^7`, the last line that ships `dist/commonjs`. Behaviour is identical on everything you listed, `&colon;`, `&lang;` at U+27E8, `&NotEqualTilde;` as two code points, unknown names and missing semicolons untouched. I also took the `<!ENTITY>` guard, and since a declared name that is not a real one is left alone by the decoder anyway, the test declares `copy` so it actually bites.

tony8713:
Reviewed this adversarially against a fresh clone of 522c1f2, on the assumption that the tests have already been shown insufficient once. I could not break the behaviour. The latin1 round trip really is byte preserving, the comment and CDATA skip held up against every nesting I could build, and the success path is genuinely untouched rather than merely claimed to be. This is a yes from me. GitHub will not let me file a formal approving review on a PR opened by this same account, so this comment is the approval. Three things I would still fix, all in the tests except the last.

**The latin1 regression test only fails on a machine with a text font.** `keeps the bytes of a document that is not UTF-8` compares two rendered webps, and latin1 versus utf8 is only observable as glyphs. I put `utf8` back in both places in `resize()` and ran the suite under an empty fontconfig: 15/15 green, that test included. With the system fonts here it fails as intended, and ubuntu-latest has fonts, so CI does catch it today. Still, the single test guarding the regression that already shipped once is the one that depends on the box. `carries non-ASCII bytes through a latin1 round trip` does not close the gap, because it calls `numericizeHtmlEntities` with a string and the encoding decision lives in `resize()`. Asserting on the bytes the retry hands to sharp would hold everywhere.

**Two guards have no test that can fail.** Delete `if (!isUndefinedEntityError(err)) throw err;` and the suite is 15/15 green. Delete `if (rewritten === svg) throw err;` and it is 15/15 green again. `rethrows sharp errors that are not entity errors` survives the first deletion because `Buffer.from('not an image at all')` contains no `&name;`, so the second guard stops it anyway. One assertion each kills both: `resize(Buffer.from('not an image at all &copy;'))` rejects with `cause` undefined today and with `cause` set once the gate goes, and `rethrows unchanged when there is no entity it can rewrite` behaves the same way against the second guard.

**An `<!ENTITY` inside a comment silently disables the whole fix.** The `declared` scan runs over the raw document, so it never gets the comment and CDATA skip that the replace pass does. `<svg xmlns="http://www.w3.org/2000/svg" width="64" height="64"><!-- <!ENTITY nbsp "x"> --><text y="20">&nbsp;</text></svg>` renders through `resize()` without the comment and throws with it, straight back to the blockie. `<![CDATA[<!ENTITY nbsp "x">]]>` does the same. It is only ever a false negative, never a wrong render, but it is the same skip the previous round was about, and collecting the declarations in the same pass would close it. Processing instructions are the other sibling: `<?some-pi a="&copy;"?>` gets rewritten even though PI content is as literal as CDATA, and librsvg does act on PIs, so it is not inert. I could not turn that one into a visible difference.

Smaller: `does not rewrite a valid SVG carrying XML entities` never reaches `numericizeHtmlEntities`, so it does not test the five XML names, it is a third copy of the pass-through assertion. `leaves the names XML defines alone` does cover them.

What I attacked and could not break:

- Encoding. Retry output is byte identical to the same document written with `&#160;` on UTF-8 carrying CJK and emoji, on an explicit UTF-8 prolog, on a UTF-8 BOM, and on a prolog claiming ISO-8859-1 over UTF-8 bytes. UTF-16 is left alone rather than mangled, since the regex never matches and the original error is rethrown.
- Markup. The rewrite only ever emits `&#N;`, so no name can be turned back into a raw `<`, `&` or quote. Astral names come out right (`&ascr;` to `&#119990;`), the longest HTML5 name is 31 characters so `{0,31}` fits it, and every name I could enumerate out of the table matches the regex charset.
- Comment and CDATA nesting. CDATA holding a comment open, a comment holding a CDATA open, interleaved, unterminated, a fake nested comment. Leftmost alternation lands on the same token boundaries libxml2 does in all of them.
- ReDoS on the lazy `[\s\S]*?`. It is quadratic, but not reachable: enough unterminated comment openers to matter dies on "Double hyphen within comment" before any entity error is reported.
- An entity expansion bomb hidden behind a named entity. libxml2 caps the amplification with or without the retry.
- `entities`. The lockfile resolves 7.0.1, there is only one copy, and the `require` condition points at `dist/commonjs`. `decodeHTMLStrict` returns the input unchanged for an unknown name and for a missing semicolon, which is what the code assumes.
- The error path. The rethrown object is the original, its message names the caller's entity and `cause` names the retry's.
- Blast radius. `tsc --noEmit` and eslint clean, unit suite 105/105, and three of the "inputs that already parse" tests never enter the entity code at all.

wa0x6e (approved):
utAck

PR #528, opened 2026-08-15
refactor(lookupDomains): report Shibarium failures through index.ts

tony8713:
`src/lookupDomains/shibarium.ts` caught its own failures, classified them against `isSilencedError`, called `capture` itself and returned whatever pages it had already collected. `ens` and `unstoppableDomains` throw and let the fan-out in `index.ts` do all three. So Shibarium's errors never reached the shared handler and never carried the context it adds.

It now throws. `index.ts` catches, classifies and captures, which is what the other two providers do. Both `try/catch` blocks are gone along with the `capture` and `isSilencedError` imports; the `finally` that clears the abort timer stays, and a 404 still breaks the pagination loop as no-data rather than an error.

## This changes what a caller gets, not just what Sentry gets

Worth stating plainly rather than filing under unifying error reporting.

Today, if page 1 of a paginated fetch succeeds and page 2 fails, Shibarium reports the failure and returns page 1. The caller receives a list that looks complete and is not. After this change it returns nothing for that provider, matching `ens` and `unstoppableDomains`.

A partial list that looks complete is arguably worse than an empty one, especially since the caller has no way to tell the two apart, but it is a real behaviour change and it is visible through the `lookup_domains` API method.

Nothing depends on the old behaviour. The only consumer of the provider is the registry in `index.ts`, and thence `api.ts`. The only other importer is its own test.

## What does not change

The classification. `index.ts` calls the same `isSilencedError` with the same default mute list, one level up, so a 429 or a 504 is still silenced and still produces no Sentry event. There is a test for that, since it is the thing most likely to regress unnoticed.

One small loss: the old internal capture carried `skip`, so a report said which page failed. The shared handler carries the address and the chain id, like the other providers. That could be added later if it turns out to matter, but inventing a mechanism to carry it through the throw did not seem worth it here.

## Tests

Six of the eight cases in `test/integration/lookupDomains/shibarium.test.ts` changed, because their subject moved rather than because they broke. Five asserted that the provider captures and returns empty, and now assert that it throws carrying the HTTP status, which is the shape `test/integration/lookupDomains/unstoppableDomains.test.ts` already uses for the same contract. The sixth, `keeps the domains collected before a rate limit interrupts pagination`, asserted exactly the partial-return behaviour being removed, and is now `throws when a later page fails, instead of returning the pages already collected`.

`returns the domains on a successful response` and `does not call the API on an unsupported chain` are unedited.

Three cases are added, driving the real `index.ts` with `node-fetch` mocked, to prove the failure now travels the shared path: that a 500 is captured with the address and chain as context, that a 429 is still silenced, and that the earlier pages are dropped.

Part of #511.

### DIFF
--- a/src/lookupDomains/shibarium.ts
+++ b/src/lookupDomains/shibarium.ts
@@ -1,6 +1,4 @@
-import { capture } from '@snapshot-labs/snapshot-sentry';
 import fetch from 'node-fetch';
-import { isSilencedError } from '../addressResolvers/utils';
 import constants from '../constants.json';
 import { Address, Handle } from '../utils';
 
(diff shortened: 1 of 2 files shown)


tony8713:
Adversarial review, verdict approve. I tried to break this and could not; two non blocking notes.

**The justification does not fit the case the new test uses.** `hasMore = domains.length === PAGE_SIZE`, so an address owning an exact multiple of 25 domains always makes one extra request that carries no data. A 429 on that probe used to return the complete, correct 25 and now returns nothing, so the list was not partial and "a partial list that looks complete is worse than an empty one" is not the trade being made there. The fixture in `throws when a later page fails, instead of returning the pages already collected` is exactly that shape, `page(PAGE_SIZE)` then 429, with nothing left uncollected.

It is worth saying because 429 is silenced, so nothing is reported, and `apps/hub/src/graphql/helpers.ts` in sx-monorepo feeds this result straight into `where.id_in` for `spaces(where: { controller })`. A Shibarium rate limit now drops a user's Shibarium spaces out of their own space list with no signal anywhere. Only addresses with 25 or more Shibarium domains are affected and I do not think keeping the partial return is the answer, but I would put this in the description in place of the partial list framing.

**504 and 401 lost their only test.** `expect(capture).not.toHaveBeenCalled()` in the new `it.each` cannot fail now that `shibarium.ts` no longer imports `capture`, so collapsing the four statuses into one case leaves 504 silencing and 401 capturing asserted nowhere. 504 is the fragile one, since `HTTP 504: Gateway Timeout` does not contain `status=504` and is therefore silenced only through the numeric `error.status`. Two more rows in the through-the-shared-handler describe would cover both.

**What survived.** I ran 18 error scenarios through `index.ts` on master and on this branch and diffed the observable result. Classification is identical on every one: 429, 504, ECONNRESET, ETIMEDOUT and 404 silenced on both, and 401, 403, 500, 502, ENOTFOUND, AbortError and invalid JSON captured exactly once on both with the same message and the same `status`. The only context delta is the documented loss of `skip`, and only three rows change the returned value, all of them the intended ones.

The `finally` holds. Counting the 10000ms abort timers across eight paths, including the 404 break on page 2, every one is cleared and none is left live, and a hanging request is still aborted at 10s and reported through the shared handler. The unsupported chain path still short circuits for 1, 146, 157, test and the empty string, with zero fetches and no throw, and a 404 part way through pagination still returns the earlier pages on both branches. `typecheck`, `eslint src/ test/` and all 136 unit tests are green on the branch.

Posting this as a comment rather than a formal review because GitHub will not let this account approve its own pull request.

wa0x6e (approved):
utAck

PR #529, opened 2026-08-15
refactor(resolvers): let the map own the resolver failure contract

tony8713:
Each individually written resolver ended in a whole-body `catch { return false }`, so "this resolver broke" and "this address has no avatar" were the same answer, written out separately in eight files. That answer moves to the wrapper in `index.ts`, where the rest of the pipeline already lives, and comes out of the eight.

## This is smaller than removing all the catches, deliberately

The shared factories keep theirs. `createPropertyResolver` in `snapshot.ts` produces five entries and the one in `space-sx.ts` produces two, and in both cases the factory serves bare entries as well as wrapped ones. `user-cover`, `space-cover`, `space-logo` and `space-cover-sx` have no wrapper to fall back on, so taking the catch out of the factory would take it away from them in the same edit and let them throw into the image route, which has no guard of its own (#495). `blockie` and `jazzicon` stay outside the contract for the same reason: `api.ts` hands the fallback resolver's result straight to `resize()`, so a fallback answering `false` would make sharp throw on the same unguarded path.

What makes the eight safe is that they line up exactly with the wrapped entries. Every individually written resolver that had a catch is `resize: true`, and every `resize: false` entry is either factory-produced or has no catch to move. So the set of entries covered by a catch is identical before and after. The factories come out once the flag is split and every entry has a wrapper, which is the next step, not this one.

## What changes for a caller

Nothing, and that is the intent.

`src/resolvers/index.ts` is the only production consumer of the resolver modules, so every production call already goes through the map. Before, the resolver caught and answered `false`; now the wrapper catches and answers `false`. Same value, same fallback, same path.

No new Sentry volume either. The wrapper keeps two policies and they are deliberately different: a resolver throwing is silent, a sharp failure is still reported exactly as it is today. Silent is not the end state, it is the correct state until the no-data answers are normalized. Turning capture on now would report every routine `metadata.ens.domains` 404 and every token without a trustwallet logo as an error, which is the ordering constraint #511 was built around.

## One existing test moved

`test/unit/graphQlEnvelope.test.ts`, the `src/resolvers/lens.ts` case, asserted `lensResolve(...)` resolves to `false`. It imports the module directly rather than going through the map, so it was asserting the contract this PR moves. It now asserts the rejection, and the assertion that carries the test's actual point, that a payload the upstream flagged is never fetched, is unchanged.

The `snapshot.ts` and `space-sx.ts` cases beside it still assert `false` and needed no edit, which is the scoping above visible in the test file.

## On #483

It does not unblock it. Worth saying plainly, because the opposite was the expectation.

`src/resolvers/ens.ts` is one of the eight and #483 conflicts there, but touching the same file does not clear a conflict. Merging #483 against master and against this branch produces the same three conflicted paths either way: `package.json`, `src/resolvers/ens.ts` and `yarn.lock`. All three already conflict on master today.

What does change is the shape of the resolution. After this, the target for that file is the fetch logic on its own, with no `try`/`catch` and no `resize`, so #483 keeps its Universal Resolver body and drops both wrappers rather than dropping one and keeping the other. Easier to resolve, still a manual resolve.

Refs #511.

### DIFF
--- a/src/resolvers/basename.ts
+++ b/src/resolvers/basename.ts
@@ -2,12 +2,8 @@ import { fetchHttpImage } from './utils';
 import { getAvatar } from '../addressResolvers/basename';
 
 export default async function resolve(nameOrAddress: string) {
-  try {
-    const url = await getAvatar(nameOrAddress);
-    if (!url) return false;
+  const url = await getAvatar(nameOrAddress);
+  if (!url) return false;
 
-    return await fetchHttpImage(url);
-  } catch {
-    return false;
-  }
+  return await fetchHttpImage(url);
 }
--- a/src/resolvers/coingecko.ts
+++ b/src/resolvers/coingecko.ts
@@ -15,19 +15,15 @@ const COINGECKO_ASSET_PLATFORMS = {
 export default async function resolve(address: string, chainId: string) {
   if (!API_KEY) return false;
 
-  try {
-    const assetPlatformId = COINGECKO_ASSET_PLATFORMS[chainId];
+  const assetPlatformId = COINGECKO_ASSET_PLATFORMS[chainId];
 
-    if (!assetPlatformId) return false;
+  if (!assetPlatformId) return false;
 
-    const checksum = getAddress(address);
-    const url = `https://pro-api.coingecko.com/api/v3/coins/${assetPlatformId}/contract/${checksum}`;
+  const checksum = getAddress(address);
+  const url = `https://pro-api.coingecko.com/api/v3/coins/${assetPlatformId}/contract/${checksum}`;
 
-    const data = await fetch(url, { headers: { 'x-cg-pro-api-key': API_KEY } }).then(res =>
-      res.json()
-    );
-    return await fetchHttpImage(data?.image?.large);
-  } catch {
-    return false;
-  }
+  const data = await fetch(url, { headers: { 'x-cg-pro-api-key': API_KEY } }).then(res =>
+    res.json()
+  );
+  return await fetchHttpImage(data?.image?.large);
 }
--- a/src/resolvers/ens.ts
+++ b/src/resolvers/ens.ts
@@ -12,23 +12,19 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
 }
 
 export default async function resolve(nameOrAddress: string) {
-  try {
-    const provider = getProvider(1);
-    const ensName = await castToEnsName(nameOrAddress);
+  const provider = getProvider(1);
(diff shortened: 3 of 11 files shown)


tony8713:
LGTM. I went at the central claim rather than spot-checking it and could not break it.

The set of entries covered by a catch really is identical. The eight losing one are `basename`, `coingecko`, `ens`, `farcaster`, `lens`, `selfid`, `starknet`, `trustwallet`, all `resize: true`; of the six `resize: false` entries, four come from the two factories that keep theirs and `blockie`/`jazzicon` never had one. I checked that by behaviour rather than by reading: breaking the transport under all 17 registry entries and calling each through the map gives identical answers on this branch and on its parent, `false` everywhere except `blockie` and `jazzicon`, which still throw.

Both capture policies hold too. Feeding bytes sharp refuses through the same entries is also identical to the parent, one `capture` on the wrapped ones and none on the `resize: false` ones, which keep handing the bad bytes to the unguarded `resize()` in `api.ts` exactly as before. A resolver throw reaches the wrapper's bare `catch` and nothing else, so there is no new Sentry volume on a routine 404 or a logo-less token.

`farcaster.ts`'s two remaining catches are local and unmoved. `normalizeAddress` turns an unparseable address into `null` and `fetchAddressImageUrl` turns a Neynar failure into `null`, both answering `null` rather than `false`, both predating the contract, and the diff touches neither.

The `graphQlEnvelope.test.ts` change tracks the move rather than hiding a regression. `graphQlCall` throws on a flagged envelope and `lens.ts` no longer catches it, so `rejects.toThrow` is the honest assertion, and the `fetchHttpImage` line that carries the test's actual point is unchanged. The two cases beside it still assert `false` because both resolvers are factory-produced and the factory catch is still there, so that reading is right.

`src/api.ts` is the only production consumer. Nothing under `src/` imports a resolver module directly and there are no dynamic imports of one; the other two `constants.resolvers` hits are a route-label regex in `helpers/metrics.ts` and an unrelated local variable in `addressResolvers/spaceId.ts`. In the tests, `graphQlEnvelope.test.ts` is the only file importing modules directly, and it is the one you updated.

`failureContract.test.ts` is a real guard rather than a restatement: dropping the wrapper's resolver catch, adding `capture` to it, dropping the sharp `capture`, and flipping `blockie` to `resize: true` each fail it.

Also confirmed the #483 note, merging it into master and into this branch conflicts on the same three paths.

(Posted as a comment: GitHub will not take a formal approving review from the account that opened the PR.)

wa0x6e (approved):
utAck

PR #532, opened 2026-08-15
feat(lookupDomains): record each provider in a metric, and stop reporting our own aborts

tony8713:
A Shibarium abort was the only provider timeout still reaching Sentry. The other two providers already time out into errors the shared classifier silences, so this one was loud by accident rather than by decision.

The fix is a name check in `isSilencedError` rather than a mute list on the provider. Every transport words an abort differently, so matching the message pins the behaviour to one library at one version, while the name is stable. It belongs in the shared classifier because nothing here aborts a request except one of our own deadlines, which is the same reasoning that already silences the other timeout codes.

Silencing on its own would have been a straight loss, and that is what the histogram is for. `addressResolvers` can afford to mute because it times every resolver and records the failure whether or not it reports it, so the mute hides the report and not the event. `lookupDomains` had no metric at all, so this adds the matching per-provider histogram and records the status in a `finally`, meaning a silenced failure still counts. The two halves are only correct together, which is why they are one PR.

An earlier revision of this branch added a per-provider `mutedErrors` field and put the abort in Shibarium's. That field is gone. It keyed on an English string, so a transport or version change would have turned the mute off with the suite still green, and the classifier fix leaves it with nothing to hold.

Part of #511. No closing keyword, since the resolvers half of that issue is still open.

### DIFF
--- a/src/addressResolvers/utils.ts
+++ b/src/addressResolvers/utils.ts
@@ -72,6 +72,9 @@ export function normalizeHandles(handles: Handle[]): Handle[] {
 }
 
 export function isSilencedError(error: any, additionalMessages?: string[]): boolean {
+  // An abort is always one of our own deadlines, and each transport words it differently.
+  if (error.name === 'AbortError') return true;
+
   const messages = [
     'invalid token ID',
     'is not supported',
--- a/src/helpers/metrics.ts
+++ b/src/helpers/metrics.ts
@@ -26,6 +26,12 @@ export const timeAddressResolverResponse = new client.Histogram({
   labelNames: ['provider', 'method', 'status']
 });
 
+export const timeLookupDomainsResponse = new client.Histogram({
+  name: 'lookup_domains_response_duration_seconds',
+  help: "Duration in seconds of each domain provider's response.",
+  labelNames: ['provider', 'chainId', 'status']
+});
+
 export const addressResolversCacheHitCount = new client.Counter({
   name: 'address_resolvers_cache_hit_count',
   help: 'Number of hit/miss of the address resolvers cache layer',
--- a/src/lookupDomains/index.ts
+++ b/src/lookupDomains/index.ts
@@ -1,6 +1,7 @@
 import { isAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import { isSilencedError } from '../addressResolvers/utils';
+import { timeLookupDomainsResponse as timeResponse } from '../helpers/metrics';
 import { Address, Handle } from '../utils';
 import ens, {
   CHAIN_IDS as ENS_CHAIN_IDS,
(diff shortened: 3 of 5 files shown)


tony8713:
Reviewed adversarially: cloned it, measured the three timeouts against a socket that accepts the connection and never answers, and mutated the source to check the new tests can fail. Approving. GitHub will not accept a formal approving review from the account that opened the PR, so this is an issue comment rather than an Approve.

**Parity holds on all three.** Measured on this lockfile, where `node-fetch` resolves to a single version (2.7.0):

- ens, through `graphQlCall`'s axios `timeout: 5e3`, rejects with `timeout of 5000ms exceeded` carrying `code: 'ECONNABORTED'`. `isSilencedError` returns true today.
- unstoppableDomains, on undici's default `headersTimeout`, rejects after 303194 ms with `TypeError: fetch failed` whose `cause.code` is `UND_ERR_HEADERS_TIMEOUT`. `isSilencedError` returns true today, off `error.cause?.code`.
- shibarium rejects with `AbortError: The user aborted a request.` and `code`, `status` and `cause` all undefined. `isSilencedError` returns false today and true with the new list.

**The blast radius is as narrow as the description claims.** The provider can only throw `HTTP <status>: <statusText>`, `Invalid JSON response: <message>`, or whatever node-fetch raises, and the controller is created inside the pagination loop with no caller-supplied signal, so there is no caller-initiated abort to swallow. The one case I expected to escape does not: when the deadline fires after the headers land, node-fetch re-emits the same AbortError on the response body, the provider wraps it as `Invalid JSON response: The user aborted a request.`, and the substring still matches. Checked against a server that writes `{"pageItems":` and then hangs, so the mute covers the whole 10s window rather than just the connect phase.

**`undefined` reads as the default list.** `isSilencedError(new Error('execution reverted'), undefined)` is still true and `isSilencedError(new Error('boom'), undefined)` is still false. The untouched unit mocks leave `mutedErrors` undefined for all three providers, and nothing in `test/unit/lookupDomains.test.ts` asserts anything that depends on the shibarium list, so no assertion there goes vacuous.

**Both new tests discriminate.** Seven source mutations, each killing the right one: dropping the `mutedErrors` argument at the call site, unwiring the registry field, and replacing the literal each fail only `does not report the abort`; dropping `signal: controller.signal`, making the timer a no-op, and moving `TIMEOUT` to 9000 each fail only `still aborts the request`.

Two things worth a follow-up, neither blocking.

The mute test passes vacuously on its own. Make the provider `return []` before it ever reaches fetch and `does not report the abort` stays green, since an empty result with no `capture` is what it asserts either way. It only fails today because its sibling in the same describe catches that mutation. `expect(mockedFetch).toHaveBeenCalled()` closes it, and that is the shape #528 landed in.

Nothing ties `'The user aborted a request.'` back to node-fetch. Both new tests build the error by hand and node-fetch is mocked in the only file that imports it, so no test in the repo ever sees the real string. node-fetch 3.3.2 raises `The operation was aborted.` and this Node's global fetch raises `This operation was aborted`, both measured, so a major bump or a move to global fetch switches the mute off with the suite still green. One assertion driving a real node-fetch abort against a local socket would pin it.

wa0x6e (approved):
utAck

PR #533, opened 2026-08-16
refactor: simplify the lookupDomains and resolvers registries

tony8713:
Follow-up to the lookupDomains and resolvers series that just merged, not a fix. It changes no behaviour: no runtime path, no error policy and no reported payload moves. A quality pass over the lines that series touched, and nothing else.

Four commits, each independent, so any one can be dropped without disturbing the others.

## 1. `refactor(lookupDomains): let the modules be the provider registry`

The registry restated each provider's exported constants under new lowercase names, which took a block of aliased imports per provider to feed. The providers already export the names the registry wants, and the address resolver registry next door already imports its providers as namespaces, so the module can be the provider directly and a mis-wired alias stops being representable.

## 2. `refactor(lookupDomains): fan out with flatMap instead of an accumulator`

Drops the mutable promise array the fan-out pushed into.

**This is the one commit that leaves the series' own lines.** The accumulator and its `Promise.all` predate the series; only the nested loop between them was series code. Drop this commit if the pass should stay strictly inside those lines. The other three do not depend on it.

## 3. `refactor(resolvers): keep each resolver's own parameters on the map`

The exported map was cast to a record keyed by resolver name with every signature flattened to a rest tuple of `any`. The names were checked, the arguments were not, so calling a resolver with the wrong arity compiled. Keying the map off the registry entries keeps both, since those entries already carry the real function types.

The return type stays wider than the entry's own on purpose. The resize wrapper answers `false` for a resolver that throws, including resolvers whose own signature says they only ever return a buffer, so taking the return type from the entry would let a caller treat one of those as never failing.

## 4. `test(lookupDomains): type the shibarium status table`

The untyped `it.each` table widened its columns, so each case cast them back at the call site. Annotating the table removes the casts.

### DIFF
--- a/src/lookupDomains/index.ts
+++ b/src/lookupDomains/index.ts
@@ -3,51 +3,21 @@ import { capture } from '@snapshot-labs/snapshot-sentry';
 import { isSilencedError } from '../addressResolvers/utils';
 import { timeLookupDomainsResponse as timeResponse } from '../helpers/metrics';
 import { Address, Handle } from '../utils';
-import ens, {
-  CHAIN_IDS as ENS_CHAIN_IDS,
-  DEFAULT_CHAIN_ID as ENS_DEFAULT_CHAIN_ID,
-  NAME as ENS_NAME
-} from './ens';
-import shibarium, {
-  CHAIN_IDS as SHIBARIUM_CHAIN_IDS,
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
utAck

PR #534, opened 2026-08-16
fix(lookupDomains): bound every provider call with a shared deadline

### LINKED ISSUE #531: lookupDomains/unstoppableDomains sets no request deadline, so a hung upstream stalls the whole call
This is a report of measured behaviour, taken on 2026-08-15 on Node v22.23.1 against a socket that accepts the connection and then never responds.

`src/lookupDomains/unstoppableDomains.ts` calls global `fetch` with no `signal` and no timeout. Both its siblings set a deadline: `ens` gets 5s from the axios `timeout: 5e3` inside `graphQlCall`, and `shibarium` sets a 10s `AbortController`. So this is one provider out of step rather than a general gap.

Measured, all three against the same hanging socket:

| provider | deadline set | time to throw | error | silenced today |
|---|---|---|---|---|
| ens | axios `timeout: 5e3` | 5.1s | `ECONNABORTED` | yes |
| shibarium | `AbortController`, 10s | 10.0s | `AbortError` | no |
| unstoppableDomains | none | 301.1s | `TypeError: fetch failed`, cause `UND_ERR_HEADERS_TIMEOUT` | yes |

The 301s is undici's own headers timeout. It is the only thing bounding the call.

## Why it matters

Unstoppable Domains serves chain 146, which is in the default chain set, so an ordinary `lookup_domains` call that passes no `network` includes it. This is the common path, not an edge case.

`index.ts` ends in `await Promise.all(promises)`. The per-promise `.catch` from #493 makes the fan-out best effort against a provider that *fails*, but a promise that never settles is never caught, so it does not cover a provider that never answers. ENS returning in under 5s buys nothing. The caller gets nothing for up to five minutes and then a 200 carrying only the ENS names.

It also fails silently when it finally gives up: `UND_ERR_HEADERS_TIMEOUT` contains the substring `TIMEOUT`, which `isSilencedError` already matches. Slow and invisible.

## Two second-order points

The ceiling is per page rather than per call, and the provider paginates in a `while (cursor !== null)` loop, so the total is unbounded in the number of pages. Shibarium's 10s has the same per-page shape. That matters far less at 10s, but it is the same latent shape and worth knowing before anyone treats 10s as the worst case.

`src/getOwner/shibarium.ts` also calls global `fetch` with no signal, so `get_owner` carries the same exposure.

## What I could not establish

There is no `fly.toml` and no `Dockerfile` in this repo, so I could not confirm or rule out a platform proxy timeout in front of the service. A proxy may well cut the connection well before 300s, which would cap what the client waits for but not what the process does.

What I can say is that there is no backstop in the application: `src/index.ts` calls `app.listen(PORT)` with no `server.setTimeout` and no timeout middleware, and Node's `server.requestTimeout` bounds reading the incoming request rather than how long a handler may run.

## Shape of a fix, as a proposal rather than a decision

Shibarium already has the in-repo pattern, an `AbortController` against a constant, and copying it into `unstoppableDomains` would be the smallest change. A single shared deadline applied where the providers are called would be better than a third hand-rolled `AbortController`, since the fan-out already knows it is calling several untrusted upstreams and already owns the failure contract for them. Not designed further here.

Related: #493, which made this fan-out best effort against failures. #495 is a different route with a similar symptom.

tony8713:
`lookupDomains/unstoppableDomains` called the global `fetch` with no signal and
no timeout, so a provider that accepts the connection and then never answers
held the whole `lookup_domains` call open. The fan-out awaits `Promise.all`, and
the per-promise catch from #493 covers a provider that *fails* rather than one
that never settles, so nothing in the application bounded the wait. Its two
siblings each set a deadline of their own, in two different ways, and
`getOwner/shibarium` set none at all.

This puts the deadline behind one `withDeadline` helper and uses it in all three
places. It wraps a whole provider call rather than a single request, so the
providers that paginate are bounded by it too. Where the callee takes no signal
at all, as `getOwner`'s DNS resolution does not, `untilAborted` bounds the wait
instead: the request it started keeps running, but nothing waits on it past the
deadline.

Aborting through an `AbortController` rather than `AbortSignal.timeout` is what
keeps the error ours: `isSilencedError` matches `AbortError` by name. In
`lookupDomains` that means a deadline raises no Sentry event and is still
recorded in the per-provider histogram as a failure. `getOwner` has neither. It
reports through the unconditional capture in `api.ts`, as its undici headers
timeout already did, so there the deadline changes when it reports, not whether.

Closes #531

### DIFF
--- a/src/getOwner/shibarium.ts
+++ b/src/getOwner/shibarium.ts
@@ -1,6 +1,6 @@
 import { DNSConnect } from '@webinterop/dns-connect';
 import constants from '../constants.json';
-import { Address, EMPTY_ADDRESS, Handle, withDeadline } from '../utils';
+import { Address, EMPTY_ADDRESS, Handle, untilAborted, withDeadline } from '../utils';
 
 const MAINNET = '109';
 const TESTNET = '157';
@@ -46,7 +46,12 @@ async function getClaimedOwner(handle: Handle, chainId: string): Promise<Address
 async function getResolvedAddress(handle: Handle, chainId: string): Promise<Address> {
   const dnsConnect = new DNSConnect({ dns: { forwarderDomain: constants.d3[chainId].forwarder } });
 
-  return (await dnsConnect.resolve(handle, NETWORK)) || EMPTY_ADDRESS;
+  // dns-connect passes no signal to the DNS-over-HTTPS fetches it makes, so
+  // only the wait can be bounded here, not the request.
+  return withDeadline(
+    async signal =>
+      (await untilAborted(signal, dnsConnect.resolve(handle, NETWORK))) || EMPTY_ADDRESS
+  );
 }
 
 /**
--- a/src/getOwner/shibarium.ts
+++ b/src/getOwner/shibarium.ts
@@ -1,6 +1,6 @@
 import { DNSConnect } from '@webinterop/dns-connect';
 import constants from '../constants.json';
-import { Address, EMPTY_ADDRESS, Handle } from '../utils';
+import { Address, EMPTY_ADDRESS, Handle, withDeadline } from '../utils';
 
 const MAINNET = '109';
 const TESTNET = '157';
@@ -16,28 +16,31 @@ async function getClaimedOwner(handle: Handle, chainId: string): Promise<Address
   if (!handle.endsWith(`.${TLD}`) || !constants.d3[chainId]?.apiUrl || !API_KEYS[chainId])
     return EMPTY_ADDRESS;
 
-  const response = await fetch(
-    `${constants.d3[chainId].apiUrl}/v1/partner/token/${handle.replace(/\.shib$/, '')}/${TLD}`,
-    {
-      method: 'GET',
-      headers: {
-        accept: 'application/json',
-        'Api-Key': API_KEYS[chainId]
+  return withDeadline<Address | false>(async signal => {
+    const response = await fetch(
+      `${constants.d3[chainId].apiUrl}/v1/partner/token/${handle.replace(/\.shib$/, '')}/${TLD}`,
+      {
+        method: 'GET',
+        headers: {
+          accept: 'application/json',
+          'Api-Key': API_KEYS[chainId]
+        },
+        signal
       }
+    );
+
+    if (response.status === 404) return EMPTY_ADDRESS;
+    if (!response.ok) throw new Error(`Error fetching owner: ${response.statusText}`);
+
+    const data = await response.json();
+
+    if (data.status !== 'registered' || new Date(data.expirationDate) < new Date()) {
+      return EMPTY_ADDRESS;
     }
-  );
 
-  if (response.status === 404) return EMPTY_ADDRESS;
-  if (!response.ok) throw new Error(`Error fetching owner: ${response.statusText}`);
-
-  const data = await response.json();
-
-  if (data.status !== 'registered' || new Date(data.expirationDate) < new Date()) {
-    return EMPTY_ADDRESS;
-  }
-
-  // owner field will be missing on unclaimed names
-  return data.owner || false;
+    // owner field will be missing on unclaimed names
+    return data.owner || false;
+  });
 }
 
 async function getResolvedAddress(handle: Handle, chainId: string): Promise<Address> {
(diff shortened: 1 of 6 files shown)


src/getOwner/shibarium.ts
```diff
@@ -16,28 +16,31 @@ async function getClaimedOwner(handle: Handle, chainId: string): Promise<Address
...
+    }
 
-  // owner field will be missing on unclaimed names
-  return data.owner || false;
+    // owner field will be missing on unclaimed names
+    return data.owner || false;
+  });
 }
 
 async function getResolvedAddress(handle: Handle, chainId: string): Promise<Address> {
```
wa0x6e, line 46: `getResolvedAddress` (lines 46-49, not touched by this diff) still calls `dnsConnect.resolve()` with no deadline: `@webinterop/dns-connect`'s `DNSOverHTTPSResolver` does a bare `fetch()` with no signal/timeout, and `DNSOverHTTPSResolverOptions` is `{ logger, dnsServer }` only (0.4.0 is the same, plus `async-retry` × 3 around that fetch — worse for a hang). This path is hit for every unclaimed `.shib` name (see this function's docstring), worst case 3 serial DoH fetches each backstopped only by undici's ~300s headers timeout — the exact exposure #531 measured, so `get_owner` is not bounded end-to-end yet.

Note `withDeadline` can't bound this as-is: it has no internal race, it only works because callers pass its `signal` into `fetch`; a callee that ignores the signal gets no protection.

Smallest fix — bound the wait locally (the DoH request keeps running after the deadline, capped by undici):

```ts
async function getResolvedAddress(handle: Handle, chainId: string): Promise<Address> {
  const dnsConnect = new DNSConnect({ dns: { forwarderDomain: constants.d3[chainId].forwarder } });

  // dns-connect takes no signal, so this bounds the wait rather than the
  // request: the DNS-over-HTTPS fetch it makes keeps running after the
  // deadline fires.
  let timer: NodeJS.Timeout;
  const deadline = new Promise<never>((_, reject) => {
    timer = setTimeout(() => reject(new Error('DNS resolution timed out')), UPSTREAM_TIMEOUT);
  });

  try {
    return (await Promise.race([dnsConnect.resolve(handle, NETWORK), deadline])) || EMPTY_ADDRESS;
  } finally {
    clearTimeout(timer!);
  }
}
```

(needs `export const UPSTREAM_TIMEOUT` in `utils.ts`.)

If real cancellation is preferred: `DNSModuleOptions.resolver?: DNSResolver` is a public seam — one method, `resolve(hostname, type, dnssecVerification)`. A local resolver that does the same `dns-json` fetch against `https://cloudflare-dns.com/dns-query` with the `withDeadline` signal composes with this PR's helper directly, at the cost of ~25 lines copied from the vendor's `DNSOverHTTPSResolver` (NXDomain → `undefined`, non-NoError → throw, DNSSEC `AD` check).

Either way, worth narrowing the PR body's "uses it in all three places" so it doesn't read as `get_owner` bounded end-to-end.

tony8713: Confirmed and fixed. Reproduced it first with only cloudflare-dns pointed at a socket that accepts and never answers: `get_owner` on an unclaimed name was still pending at 15s, and now ends at 10s.

I bounded the wait rather than taking the resolver seam. Copying the vendor's NXDomain and DNSSEC handling in to buy cancellation we can't otherwise use is the worse trade, the AD check especially.

One deviation from your patch: the race went into a small `untilAborted(signal, work)` beside `withDeadline` rather than a local timer, so it reuses the controller that already exists and there's no fourth hand-rolled deadline. I did try folding it into `withDeadline` itself and that is wrong, it makes the paginating providers reject with a synthesised AbortError instead of the transport's own, which `lookupDomains/shibarium`'s deadline test catches.

Two notes. 0.4.0 wouldn't have helped, it's the same signal-less fetch with `async-retry` at 3 retries wrapped around it. And I left "all three places" in the body alone since `get_owner` is bounded end to end now, adding a sentence on wait versus request instead. `addressResolvers/shibarium.ts` has the same DNSConnect exposure, but that's #536 and a different route, so I kept it out of here.

wa0x6e (approved):
utAck

PR #535, opened 2026-08-16
feat(resolvers): tag captured failures with the provider that produced them

tony8713:
`lookupDomains` already tags the provider on every capture. These are the two families that still do not, so the same fact is now sent the same way everywhere.

`src/addressResolvers/index.ts` reported nine resolvers under one shape with nothing to tell them apart, though the name was already in scope as a Prometheus label two lines up. `src/resolvers/index.ts` reported sharp failures as a bare libvips message with neither a name nor an input, which is why tracing one back to an address was a manual job.

Two things that are not one-liners:

**The shape has to be the explicit two-key form.** Writing `{ input, tags }` looks correct and compiles, and silently loses the input. The auto-wrap that turns a bare `{ input }` into `{ contexts: { input } }` only runs when the object carries no reserved key, so adding `tags` skips it and leaves `input` as a top-level key that is not part of the event. The naive form and the correct one, through the real normalizer:

```
NAIVE   {input,tags}      -> {"input":{...},"tags":{"provider":"Ens"}}              input dropped
THIS PR {tags,contexts}   -> {"tags":{"provider":"Ens"},"contexts":{"input":{...}}} input kept
```

Both existing assertions in `test/unit/addressResolvers.test.ts` pinned the old shape and move with it. The resolver map had no capture assertion, so one is added. Rewriting either call as the one-liner compiles clean and turns those tests red, which is the point of pinning them.

**The resolver map did not have the name in scope.** `withResize` received only the function, so it takes the name too and the map build passes it.

## Comments

The block above `withResize` is gone. Testing it line by line, everything in it either restated the code or explained the reasoning, and both catches are already pinned by tests. What survived is the one thing no test states: the resolver catch is silent deliberately, because capturing there reports every routine upstream 404. That is now one line at the catch it applies to. The issue references went with the rest.

The same trim applies to a comment in the test file, which carried a reference for the same reason.

### DIFF
--- a/src/addressResolvers/index.ts
+++ b/src/addressResolvers/index.ts
@@ -70,7 +70,11 @@ async function _call(fnName: string, input: string[], maxInputLength: number) {
               status = 1;
             } catch (err) {
               if (!isSilencedError(err, r.MUTED_ERRORS)) {
-                capture(err, { input: { [fnName]: _input } });
+                // A top-level `input` beside `tags` is dropped rather than wrapped.
+                capture(err, {
+                  tags: { provider: r.NAME },
+                  contexts: { input: { [fnName]: _input } }
+                });
               }
             }
             end({ status });
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -28,23 +28,14 @@ type Resolver = {
   resize: boolean;
 };
 
-// Two failures, two policies, deliberately different.
-// A resolver throwing answers false and is not reported. That is the contract
-// each resolver used to implement for itself; it stays silent because the
-// no-data answers are not normalized yet (#511), and reporting before they are
-// would page on every routine upstream 404.
-// Sharp refusing the bytes is reported, because bytes we fetched but cannot
-// process are our problem rather than the upstream's.
-// Either way the wrapper answers false rather than throwing into the image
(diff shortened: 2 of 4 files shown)


tony8713:
Right, and I reproduced it before fixing it. Your mutation, `r.NAME` to `'Ens'` and `entry.name` to `'ens'`, gave me `tsc --noEmit` exit 0 and `Tests: 13 passed, 13 total`. Every capture attributed to one provider, suite green. That is the regression the PR is for, so the tests were not doing their job.

Both are now table driven off the registered providers rather than one case each.

`addressResolvers` drives all nine off the `RESOLVERS` array the file already had, taking the expected tag from each module's own `NAME` export, so a provider added to the registry is covered without touching the test. The resolver map drives four wrapped entries; it mocks per module, so it is the individually written single-argument ones rather than the whole registry. The factory-produced entries share a module and would need a different setup.

Same mutation against the new tests: still `tsc` exit 0, now `Tests: 11 failed, 14 passed`. Eight non-`Ens` in `addressResolvers`, three non-`ens` in the map, which is exactly the set that should notice.

The literal assertions on `Ens` stay alongside the tables, so the tag value is still pinned and not only derived.

wa0x6e (approved):
utAck

PR #538, opened 2026-08-16
refactor(resolvers): split the resize flag from the failure contract

### LINKED ISSUE #537: refactor(resolvers): split the resize flag from the failure contract
A proposal, not a decision. #511 describes this series; this is the step that is next up in it.

## What is wrong

`src/resolvers/index.ts` gives each entry one `resize` flag, and the map build reads it as the answer to two different questions:

```ts
entry.resize ? withResize(entry.name, entry.fn) : entry.fn
```

`resize: false` means "keep the upstream aspect ratio", and it also means "get no wrapper at all". Six of the seventeen entries are bare because of it, and only two of those six want to be. `user-cover`, `space-cover`, `space-logo` and `space-cover-sx` sit outside the failure contract for a reason that has nothing to do with failure: their bytes must not be center-cropped, because `resize()` passes no fit option and a banner cached square can never be served wide again.

## The change

The registry carries the two answers separately and the map composes them: resize the entry if it asks to be resized, then put it under the contract if it belongs there. Those four entries come under the contract while staying unresized, which is the combination one flag cannot express.

`blockie` and `jazzicon` stay outside it. `src/api.ts` hands the fallback resolver's result straight to `resize()`, so a fallback answering `false` makes sharp throw on a path with no guard of its own (#495). `test/unit/resolvers/failureContract.test.ts` already pins that and does not move.

## What it unblocks

The two remaining factory catches, in `createPropertyResolver` in `src/resolvers/snapshot.ts` and in `src/resolvers/space-sx.ts`. #529 had to keep them, because those factories serve the bare entries as well as the wrapped ones and taking the catch out would have let the bare ones throw into the image route in the same edit. Once every entry they produce has a wrapper, both come out, and the catch move is finished.

Two cases in `test/unit/graphQlEnvelope.test.ts` import those factories directly and assert they answer `false` on a flagged envelope, so they move to asserting the rejection, the way the `lens` case did in #529. What the file is actually about, that a payload the upstream flagged is never fetched, does not change.

tony8713:
`resize` was answering two unrelated questions. Whether the base cache keeps the upstream aspect ratio, and whether the entry gets a wrapper at all. Covers and logos want the first and not the second, and one flag cannot say that, so they sit outside the failure contract for a reason that has nothing to do with failure.

The registry carries the two answers separately now and the map composes them: resize the entry if it asks to be resized, then put it under the contract if it belongs there. `user-cover`, `space-cover`, `space-logo` and `space-cover-sx` come under the contract while staying unresized.

## The factory catches go with it

`createPropertyResolver` in `snapshot.ts` and in `space-sx.ts` kept theirs in #529 because they serve bare entries as well as wrapped ones: taking the catch out would have taken it away from the bare ones in the same edit and let them throw into the image route, which has no guard of its own (#495). Every entry those factories produce now has a wrapper to fall back on, so both come out. That is the last of the catch move.

Nothing changes for a caller. `src/resolvers/index.ts` is the only production consumer of these modules, so every production call already goes through the map: before, the factory caught and answered `false`; now the wrapper catches and answers `false`. No new Sentry volume either, because the contract wrapper is as silent as the catches it replaces.

## blockie and jazzicon stay outside the contract

`api.ts` hands the fallback resolver's result straight to `resize()`, and guards neither call, so a fallback answering `false` and a fallback that throws both end as an unhandled rejection out of the route. The exemption does not keep the route alive, it surfaces the resolver's own error rather than sharp's. It is the one combination the split deliberately does not enable, and the case pinning it in `test/unit/resolvers/failureContract.test.ts` is untouched.

## Tests

`test/unit/graphQlEnvelope.test.ts` imports the two factories directly and asserted they answer `false` on a flagged envelope, which is the contract this PR moves out of them. Those cases now assert the rejection, the way the `lens` case did in #529. What the file is actually about, that a payload the upstream flagged is never fetched, is unchanged throughout.

The failure contract file picks up the entries that just joined it: that they answer `false` when the factory throws, and that they still come back at the size the upstream sent. The second is what goes red if the two flags are ever folded back into one.

The registry gets an assertion of its own, because the flag now decides on its own whether an entry can throw out of the map, where `resize: true` used to be a total guard. A wrong value compiles and passes and only shows up as a request that never answers, so the rule is written down: every resolver a chain in `constants.json` can reach is under the contract, and the only entries outside it are the two fallbacks.

## Left for later

Guarding the fallback call in `api.ts` rather than exempting two resolvers from the contract is the deeper fix, and it is the reason the exemption exists at all. It changes behaviour, since a broken fallback would answer a 500 instead of hanging, so it is not folded in here.

Closes #537. Refs #511.

### DIFF
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -55,13 +55,7 @@ function withResize(name: string, resolve: ResolverFn): ResolverFn {
   };
 }
 
-// Covers and logos take resize: false so the base cache keeps its upstream
-// aspect ratio: resize() passes no fit option, so sharp center-crops, and a
-// banner cached as a square can never be served wide again (api.ts resizes
-// every response to the requested w x h regardless).
-// blockie and jazzicon take resize: false because they already call resize()
-// themselves, and failureContract: false because api.ts hands the fallback's
-// result straight to resize(), which throws on a false.
+// blockie and jazzicon resize themselves.
 export const RESOLVERS = [
   { name: 'blockie', fn: blockie, resize: false, failureContract: false },
   { name: 'jazzicon', fn: jazzicon, resize: false, failureContract: false },
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -26,19 +26,23 @@ type Resolver = {
   name: string;
   fn: ResolverFn;
   resize: boolean;
+  failureContract: boolean;
 };
 
-function withResize(name: string, resolve: ResolverFn): ResolverFn {
+function withFailureContract(resolve: ResolverFn): ResolverFn {
   return async (...args) => {
-    let input: Buffer | false;
-
     try {
-      input = await resolve(...args);
+      return await resolve(...args);
     } catch {
       // Silent on purpose: capturing here reports every routine upstream 404.
       return false;
     }
+  };
+}
 
+function withResize(name: string, resolve: ResolverFn): ResolverFn {
+  return async (...args) => {
+    const input = await resolve(...args);
     if (!input) return false;
 
     try {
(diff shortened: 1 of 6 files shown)


src/resolvers/index.ts
```diff
@@ -55,27 +59,27 @@ function withResize(name: string, resolve: ResolverFn): ResolverFn {
...
-  { name: 'space-logo', fn: sResolveSpaceLogo, resize: false },
-  { name: 'space-sx', fn: sxResolveAvatar, resize: true },
-  { name: 'space-cover-sx', fn: sxResolveCover, resize: false },
-  { name: 'selfid', fn: selfid, resize: true },
-  { name: 'lens', fn: lens, resize: true },
-  { name: 'starknet', fn: starknet, resize: true },
-  { name: 'farcaster', fn: farcaster, resize: true }
+// blockie and jazzicon take resize: false because they already call resize()
+// themselves, and failureContract: false because api.ts hands the fallback's
+// result straight to resize(), which throws on a false.
```
wa0x6e: The rationale here doesn't distinguish the two outcomes it's meant to: `api.ts` has no guard on the fallback call *or* the `resize()` after it, so a fallback that throws and a fallback answering `false` (→ `resize(false, …)` → sharp throws) both end as an unhandled rejection out of the route (process crash without a global handler, hung request with Sentry's `warn`-mode one). The exemption buys the resolver's real error over sharp's opaque one — worth having, but not route survival.

```suggestion
// themselves, and failureContract: false because api.ts has no guard on
// either fallback path: a thrown error and a `false` fed to resize() both
// end as an unhandled rejection out of the route. The exemption only buys
// the resolver's real error over sharp's opaque one.
```

test/unit/resolvers/failureContract.test.ts
```diff
@@ -48,6 +74,20 @@ describe('resolvers - failure contract', () => {
...
+  it.each(UNRESIZED)('serves %s at the size the upstream sent', async (name, fn) => {
+    const bytes = Buffer.from('not an image');
+    (fn as jest.Mock).mockResolvedValue(bytes);
+
+    await expect(resolvers[name](ADDRESS)).resolves.toBe(bytes);
+    expect(capture).not.toHaveBeenCalled();
+  });
+
   // A fallback answering false makes sharp throw on the image route, which has
   // no guard of its own.
```
wa0x6e: Same framing as the registry comment — a `false` here would reach the same unguarded `resize()` call in `api.ts` that a throw already reaches; the difference is the error text, not whether the route survives.

```suggestion
  // A fallback answering false would make resize() throw on the same
  // unguarded api.ts path that a thrown error already reaches; letting it
  // throw here only keeps the resolver's own error message.
```

src/resolvers/index.ts
tony8713: Right, and it was worse than imprecise. I checked both halves on this branch: `sharp(false)` throws `Unsupported input 'false' of type boolean`, and express 4.19.2 does not forward an async handler rejection to `fallbackLogger`, so the request hangs with no response either way. A fallback that throws and a fallback answering `false` end identically. The exemption buys the resolver's error instead of sharp's, not route survival.

I removed the block rather than reword it. Both flag values it explained have a named test behind them (`leaves the fallback resolvers outside it` in `resolvers.test.ts`, `serves %s at the size the upstream sent` in `failureContract.test.ts`), so it was rationale a reader can get from a red test. The one line I kept is the one nothing pins: normalising `blockie` or `jazzicon` to `resize: true` double-resizes and no test notices.

test/unit/resolvers/failureContract.test.ts
tony8713: Same answer, removed rather than reworded. The test name already says what it pins, and the rationale was the part doing the misleading.

tony8713:
On the comment rule, the honest answer is that this branch broke it. The registry block in `src/resolvers/index.ts` and the two new ones in the tests were all rationale, and the half you flagged was rationale that did not survive contact with `api.ts`.

Removed: the whole registry block, both new comments in `test/unit/resolvers.test.ts`, the `AggregateError` note in `test/unit/graphQlEnvelope.test.ts`, and the fallback note in `test/unit/resolvers/failureContract.test.ts`. Each of them explained a flag value or an assertion that already has a named test behind it, so a wrong edit goes red rather than needing a reader to have read a paragraph.

Kept one line, `blockie and jazzicon resize themselves`. It is the only value in the table nothing pins: set both to `resize: true` and the resolver suite stays green, image snapshots included, while every fallback response gets encoded twice.

Two of the removed ones were pre-existing rather than mine, so flagging them rather than dropping them quietly: the covers rationale in the registry block, which is accurate but still rationale, and the fallback note you flagged. Say the word if you want either back.

Comments only, nothing else moved: `tsc --removeComments` emits identical JS on both sides.

wa0x6e (approved):
utAck

PR #540, opened 2026-08-16
refactor(resolvers): normalize each resolver's no-data answer, and harden isSilencedError

### LINKED ISSUE #539: refactor(resolvers): normalize each resolver's no-data answer, and harden isSilencedError
A proposal, not a decision. #511 describes this series; this is the step that is next up in it, and it follows #537.

## What is inconsistent today

Every resolver answers `false` for "this input has no image", but only some of them do it by returning `false`. The rest throw, and the map's failure contract turns the throw into the same `false`. Measured on the #537 branch by calling each resolver module directly, rather than through the map:

| input | today |
|---|---|
| `starknet` with any EVM address | throws `Invalid starknet address` |
| `space-sx` for a space on neither API | throws `AggregateError`, `errors: [false, false]` |
| `space-sx` and `snapshot` onchain, id that is not an address | throws ethers `INVALID_ARGUMENT` |
| `space-logo` on an onchain network | throws axios `status: 400` |
| `coingecko` for a coin with no image | throws, from `fetchHttpImage(undefined)` |
| `farcaster` for an address with no account | returns `false`, and so does a 402 from a missing key |

The first row is the one that matters most. `starknet` sits on the `avatar` chain (`src/constants.json:7`) and `normalizeAddress` (`src/resolvers/starknet.ts:10`) throws for anything that is not 64 hex characters, so every ordinary EVM avatar request produces a throw on its way to a normal answer.

## The contract

A resolver returns `false` when the upstream has no image for the input, and throws only when the upstream failed. No response moves: the map already turns both into `false`.

- `starknet`: `normalizeAddress` returns null rather than throwing.
- `space-sx`: `getSpaceProperty` returns null for "not on this API" instead of `Promise.reject(false)`, and the fan-out becomes `allSettled`, rethrowing only when every API failed. Today a space that does not exist and both APIs being down are the same `AggregateError`.
- `space-sx` and `snapshot`: one shared helper for the onchain id list, answering no-data for an id that is not an address instead of letting `getAddress` throw.
- `space-logo`: do not issue the onchain query at all. `SpaceMetadataItem` carries no `logo` field, so the request is a guaranteed validation error rather than a miss.
- `coingecko`: guard the missing image, treat the upstream 404 as no-data, and carry the status on anything else in the shape `graphQlEnvelopeError` already uses.
- `farcaster`: treat the 404 as no-data and stop swallowing every other non-2xx.

`ens` and `trustwallet` keep throwing on their routine 404. Silencing that is a muting decision and belongs to the capture step.

## Why it blocks capture

Capture goes inside the map's failure contract, which is the only place that sees any of these throws. Every row above would page on normal traffic, and the first would page on every EVM avatar request.

`isSilencedError` is hardened in the same change because it is the classifier capture will consult, and its first statement reads `error.name` off the raw caught value (`src/addressResolvers/utils.ts:76`). A rejection carrying null or undefined makes the classifier throw from inside a catch block, on a route with no guard of its own (#495).

## Not in scope

`getAddress` also throws for a malformed token id, in `trustwallet` and in `coingecko`, before any upstream call. That is invalid input rather than no-data and wants an input-validation answer rather than a `false`.

tony8713:
Depends on #538.

Every resolver answers `false` for "this input has no image", but only some of them do it by returning `false`. The rest throw, and the map's failure contract turns the throw into the same `false`, so from the outside the two are the same event. `starknet` is the sharpest case: it sits on the `avatar` chain and throws for anything that is not 64 hex characters, so every ordinary EVM avatar request produces a throw on its way to a normal answer.

That is harmless while nothing looks at the throws. It stops being harmless the moment capture goes inside the failure contract, which is the next step in #511, because normal traffic would page.

So the no-data answer becomes a return rather than a throw, resolver by resolver: the starknet address check, the space-sx fan-out, the onchain space id list that `snapshot.ts` and `space-sx.ts` each built for themselves, the onchain logo query that asks for a field the API does not have, and the coingecko and farcaster responses. A resolver throws only when the upstream actually failed. No response moves, because the map already turned both into `false`.

## The space-sx fan-out

The one with a real decision in it. `getSpaceProperty` used `Promise.reject(false)` to mean "not on this API" and the two were aggregated with `Promise.any`, which made a space that is on neither API and both APIs being down the same `AggregateError`. It fans out with `allSettled` now and rethrows only when every API failed, so one API being down while the other answers "no such space" stays no-data. That two-API resilience is why both are asked at all.

The cost is that both are now always waited for, where `Promise.any` returned on the first hit. Both endpoints answer a miss about as fast as they answer a hit, and this sits behind the base cache, so ordinarily that is the difference between the slower and the faster of two calls. Where it costs something real is a degraded endpoint: the wait is then bounded by the axios timeout already on the call, and the old code could have returned on the other endpoint's hit. Telling a space that does not exist from both APIs being down is worth that.

## isSilencedError

Hardened in the same change because it is the classifier capture will consult, and it reads properties off the raw caught value as its first statement. A rejection carrying null makes the classifier itself throw, from inside a catch block, on a route with no guard of its own (#495).

It silences such a rejection rather than reporting it, because reporting is not available: `capture` reads `.error` off whatever it is handed and throws on the same values, so classifying one as worth reporting only moves the throw one line down. There is nothing in a bare null to report anyway.

## Left alone

`ens` and `trustwallet` still throw on their routine upstream 404. Silencing that is a muting decision and belongs with capture rather than here.

`getAddress` also throws for a malformed token id, in `trustwallet` and in `coingecko`, before any upstream call. That is invalid input rather than no-data, and a `false` is the wrong answer to it.

Closes #539. Refs #511.

### DIFF
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
@@ -1,4 +1,9 @@
 export function isSilencedError(error: any, additionalMessages?: string[]): boolean {
+  // A rejection carries whatever it was given, null included. There is nothing
+  // in one to classify, and reporting it is not an option either: `capture`
+  // dereferences it and throws, from inside the catch block that called this.
+  if (!error) return true;
+
   // An abort is always one of our own deadlines, and each transport words it differently.
   if (error.name === 'AbortError') return true;
 
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
@@ -1,12 +1,27 @@
 import http from 'http';
 import https from 'https';
+import { getAddress } from '@ethersproject/address';
 import axios from 'axios';
+import { isStarknetAddress } from './address';
 
 export const axiosDefaultParams = {
   httpAgent: new http.Agent({ keepAlive: true }),
   httpsAgent: new https.Agent({ keepAlive: true }),
   timeout: 5e3
 };
+
+// Spaces are keyed by address on the onchain APIs, and both accept the raw id
+// as well as the checksummed one. An id that is not an address is not a space
+// there, so it is no-data rather than something to ask about.
+export function spaceIds(id: string): string[] | null {
+  if (isStarknetAddress(id)) return [id];
+
+  try {
+    return [id, getAddress(id)];
+  } catch {
+    return null;
+  }
+}
 
 export async function fetchHttpImage(url: string): Promise<Buffer> {
   return (
--- a/src/resolvers/image/coingecko.ts
+++ b/src/resolvers/image/coingecko.ts
@@ -1,5 +1,6 @@
 import { getAddress } from '@ethersproject/address';
 import { fetchHttpImage } from '../../helpers/http';
+import { httpError } from '../../utils';
 
 const API_KEY = process.env.COINGECKO_API_KEY;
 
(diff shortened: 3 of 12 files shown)


wa0x6e (approved):
utAck

wa0x6e (approved):
utAck

Issue #536, opened 2026-08-16
Resolver fan-outs set no request deadline, so a hung upstream stalls the whole request


tony8713:
This is a report of measured behaviour, taken on 2026-08-16 on Node v22.23.1 against a socket that accepts the connection and then never responds. Every outbound call in `src/` was redirected to it, so the only thing that could end a call was a deadline the caller had set.

Four call sites have none. Three of them are on a request path.

## The shape

`src/api.ts:137` fans out over the resolver chain with `Promise.all`:

```js
const files = await Promise.all(
  currentResolvers.map(r => resolvers[r](address, network, networkId))
);
```

`withResize` in `src/resolvers/index.ts:41` turns a resolver throw into `false`, and its own comment on line 39 says it does that because the image route "has no guard of its own (#495)". It does nothing for a promise that never settles, because there is no catch for a promise that never rejects. `src/index.ts:34` is a bare `app.listen(PORT)` with no `server.setTimeout` and no timeout middleware, so nothing downstream ends the request either.

`src/addressResolvers/index.ts:59` has the same shape for `lookup_addresses` and `resolve_names`: `Promise.all` over the resolvers, with a catch that only covers a throw.

## The four sites

Time to settle against the hanging socket:

| site | client | deadline | observed |
|---|---|---|---|
| `src/resolvers/farcaster.ts:30` | node-fetch v2 | none | still pending at 130s |
| `src/resolvers/coingecko.ts:25` | global fetch | none | still pending at 130s |
| `src/addressResolvers/shibarium.ts:31,49` | `DNSConnect` | none | still pending at 130s |
| `src/getOwner/shibarium.ts:46` | `DNSConnect` | none | still pending at 130s |

`farcaster` is the worst of the four because it has no backstop of any kind. node-fetch v2 defaults `timeout` to `0` (`node-fetch/lib/index.js:188`) and only arms a timer `if (request.timeout)` (line 1491), and Node sets no default socket timeout, so the call hangs until the OS gives up on the connection. It sits on the `avatar` chain (`src/constants.json:7`), which is the common image path.

`coingecko` is a bare global `fetch` with no signal, reached before the bounded `fetchHttpImage` later in the same function. Its only backstop is undici's own headers timeout, measured at 301s on the same box. It sits on the `token` chain (`src/constants.json:9`).

[...]

PR #541, opened 2026-08-16
fix(resolvers): bound the farcaster and coingecko calls with a deadline

tony8713:
Two of the four call sites listed in #536.

Neither `farcaster` nor `coingecko` sets a deadline of any kind, so an upstream that accepts the connection and then never answers holds the resolver open. `src/api.ts` fans the resolver chain out with `Promise.all`, and `withResize` in `src/resolvers/index.ts` turns a resolver throw into `false` but has nothing to catch when the promise simply never settles, so the whole image request waits behind it. `farcaster` is the worse of the two: node-fetch v2 arms no timer unless one is asked for, so there is no backstop at all, where `coingecko` at least has undici's.

Both move onto the `withDeadline` helper added in #534, rather than a per-transport option. Neither transport's own option does the job here. node-fetch's `timeout` raises a `FetchError`, so a deadline this code set for itself would not be recognised as the abort it is. The global `fetch` has no `timeout` option at all and ignores the key silently, which typechecks and does nothing.

The deadline wraps the whole call rather than the request. Both transports resolve as soon as the response headers arrive, so a body that then stalls falls outside a budget that ends at the request.

`farcaster` already discarded its own errors, so nothing about its result changes and a deadline only makes it end. `coingecko` raises the abort onto the same path any other failure there takes.

### DIFF
--- a/src/resolvers/image/farcaster.ts
+++ b/src/resolvers/image/farcaster.ts
@@ -2,7 +2,7 @@ import { getAddress } from '@ethersproject/address';
 import fetch from 'node-fetch';
 import { max } from '../../constants.json';
 import { fetchHttpImage } from '../../helpers/http';
-import { Address, httpError } from '../../utils';
+import { Address, httpError, withDeadline } from '../../utils';
 
 const NEYNAR_API_URL = 'https://api.neynar.com/v2/farcaster/user/bulk-by-address';
 const API_KEY = process.env.NEYNAR_API_KEY ?? '';
@@ -25,19 +25,25 @@ function normalizeAddress(address: Address): Address | null {
   }
 }
 
+// The body read is inside the deadline, not just the request: node-fetch hands
+// back a response as soon as the headers land, so a budget ending at the request
+// would leave a stalled body unbounded.
 async function fetchAddressImageUrl(normalizedAddress: string): Promise<string | null> {
-  const response = await fetch(`${NEYNAR_API_URL}?addresses=${normalizedAddress}`, {
-    headers: { Accept: 'application/json', api_key: API_KEY }
+  return withDeadline(async signal => {
+    const response = await fetch(`${NEYNAR_API_URL}?addresses=${normalizedAddress}`, {
+      headers: { Accept: 'application/json', api_key: API_KEY },
+      signal
+    });
+
+    // Neynar answers 404 for an address with no Farcaster account, which is the
+    // routine no-avatar case rather than a failure.
+    if (response.status === 404) return null;
+    if (!response.ok) throw httpError('farcaster', response.status, response.statusText);
+
+    const data: Record<Address, UserDetails[]> = await response.json();
+
+    return data[normalizedAddress.toLowerCase()]?.[0]?.pfp_url ?? null;
   });
-
-  // Neynar answers 404 for an address with no Farcaster account, which is the
-  // routine no-avatar case rather than a failure.
-  if (response.status === 404) return null;
-  if (!response.ok) throw httpError('farcaster', response.status, response.statusText);
-
-  const data: Record<Address, UserDetails[]> = await response.json();
-
-  return data[normalizedAddress.toLowerCase()]?.[0]?.pfp_url ?? null;
 }
 
 export default async function resolve(address: string): Promise<Buffer | false> {
--- a/src/resolvers/image/coingecko.ts
+++ b/src/resolvers/image/coingecko.ts
@@ -1,6 +1,6 @@
 import { getAddress } from '@ethersproject/address';
 import { fetchHttpImage } from '../../helpers/http';
-import { httpError } from '../../utils';
+import { httpError, withDeadline } from '../../utils';
 
 const API_KEY = process.env.COINGECKO_API_KEY;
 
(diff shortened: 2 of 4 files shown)


src/resolvers/image/farcaster.ts
```diff
@@ -27,17 +27,20 @@ function normalizeAddress(address: Address): Address | null {
...
+        throw new Error(`Invalid network response (${response.url} ${response.status})`);
+      }
 
-    const data: Record<Address, UserDetails[]> = await response.json();
+      const data: Record<Address, UserDetails[]> = await response.json();
 
-    return data[normalizedAddress.toLowerCase()]?.[0].pfp_url;
+      return data[normalizedAddress.toLowerCase()]?.[0].pfp_url;
+    });
   } catch {
```
wa0x6e: A deadline hit is invisible here: this `catch` discards the `AbortError` (and on the coingecko side `withResize`'s "silent on purpose" catch does the same), and `src/helpers/metrics.ts` has no per-provider metric for the image resolvers — so a Neynar/CoinGecko stall that trips the 10s budget is indistinguishable from "no avatar", and there is no way to tell afterwards whether 10s was the right number. Pre-existing blind spot, but this is the first bounded, latency-dependent failure routed into it. `lookupDomains/index.ts` already records aborts in a per-provider histogram (`timeResponse`), silent to Sentry but visible in Prometheus — the image chain has no equivalent. Route-level `/avatar/#id` p99 is ~9.6s over the last 7 days, so traffic sits close to the boundary. Fine as a follow-up, just flagging.

tony8713: Confirmed, and one part of it is worse than it looks: `isSilencedError` never sees this path, so it is not what mutes the abort.

I ran the production resolver map against an upstream that accepts and never answers, with `capture` patched to count calls rather than trusting the `NODE_ENV=test` early return. As things stand both resolvers return `false` at ~10s and `capture` is called 0 times, and removing farcaster's `catch` changes nothing observable, because `withFailureContract` swallows it identically.

The two only diverge once the image chain starts capturing. With that one catch changed to report, farcaster still records nothing while coingecko produces an unmuted `AbortError`: nothing under `src/resolvers/image` imports `isSilencedError`, it lives on the address side. So whoever enables capture there has to add that check or every deadline turns into a Sentry event.

Agreed it is a follow-up, not this PR.

tony8713:
You were right, and I checked it before changing anything rather than taking it on trust.

Worth recording how the check nearly went wrong: my first attempt at your mutation did not typecheck, because node-fetch is untyped here so `withDeadline` infers `unknown`, and the tests went red on a compile error in `beforeAll` with empty failure messages. That red looked like discrimination and was not. With a version that compiles, all the old cases stayed green, exactly as you said.

There were two causes, not one. The server never sent headers, and the harness fired the deadline's timer callback by hand, which aborts the signal even where the code under test had already cleared it. Fixing only the first left the tests green under the same mutation.

The cases now cover both shapes, and the deadline is waited out instead of fired. Under the body mutation the two header cases stay green and the two body cases fail on timeout, which is the contrast that was missing; across the related set only this file goes red and the other 15 suites stay green. Removing `signal` outright fails all four.

On the rebase: master rather than #540. #540 is behind master and currently CONFLICTING, and master has since taken the resolver move, so master is the base that applies. The resolvers are at `src/resolvers/image/` and the cases at `test/integration/resolvers/image/deadline.test.ts`.

wa0x6e (approved):
utAck

src/resolvers/image/farcaster.ts
tony8713: Update now that #540 has landed: the half you pointed at directly is gone. farcaster no longer has that `catch`, so both resolvers hand the abort to the map rather than one of them swallowing its own.

The observable behaviour has not changed, though. Re-running the same probe on the rebased head, both still return `false` at ~10s with `capture()` called 0 times, because `withFailureContract`'s silent catch is now what mutes it, for both rather than just one. And `src/helpers/metrics.ts` still carries only `address_resolver_response_duration_seconds` and `lookup_domains_response_duration_seconds`, nothing the image chain can land an abort in.

So the follow-up is still worth keeping, just narrower than when you wrote it: it is no longer an asymmetry between the two resolvers, it is that the image chain's failure contract is silent and has no per-provider metric.

tony8713:
Rebased onto master now that #540 has merged. Head `b4eeb02`, conflict resolved, CI green.

Both resolver files conflicted for real rather than textually, and I took #540's shape: the status checks and the `httpError` throws are unchanged, and the `signal` plus the body read now sit inside `withDeadline`. farcaster's expectation is reconciled the other way from before, since #540 removed the `catch` that used to turn its errors into a `null`, so it raises the abort exactly as coingecko does and the map's `withFailureContract` is what turns that into a `false`.

One thing worth flagging, because the obvious smaller resolution is wrong. #540 reads the body in a statement after the status checks, so wrapping only the `fetch` would be a one-line change per site but would leave `response.json()` outside the budget. That is precisely the mutation these cases exist to catch: applying it leaves the two header cases green at ~10s and hangs the two body cases until the 30s test timeout. Keeping the whole call inside the deadline means indenting #540's checks into the callback, so the source diff is a little larger than master rather than smaller. Removing `signal` outright still fails all four.

PR #542, opened 2026-08-16
fix(addressResolvers): bound the shibarium DNS fan-out

tony8713:
`lookup_addresses` and `resolve_names` fan out over `DNSConnect` with `Promise.all`, and nothing bounds that call. The per-provider catch in `addressResolvers/index.ts` covers a provider that throws rather than one that never settles, so a DNS-over-HTTPS request that is accepted and never answered holds the whole call open, with only undici's own headers timeout underneath it.

`dnsConnect.resolve()` and `reverseResolve()` take no signal, and `DNSConnectOptions` exposes none, so this bounds the wait rather than the request, the way #534 did for `getOwner`. Cancelling the request instead would mean supplying our own `DNSResolver`, and with it a copy of dns-connect's NXDomain handling and of its DNSSEC `AD` check, which is what stops a resolution downgrade from passing as an answer.

The budget wraps the whole fan-out rather than a single query, because each entry issues several DNS queries in sequence and `lookup_addresses` accepts up to 50 addresses. A per-query bound would put no ceiling on the call, for the same reason the deadline in `lookupDomains` sits outside the paging loop rather than inside it.

The abort surfaces as an `AbortError`, which `isSilencedError` already matches, so a deadline reports nothing to Sentry.

One thing worth your judgement: the budget is all or nothing over the batch, so if it trips every address in that batch comes back unresolved, and `addressResolvers/index.ts` then caches an unresolved address exactly as it does after any other provider failure. `lookup_addresses` accepts 50 addresses, and a cold connection pool makes a batch that size the case that sits closest to the shared budget.

Part of #536.

### DIFF
--- a/src/addressResolvers/shibarium.ts
+++ b/src/addressResolvers/shibarium.ts
@@ -1,5 +1,5 @@
 import { DNSConnect } from '@webinterop/dns-connect';
-import { Address, Handle } from '../utils';
+import { Address, Handle, untilAborted, withDeadline } from '../utils';
 import { isEvmAddress, withoutEmptyValues } from './utils';
 import constants from '../constants.json';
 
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

Issue #515, opened 2026-08-14
COINGECKO_API_KEY is missing from .env.example and the CI workflow


tony8713:
`src/resolvers/coingecko.ts` reads `COINGECKO_API_KEY` and returns false immediately when it is unset. The key is in neither `.env.example` nor the `env:` block of `.github/workflows/test.yml`, so in CI the coingecko resolver never makes a request. It returns false for every token, and its integration test reports itself as todo rather than failing, which is why nothing has flagged it.

That is worth fixing on its own, but #514 makes it sharper: with zapper gone the token chain is trustwallet and coingecko, so in CI it is effectively trustwallet alone. Any regression in coingecko resolution passes CI unnoticed.

Two separate things to decide: whether to add the secret to the test workflow so the resolver is actually exercised, and whether to list the key in `.env.example` so it is discoverable for local setup. The second is worth doing either way.

PR #551, opened 2026-08-16
feat(resolvers): report resolver failures

### LINKED ISSUE #543: feat(resolvers): report resolver failures, and record each resolver in a metric
A proposal, not a decision. #511 describes this series; this is the last step in it, and it follows #539.

## What is silent today

`src/resolvers/index.ts` turns every resolver failure into `false`. A resolver dying and an address simply having no avatar are the same event: blockie fallback, nothing reported. `src/resolvers` also has no metric at all, where `addressResolvers` and `lookupDomains` each have one.

The comment on the failure contract says why it is silent: capturing there reports every routine upstream 404. #539 is what makes changing it safe, by turning each resolver's no-data answer into a clean `false`, so what reaches the catch is a failure rather than a miss.

## What is still routine after #539

Measured by calling the resolver map against the real upstreams, on #540's branch, over the `avatar` and `token` chains:

| path | what arrives in the catch |
|---|---|
| `snapshot` and `user-cover`, for any id that is not an address | the hub answers `user(id:)` with `Invalid address` in the GraphQL envelope, HTTP 200. The avatar route serves names, so this is every ENS-name request |
| `ens`, for a name with a resolver and no avatar record | `metadata.ens.domains` 404 |
| `trustwallet`, for a token that is not in the assets repo | `raw.githubusercontent.com` 404 |
| `trustwallet` and `coingecko`, for a malformed id | ethers `INVALID_ARGUMENT`. `parseQuery` validates nothing, so `/token/<anything>` reaches `getAddress` |

The first row is a no-data path #539 did not cover, and it wants the treatment #539 gave the rows it did cover: do not ask, because the answer is a guaranteed validation error rather than a miss.

The other three are the muting decision #539 deferred to this step. In this family a 404 is the image not being there: every URL a resolver fetches is either a third party's or a template over an id from the request. And an ethers argument error is the request carrying something that is not an address, which the route never checked.

Both are policy for this family rather than for every caller, so they belong beside the failure contract rather than inside `isSilencedError`.

## What stays loud

`farcaster`'s 402 and `coingecko`'s 401, among everything else. A degraded key is exactly what should be visible, and it is the case the metric alone would not distinguish from a quiet day.

## The metric, deferred

**Update, 2026-08-16.** Wan's call is that the metric waits for more refactoring of the registry, so this step ships as capture only and the section below stays as the spec for when it comes back. It was written, reviewed and then reverted in its own commit on #551, so bringing it back is a `git revert` rather than a rewrite.

What that costs while it is deferred: a failure the capture layer mutes produces no signal anywhere, in Sentry by design and in a metric because there is none. The sharp case is a 404 from one of the GraphQL endpoints on a route with a single resolver in its chain, where a dead upstream serves fallbacks silently. Nothing regresses against today, since none of it was visible before either.

`lookup_domains_response_duration_seconds` from #532 is the shape: a histogram labelled by provider, with the status recorded in a `finally` so a failure counts whether or not it was reported. Labels are `provider` and `status` only. The resolvers' other arguments come out of the URL path, so their cardinality is unbounded and they cannot be labels.

Muting without a metric would be a straight loss, which is #532's reasoning.

No per-resolver mute list, which is what #511 sketched. The policy above does not vary per resolver, and a registry field carrying the same value on every entry is worse than no field. That is where the same question landed for `lookupDomains`.

tony8713:
Depends on #540.

`src/resolvers/index.ts` turns every resolver failure into `false`, so a resolver dying and an address simply having no avatar are the same event: blockie fallback, nothing reported. The comment on the failure contract says why it had to stay that way, and #540 is what removes the reason, by making each resolver's no-data answer a clean `false` so what reaches the catch is a failure rather than a miss.

The catch reports now, through the same classifier and in the same context shape `addressResolvers` and `lookupDomains` already use.

## What is still routine, and stays silent

Three things reach the catch on ordinary traffic even after #540. Found by calling the map against the real upstreams over the `avatar` and `token` chains rather than by reading it.

- **An upstream 404.** `ens` gets one from `metadata.ens.domains` for a name with no avatar record, and `trustwallet` gets one from the assets repo for any token that is not in it. In this family a 404 is the image not being there: every URL a resolver fetches is either a third party's or a template over an id from the request. One of the GraphQL endpoints answering 404, which is what a `HUB_URL` with a trailing slash produces, is muted along with them. That costs most on `/space`, `/space-cover`, `/space-logo`, `/user-cover` and `/space-sx`, which have a single resolver each, so a dead upstream there is a route that quietly serves nothing but fallbacks. See the last section.
- **An ethers argument error.** `parseQuery` validates nothing, so `/token/<anything>` reaches `getAddress`. That is the request carrying something that is not an address.
- **Whatever `isSilencedError` already silences**, which is why it is consulted rather than replaced.

The first two are policy for this family rather than for every caller, so they sit beside the failure contract rather than in the shared classifier. There is no per-resolver mute list, which is what #511 sketched: the policy does not vary per resolver, and a registry field carrying the same value on every entry is worse than no field. That is where the same question landed for `lookupDomains`.

A degraded key stays loud. `farcaster`'s 402 and `coingecko`'s 401 are exactly the case worth seeing. `coingecko` short-circuits on a missing key where `farcaster` does not, and giving `farcaster` the same guard is not an improvement: it is #515, where a resolver that is switched off looks exactly like one that has nothing to say.

## The no-data paths, closed rather than muted

Running the map against the real upstreams turned up paths that throw on ordinary traffic and are not failures at all. Each gets the treatment #540 gave the same shape elsewhere, which is to answer no-data rather than to mute a report. No response moves.

- The hub answers `user(id:)` with `Invalid address` in the GraphQL envelope, under HTTP 200, for anything that is not an address. The avatar route serves names, so `snapshot` and `user-cover` threw on every ENS-name request. `space(id:)` is unaffected, it takes a slug and answers null for one it does not have.
- `/avatar/.lens` reaches the Lens API with an empty localName, which it answers with a validation error rather than with "no such account". Anyone can hold that URL down. The same is true of a localName past the length the API will parse, which it rejects in the envelope under HTTP 200.
- `/avatar/.cb.id`, and any other name with an empty component, makes namehash throw before anything is asked. `isValidName` already knows which names namehash can take.
- A label past DNS's 63 bytes gets past `isValidName`, and `dnsEncode` is what enforces it, on the wildcard path an offchain resolver takes even to read a name with no record. The throw is a plain `Error` from outside ethers' own catch, so it is neither the CCIP answer below nor a routine miss. It is read where it happens rather than guarded before the call, because a name whose own resolver is onchain never reaches `dnsEncode` and resolves today.
- An offchain ENS resolver answers a missing text record with a CCIP failure where an onchain one returns null, so every `cb.id` avatar request threw. Falling through to the metadata endpoint is what a null already did. Only the gateway's own 4xx answer is read that way: an exhausted gateway carries different wording under the same code, so an outage stays visible. Ethers keeps no status inside that range, so a gateway 401 or 429 reads as its 404 does, which is a gap this leaves rather than one it opens.

## One correction beside them

`lens` cut the local name at the first `.lens` rather than the last, so `/avatar/a.lensb.lens` asked the API for account `a`. No registered handle can reach it, since a local name takes no dot, so this is malformed input being answered under a name it did not carry.

## The metric is deferred

An earlier revision of this branch added the per-resolver histogram `src/resolvers` still lacks, in the shape #532 used for `lookupDomains`. It is taken back out, in its own commit, because the registry wants more refactoring first.

That leaves one gap, and it is worth stating rather than discovering later. A failure this change mutes now produces no signal anywhere: not in Sentry, by design, and not in a metric, because there is none. The sharp case is the 404 rule above meeting a route with one resolver in its chain, where a wholly dead upstream serves fallbacks silently. Nothing regresses against today, since none of that was visible before either, but the mute is the one part of this change that would have been safer paired with the metric than on its own.

Closes #543. Refs #511.

### DIFF
--- a/src/resolvers/image/index.ts
+++ b/src/resolvers/image/index.ts
@@ -30,8 +30,6 @@ type Resolver = {
   failureContract: boolean;
 };
 
-// Deleting either rule turns a routine miss into a report: a 404 is the image
-// not being there, and the route hands the resolvers an id it never validated.
 function isRoutineMiss(error: any): boolean {
   return (
     error?.status === 404 || error?.response?.status === 404 || error?.code === 'INVALID_ARGUMENT'
--- a/src/resolvers/image/index.ts
+++ b/src/resolvers/image/index.ts
@@ -18,7 +18,9 @@ import { resolveAvatar as sxResolveAvatar, resolveCover as sxResolveCover } from
 import starknet from './starknet';
 import trustwallet from './trustwallet';
 import { max } from '../../constants.json';
+import { isSilencedError } from '../../helpers/errors';
 import { resize } from '../../helpers/image';
+import { timeImageResolverResponse as timeResponse } from '../../helpers/metrics';
 
 type ResolverFn = (...args: any[]) => Promise<Buffer | false>;
 
@@ -29,13 +31,30 @@ type Resolver = {
   failureContract: boolean;
 };
 
-function withFailureContract(resolve: ResolverFn): ResolverFn {
+// Deleting either rule turns a routine miss into a report: a 404 is the image
+// not being there, and the route hands the resolvers an id it never validated.
+function isRoutineMiss(error: any): boolean {
+  return (
+    error?.status === 404 || error?.response?.status === 404 || error?.code === 'INVALID_ARGUMENT'
+  );
+}
+
+function withFailureContract(name: string, resolve: ResolverFn): ResolverFn {
   return async (...args) => {
+    const end = timeResponse.startTimer({ provider: name });
+    let status = 0;
+
     try {
-      return await resolve(...args);
-    } catch {
-      // Silent on purpose: capturing here reports every routine upstream 404.
+      const image = await resolve(...args);
+      status = 1;
+      return image;
+    } catch (err) {
+      if (!isSilencedError(err) && !isRoutineMiss(err)) {
+        capture(err, { tags: { provider: name }, contexts: { input: { args } } });
+      }
       return false;
+    } finally {
+      end({ status });
     }
   };
 }
@@ -89,6 +108,6 @@ export default Object.fromEntries(
   RESOLVERS.map(entry => {
     const resolve = entry.resize ? withResize(entry.name, entry.fn) : entry.fn;
 
-    return [entry.name, entry.failureContract ? withFailureContract(resolve) : resolve];
+    return [entry.name, entry.failureContract ? withFailureContract(entry.name, resolve) : resolve];
   })
 ) as ResolverMap;
--- a/src/resolvers/image/index.ts
+++ b/src/resolvers/image/index.ts
@@ -20,7 +20,6 @@ import trustwallet from './trustwallet';
 import { max } from '../../constants.json';
 import { isSilencedError } from '../../helpers/errors';
 import { resize } from '../../helpers/image';
-import { timeImageResolverResponse as timeResponse } from '../../helpers/metrics';
 
 type ResolverFn = (...args: any[]) => Promise<Buffer | false>;
 
@@ -41,20 +40,13 @@ function isRoutineMiss(error: any): boolean {
 
 function withFailureContract(name: string, resolve: ResolverFn): ResolverFn {
   return async (...args) => {
-    const end = timeResponse.startTimer({ provider: name });
-    let status = 0;
-
     try {
-      const image = await resolve(...args);
-      status = 1;
-      return image;
+      return await resolve(...args);
     } catch (err) {
       if (!isSilencedError(err) && !isRoutineMiss(err)) {
         capture(err, { tags: { provider: name }, contexts: { input: { args } } });
       }
       return false;
-    } finally {
-      end({ status });
     }
   };
 }
--- a/src/resolvers/image/lens.ts
+++ b/src/resolvers/image/lens.ts
@@ -27,8 +27,6 @@ export default async function resolve(domainOrAddress: string) {
     request = { address: getAddress(domainOrAddress) };
   } else if (domainOrAddress.endsWith(LENS_EXTENSION)) {
     const localName = domainOrAddress.split(LENS_EXTENSION)[0];
-    // The API rejects an empty localName rather than answering that it has no
-    // such account, and `/avatar/.lens` reaches here.
     if (!localName) return false;
 
     request = { username: { localName } };
--- a/src/resolvers/image/lens.ts
+++ b/src/resolvers/image/lens.ts
@@ -26,7 +26,12 @@ export default async function resolve(domainOrAddress: string) {
   if (isAddress(domainOrAddress)) {
     request = { address: getAddress(domainOrAddress) };
   } else if (domainOrAddress.endsWith(LENS_EXTENSION)) {
-    request = { username: { localName: domainOrAddress.split(LENS_EXTENSION)[0] } };
+    const localName = domainOrAddress.split(LENS_EXTENSION)[0];
+    // The API rejects an empty localName rather than answering that it has no
+    // such account, and `/avatar/.lens` reaches here.
+    if (!localName) return false;
+
+    request = { username: { localName } };
   } else {
     return false;
   }
--- a/src/resolvers/image/lens.ts
+++ b/src/resolvers/image/lens.ts
@@ -5,6 +5,7 @@ import { fetchHttpImage } from '../../helpers/http';
 const API_URL = 'https://api.lens.xyz';
 const LENS_IPFS_GATEWAY = 'https://gw.ipfs-lens.dev/ipfs/';
 const LENS_EXTENSION = '.lens';
+const LOCAL_NAME_MAX_BYTES = 254;
 
 function normalizeImageUrl(url: string) {
   if (!url) return false;
@@ -27,7 +28,7 @@ export default async function resolve(domainOrAddress: string) {
     request = { address: getAddress(domainOrAddress) };
   } else if (domainOrAddress.endsWith(LENS_EXTENSION)) {
     const localName = domainOrAddress.split(LENS_EXTENSION)[0];
-    if (!localName) return false;
+    if (!localName || Buffer.byteLength(localName) > LOCAL_NAME_MAX_BYTES) return false;
 
     request = { username: { localName } };
   } else {
--- a/src/resolvers/image/lens.ts
+++ b/src/resolvers/image/lens.ts
@@ -27,7 +27,7 @@ export default async function resolve(domainOrAddress: string) {
   if (isAddress(domainOrAddress)) {
     request = { address: getAddress(domainOrAddress) };
   } else if (domainOrAddress.endsWith(LENS_EXTENSION)) {
-    const localName = domainOrAddress.split(LENS_EXTENSION)[0];
+    const localName = domainOrAddress.slice(0, -LENS_EXTENSION.length);
     if (!localName || Buffer.byteLength(localName) > LOCAL_NAME_MAX_BYTES) return false;
 
     request = { username: { localName } };
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "@self.id/core": "^0.3.0",
     "@snapshot-labs/snapshot-metrics": "^2.0.0",
     "@snapshot-labs/snapshot-sentry": "^3.1.0",
-    "@snapshot-labs/snapshot.js": "^0.16.0",
+    "@snapshot-labs/snapshot.js": "^0.17.0",
     "@unstoppabledomains/resolution": "^9.2.2",
     "@webinterop/dns-connect": "^0.3.1",
     "axios": "^0.25.0",
--- a/src/helpers/metrics.ts
+++ b/src/helpers/metrics.ts
@@ -32,6 +32,12 @@ export const timeLookupDomainsResponse = new client.Histogram({
   labelNames: ['provider', 'chainId', 'status']
 });
 
+export const timeImageResolverResponse = new client.Histogram({
+  name: 'image_resolver_response_duration_seconds',
+  help: "Duration in seconds of each image resolver's response.",
+  labelNames: ['provider', 'status']
+});
+
 export const addressResolversCacheHitCount = new client.Counter({
   name: 'address_resolvers_cache_hit_count',
   help: 'Number of hit/miss of the address resolvers cache layer',
--- a/src/helpers/metrics.ts
+++ b/src/helpers/metrics.ts
@@ -32,12 +32,6 @@ export const timeLookupDomainsResponse = new client.Histogram({
   labelNames: ['provider', 'chainId', 'status']
 });
 
-export const timeImageResolverResponse = new client.Histogram({
-  name: 'image_resolver_response_duration_seconds',
-  help: "Duration in seconds of each image resolver's response.",
-  labelNames: ['provider', 'status']
-});
-
 export const addressResolversCacheHitCount = new client.Counter({
   name: 'address_resolvers_cache_hit_count',
   help: 'Number of hit/miss of the address resolvers cache layer',
--- a/src/helpers/provider.ts
+++ b/src/helpers/provider.ts
@@ -14,7 +14,10 @@ export function getProviderOptions(timeout = DEFAULT_PROVIDER_TIMEOUT): {
   };
 }
 
-export function getProvider(network: string | number, timeout = DEFAULT_PROVIDER_TIMEOUT) {
+export function getProvider<T extends string | number>(
+  network: T,
+  timeout = DEFAULT_PROVIDER_TIMEOUT
+) {
   return snapshot.utils.getProvider(network, getProviderOptions(timeout));
 }
 
--- a/src/resolvers/image/snapshot.ts
+++ b/src/resolvers/image/snapshot.ts
@@ -1,6 +1,7 @@
 import { getAddress } from '@ethersproject/address';
 import snapshot from '@snapshot-labs/snapshot.js';
 import { defaultOffchainNetwork, offchainNetworks } from '../../constants.json';
+import { isEvmAddress, isStarknetAddress } from '../../helpers/address';
 import { graphQlCall } from '../../helpers/graphql';
 import { fetchHttpImage, getUrl, spaceIds } from '../../helpers/http';
 
@@ -50,6 +51,10 @@ async function getOffchainProperty(
   entity: Entity,
   property: Property
 ) {
+  // The hub answers `user(id:)` with a validation error rather than a miss for an
+  // id that is not an address, and the avatar route serves names too.
+  if (entity === 'user' && !isEvmAddress(id) && !isStarknetAddress(id)) return null;
+
   const {
     data: {
       data: { entry }
--- a/src/resolvers/image/snapshot.ts
+++ b/src/resolvers/image/snapshot.ts
@@ -51,8 +51,6 @@ async function getOffchainProperty(
   entity: Entity,
   property: Property
 ) {
-  // The hub answers `user(id:)` with a validation error rather than a miss for an
-  // id that is not an address, and the avatar route serves names too.
   if (entity === 'user' && !isEvmAddress(id) && !isStarknetAddress(id)) return null;
 
   const {
--- a/src/resolvers/image/snapshot.ts
+++ b/src/resolvers/image/snapshot.ts
@@ -1,7 +1,6 @@
 import { getAddress } from '@ethersproject/address';
 import snapshot from '@snapshot-labs/snapshot.js';
 import { defaultOffchainNetwork, offchainNetworks } from '../../constants.json';
-import { isEvmAddress, isStarknetAddress } from '../../helpers/address';
 import { graphQlCall } from '../../helpers/graphql';
 import { fetchHttpImage, getUrl, spaceIds } from '../../helpers/http';
 
@@ -51,8 +50,6 @@ async function getOffchainProperty(
   entity: Entity,
   property: Property
 ) {
-  if (entity === 'user' && !isEvmAddress(id) && !isStarknetAddress(id)) return null;
-
   const {
     data: {
       data: { entry }
--- a/test/unit/resolvers/image/failureContract.test.ts
+++ b/test/unit/resolvers/image/failureContract.test.ts
@@ -90,18 +90,24 @@ describe('resolvers - failure contract', () => {
     await expect(resolvers.ens(ADDRESS)).resolves.toBe(false);
   });
 
-  it('reports a resolver failure, with the resolver named and its arguments', async () => {
-    const error = new Error('boom');
-    (ens as jest.Mock).mockRejectedValue(error);
+  it.each([
+    ['ens', ens],
+    ['lens', lens]
+  ] as const)(
+    'reports a %s failure, with the resolver named and its arguments',
+    async (name, fn) => {
+      const error = new Error('boom');
+      (fn as jest.Mock).mockRejectedValue(error);
 
-    await resolvers.ens(ADDRESS);
+      await resolvers[name](ADDRESS);
 
-    expect(capture).toHaveBeenCalledTimes(1);
-    expect(capture).toHaveBeenCalledWith(error, {
-      tags: { provider: 'ens' },
-      contexts: { input: { args: [ADDRESS] } }
-    });
-  });
+      expect(capture).toHaveBeenCalledTimes(1);
+      expect(capture).toHaveBeenCalledWith(error, {
+        tags: { provider: name },
+        contexts: { input: { args: [ADDRESS] } }
+      });
+    }
+  );
 
   it.each(NOT_REPORTED)('does not report %s', async (_label, error) => {
     (ens as jest.Mock).mockRejectedValue(error);
--- a/test/unit/resolvers/image/failureContract.test.ts
+++ b/test/unit/resolvers/image/failureContract.test.ts
@@ -1,5 +1,4 @@
 import { capture } from '@snapshot-labs/snapshot-sentry';
-import { timeImageResolverResponse } from '../../../../src/helpers/metrics';
 import resolvers from '../../../../src/resolvers/image';
 import basename from '../../../../src/resolvers/image/basename';
 import blockie from '../../../../src/resolvers/image/blockie';
@@ -58,14 +57,6 @@ const UNRESIZED = [
   ['space-cover-sx', resolveCover]
 ] as const;
 
-async function recordedFor(provider: string) {
-  const metric: any = await timeImageResolverResponse.get();
-
-  return metric.values
-    .filter((v: any) => String(v.metricName).endsWith('_count') && v.labels.provider === provider)
-    .map((v: any) => ({ status: v.labels.status, count: v.value }));
-}
-
 const UPSTREAM_404 = Object.assign(new Error('Request failed with status code 404'), {
   response: { status: 404 }
 });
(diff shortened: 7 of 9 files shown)


tony8713:
A pre-PR adversarial review ran on this branch. What it found and what I did:

**Fixed.** `GET /avatar/.lens` sent a Sentry event on every request: the id `.lens` reaches the Lens API with an empty localName, which it answers with a validation error under HTTP 200, so neither the classifier nor the 404 rule saw it. Anyone could have held that URL down. Same class, also fixed: a name with an empty component (`/avatar/.cb.id`) makes namehash throw before anything is asked.

**Fixed.** An offchain ENS resolver answers a missing text record with a CCIP failure carrying no status, where an onchain one returns null, so every `cb.id` avatar request reported. Falling through to the metadata endpoint is what a null already did, so no response moves.

**Kept, and written into the body.** A 404 on a GraphQL endpoint is muted along with the routine ones, and the routes with a single resolver in their chain are where that costs most. The metric is the compensating signal and it counts a failure whether or not it was reported. I would rather take that than mute on an English string, and the alternative of tagging image fetches so their 404s can be told apart is a bigger change than this step should carry.

**Kept.** `farcaster` has no missing-key guard where `coingecko` does. I tried adding one and took it back out: it makes three existing tests pass while asserting nothing, and it reproduces #515, where a resolver that is switched off is indistinguishable from one with nothing to say. A rotated key reporting once per request is loud on purpose, which is also what #511 records.

**Kept.** The metric's `status` means "did not throw" rather than "found an image", so `ens` and `trustwallet` carry a high status-0 baseline. That is the same fact the 404 mute exists for, and it matches what `lookupDomains` and `addressResolvers` already record. Noted in the body rather than given a third value.

The review also confirmed the new `snapshot.ts` guard against the live hub, and mutated every test in the diff.

src/resolvers/image/ens.ts
```diff
@@ -12,18 +13,26 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
...
   }
 
-  let url = await ensResolver.getText('avatar');
+  let url: string | null = null;
+  try {
+    url = await ensResolver.getText('avatar');
+  } catch (err: any) {
+    // An offchain resolver answers "no such record" with a CCIP failure rather
+    // than a null, and every cb.id name resolves offchain.
+    if (err?.code !== 'SERVER_ERROR') throw err;
```
wa0x6e: Minor. This catch is written for the no-record case, but on ethers 5.7.2 `ccipReadFetch` uses `SERVER_ERROR` for two different things (`base-provider.js:895-908`): a gateway 4xx (`"response not found during CCIP fetch: …"`, the no-record answer) and exhausted 5xx (`"error encountered during CCIP fetch: …"`); `fetchJson` adds the same code for gateway-down / non-JSON. All rethrow unchanged past `_call`. So a cb.id gateway outage reads as "no avatar record" and falls through to `metadata.ens.domains` (which then either serves or 404s, muted) — no signal, on the branch that is adding the signal.

RPC transport failures are *not* affected — `checkError` turns those into `CALL_EXCEPTION` and `Resolver._fetch` swallows them to `null` before this catch, so the residual is CCIP-only.

If the no-record answer you observed for `cb.id` is a gateway 4xx (one live call against `greg.cb.id` with `getText` settles it), the catch can be narrowed on ethers' own wording:

```ts
if (!err?.reason?.startsWith('response not found during CCIP fetch')) throw err;
```

(and the `noData.test.ts` stub gains `reason: 'response not found during CCIP fetch: …'`). If it turns out to be a 5xx, leave the catch as is — worth a line in the comment saying an outage is muted here too.

src/resolvers/image/ens.ts
```diff
@@ -12,18 +13,26 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
 }
 
 export default async function resolve(nameOrAddress: string) {
-  const provider = getProvider(1);
   const ensName = await castToEnsName(nameOrAddress);
 
-  if (!ensName) return false;
+  // The route hands over whatever is in the path, and namehash throws on a name
+  // with an empty component rather than answering that nobody owns it.
+  if (!ensName || !isValidName(ensName)) return false;
```
wa0x6e: `isValidName` only rejects empty components — it does not enforce DNS's 63-byte label cap. `dnsEncode` does, and it throws a plain `Error` (`code: undefined`), from *outside* ethers' own try/catch in `Resolver._fetch`'s wildcard path. So `GET /avatar/<'a'×64>.cb.id` walks up to `cb.id`'s offchain resolver, hits `dnsEncode(this.name)` in `getText`, and the throw is neither `SERVER_ERROR` (rethrown here) nor a routine miss → captured, once per request. Verified against the installed `@ethersproject/hash@5.8.0` / `providers@5.7.2`. Same request-shaped class as the `.cb.id` / `.lens` guards this diff adds.

```suggestion
  if (!ensName || !isValidName(ensName)) return false;

  // isValidName doesn't enforce DNS's 63-byte label cap; dnsEncode does, and every
  // offchain resolver's wildcard path calls it, even to read a name with no record.
  try {
    dnsEncode(ensName);
  } catch {
    return false;
  }
```

Plus `import { dnsEncode, isValidName } from '@ethersproject/hash';` on line 2, and a `noData.test.ts` case beside the `.cb.id` one:

```ts
it('answers false for a label dnsEncode cannot take, without asking', async () => {
  await expect(ens(`${'a'.repeat(64)}.cb.id`)).resolves.toBe(false);
  expect(mockedGetProvider).not.toHaveBeenCalled();
});
```

src/resolvers/image/lens.ts
```diff
@@ -26,7 +26,12 @@ export default async function resolve(domainOrAddress: string) {
   if (isAddress(domainOrAddress)) {
     request = { address: getAddress(domainOrAddress) };
   } else if (domainOrAddress.endsWith(LENS_EXTENSION)) {
-    request = { username: { localName: domainOrAddress.split(LENS_EXTENSION)[0] } };
+    const localName = domainOrAddress.split(LENS_EXTENSION)[0];
+    // The API rejects an empty localName rather than answering that it has no
+    // such account, and `/avatar/.lens` reaches here.
+    if (!localName) return false;
```
wa0x6e: The empty case is one of the API's input rejections, not the class. `localName` over 254 chars answers HTTP 200 with a GraphQL envelope error (`Failed to parse "String": the string length is 255, must be less than or equal to 254`), which `graphQlCall` throws with `status: 200` — neither silenced nor a routine miss, so `/avatar/<255 chars>.lens` reports once per request. Verified live at the boundary (254 → `{account: null}`, 255 → parse error); uppercase / dotted / short names all answer cleanly, so a charset rule would only add false negatives.

```suggestion
    // The API rejects an empty or over-long localName rather than answering that
    // it has no such account: `/avatar/.lens` and a 255+ char handle both reach here.
    if (!localName || localName.length > LOCAL_NAME_MAX_LENGTH) return false;
```

with, next to `LENS_EXTENSION`:

```ts
// The API's own parse rule for AccountRequest.username.localName: anything longer fails to parse.
const LOCAL_NAME_MAX_LENGTH = 254;
```

and a `noData.test.ts` case beside the empty one:

```ts
it('answers false for a local name longer than the API takes, without asking', async () => {
  await expect(lens(`${'a'.repeat(255)}.lens`)).resolves.toBe(false);
  expect(mockedAxios).not.toHaveBeenCalled();
});
```

src/resolvers/image/index.ts
```diff
@@ -29,12 +30,22 @@ type Resolver = {
...
+}
+
+function withFailureContract(name: string, resolve: ResolverFn): ResolverFn {
   return async (...args) => {
     try {
       return await resolve(...args);
-    } catch {
-      // Silent on purpose: capturing here reports every routine upstream 404.
+    } catch (err) {
+      if (!isSilencedError(err) && !isRoutineMiss(err)) {
```
wa0x6e, line 44: Informational, no change asked. The failure path is uncached (`api.ts` `!baseImage` returns `SHORT_CACHE` without `set()`), so a repeated bad id re-runs the whole chain and re-reports every unenumerated error class — 400/422/500 from lens / hub / neynar, every HTTP-200 envelope error. Checked the precedent this PR cites: `resolvers/address` and `lookupDomains` are the same shape (uncached, same classifier) and hold the project's two largest issues in Sentry (3494 and 747 events over ~57d), grouped by fingerprint into one issue each rather than fragmenting — so grouping has carried it so far. There is no `beforeSend` / `sampleRate` in `snapshot-sentry`, so if this route turns out noisier than its siblings, the lever is a per-issue rate limit in Sentry, not code here. Worth one look at `tags.provider` volume in the first day after deploy.

src/resolvers/image/ens.ts
tony8713: Settled with the live call. `getText('avatar')` on `greg.cb.id` through `rpc.snapshot.org/1` gives `code: SERVER_ERROR` with `reason: "response not found during CCIP fetch: unknown error"`, so the answer we were reading as no-record is the gateway 4xx branch, not the exhausted 5xx one. Narrowed on ethers' own wording as you wrote it, and the `noData` stub now carries the reason.

There is a second case beside it feeding the 5xx wording (`error encountered during CCIP fetch: ...`) and expecting a reject, so an outage is pinned by a test rather than left to the reader: dropping the `startsWith` reddens exactly that one.

One residual, since the match is now on a message rather than a code. Nothing pins that wording against an ethers upgrade, because both cases are hand-built stubs and no live test reads a `cb.id` record. If the wording moves, a no-record answer starts reporting instead of an outage staying quiet, which is the direction I would rather fail in.

src/resolvers/image/ens.ts
tony8713: Confirmed, and it reaches `dnsEncode` for the reason you give. Against the live registry, `<'a'x64>.cb.id` finds no resolver of its own, walks up to `cb.id`'s at `0x1934FC75aD10d7eEd51dc7A92773cAc96A06BE56`, `supportsWildcard()` is true, and the `dnsEncode(this.name)` at `base-provider.js:370` sits in the branch before `_fetch` pushes its own try, so the plain `Error` ("invalid DNS encoded entry; length exceeds 63 bytes", `code: undefined`, `reason: undefined`) escapes uncaught. Guarded before the provider call, with the `noData` case you wrote.

Two things fell out of it that are worth your call rather than mine.

**The guard makes `isValidName` dead.** Removing `!isValidName(ensName)` now reddens nothing: `dnsEncode` turns away every name it did. Across empty components, a trailing dot, leading and inner whitespace, and zero-width and soft-hyphen labels, `isValidName` and `dnsEncode` reject exactly the same set, and the only input where they part is `''`, which `!ensName` already catches. Before this commit that line was pinned, and removing it reddened `answers false for a name namehash cannot take`. I left both in rather than collapse them on the strength of a thirty-name sample, but if you want one guard, `dnsEncode` is the one that covers both classes. Note its cap is bytes, not characters, same as the lens thread: 32 `é` is 64 bytes and rejected.

**Guarding before the provider call has a cost.** `dnsEncode` is only reached on the wildcard path, so a 64-byte label whose *own* resolver is onchain used to resolve its avatar and now answers false. A parent's resolver cannot stand in for it either way, since `_getResolver` returns null for a parent resolver that is not wildcard. So the exposure is a registered long-label name with an onchain resolver and an avatar set, against a path anyone can reach with a URL, and I took the trade. If you would rather keep that name working, the alternative is to let it reach `getText` and match the message in the catch below, at two RPC round trips per request.

Dropped the comment from the suggestion. A guard here gets a test rather than prose, and deleting the `dnsEncode` call reddens its case on its own.

src/resolvers/image/lens.ts
tony8713: Confirmed at the boundary, and one correction to the shape of the guard. 254 answers `{"data":{"account":null}}` and 255 answers the parse error, as you had it, but the number the API counts is UTF-8 bytes rather than characters: a localName of 200 `é` comes back as "the string length is 400", and 127 `é` plus one `a` is 255 to the API while `String.length` reads it as 128. So a `.length` guard passes the multi-byte case straight through to the same report. It is `Buffer.byteLength` against `LOCAL_NAME_MAX_BYTES` instead, and the constant is named for what it counts.

Three cases rather than one, because the number is the whole finding: 255 ASCII answers false without asking, 255 bytes of two-byte characters does too, and 254 still asks. Neither the guard nor the boundary nor the counting can move without something going red. No comment on it, for the same reason as above.

src/resolvers/image/index.ts
tony8713: Agreed on the substance, and nothing changed for it. The uncached failure path is the part worth watching, and `tags.provider` volume over the first day is the read to take.

One correction to the premise, since it changes what the options are. `snapshot-sentry` 3.1.0 does have a `beforeSend`: `dist/index.js:83`, where it scrubs key-shaped env values out of every exception and then returns the event unchanged. It also takes `ignoreErrors` as an `initLogger` option (`InitLoggerOptions` in `index.d.ts`), and `src/instrument.ts` calls `initLogger()` with no arguments, so that lever is unused rather than absent. `SENTRY_TRACE_SAMPLE_RATE` is traces only, so you are right that there is no error sample rate.

So a per-issue rate limit in Sentry is still the one I would reach for first, being a dial rather than a deploy, but it is a preference and not the only door.

tony8713:
All four addressed on top of `6df7196`, one commit each, replies inline.

- `7f758f4` drops the five comments this branch added. Each described a guard the suite pins, which is the same cut `afc1277` just made in `helpers/http` on master, and it is why the two suggestions above landed without their comment lines.
- `4a5d976` reads only the gateway's not-found answer as no record, so a `cb.id` outage stays visible.
- `58a8fe9` guards the DNS label cap that `isValidName` does not cover. It also makes `isValidName` redundant, which is worth your eye rather than my quiet deletion; details in that thread.
- `a5c0f91` guards a lens local name past what the API will parse. The length there counts bytes rather than characters, which the suggestion and my first pass both had as `.length`.

Nothing changed for the `index.ts` note, which asked for none, though one detail in it needed correcting and I have replied there.

It merges clean on current master, so there is no rebase.

src/resolvers/image/ens.ts
```diff
@@ -13,19 +14,30 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
...
   const ensName = await castToEnsName(nameOrAddress);
 
-  if (!ensName) return false;
+  if (!ensName || !isValidName(ensName)) return false;
+
+  try {
+    dnsEncode(ensName);
+  } catch {
+    return false;
+  }
```
wa0x6e: Follow-up on the (resolved) dnsEncode thread — the trade you flagged as my call. Preference, with the decision left to you.

**Preferred: move the length check into the `getText` catch.** The pre-call guard rejects a population ethers itself resolves: `dnsEncode` is only reached on the wildcard path (`base-provider.js:365-370`, gated by `supportsWildcard()`), and the stock Public Resolver answers `supportsWildcard: false` (checked live via `rpc.snapshot.org/1` against `vitalik.eth`, `nick.eth`, `ens.eth`, `brantly.eth`). So a >63-byte-label name with an onchain resolver and an avatar set resolves today and answers `false` after this change. Since the plain `Error('invalid DNS encoded entry; …')` escapes `getText` uncaught (as you established), the existing catch can take it:

```diff
   if (!ensName || !isValidName(ensName)) return false;

-  try {
-    dnsEncode(ensName);
-  } catch {
-    return false;
-  }
-
   // getProvider returns `any`. Without the annotation the calls below are unchecked.
   const provider: StaticJsonRpcProvider = getProvider(1);
   const ensResolver = await provider.getResolver(ensName);
@@
   let url: string | null = null;
   try {
     url = await ensResolver.getText('avatar');
   } catch (err: any) {
+    if (err?.message?.startsWith('invalid DNS encoded entry')) return false;
     if (!err?.reason?.startsWith('response not found during CCIP fetch')) throw err;
   }
```

`isValidName` stops being dead under this shape — it goes back to guarding `namehash` inside `getResolver` (the empty-component class). Costs, stated: the held-down long-label URL now pays `getResolver` + `supportsWildcard` round trips before rejection — the same price any nonexistent-name request already pays, so not a new exposure; and it adds a third version-pinned ethers message string, with the same loud-direction drift you accepted for the CCIP wording. The `noData` case flips from "provider not asked" to stubbing `getText` rejecting with the plain `Error` and expecting `false`.

**Alternative, if you'd rather keep the cheap pre-call rejection:** keep the trade as taken, but collapse the guards — drop `isValidName` and keep only the `dnsEncode` try/catch. Your thirty-name sample generalizes: `dnsEncode` calls the identical `ensNameSplit` that `isValidName` wraps and adds only the 63-byte check (`namehash.js` source), so with the pre-call in place `isValidName` provably rejects nothing `dnsEncode` doesn't. That makes it a source-level proof rather than a sample, if you want the single-guard form you left pending.

Keeping it exactly as-is is also defensible — the harmed population is narrow. Your call; nothing here blocks.

src/resolvers/image/lens.ts
```diff
@@ -26,7 +27,10 @@ export default async function resolve(domainOrAddress: string) {
   if (isAddress(domainOrAddress)) {
     request = { address: getAddress(domainOrAddress) };
   } else if (domainOrAddress.endsWith(LENS_EXTENSION)) {
-    request = { username: { localName: domainOrAddress.split(LENS_EXTENSION)[0] } };
+    const localName = domainOrAddress.split(LENS_EXTENSION)[0];
```
wa0x6e: Pre-existing, and this diff rewrites the line, so worth deciding rather than inheriting. `split(LENS_EXTENSION)[0]` truncates at the *first* `.lens`, not the trailing one: `'a.lensb.lens'.split('.lens')[0] === 'a'`, so `GET /avatar/a.lensb.lens` queries account `a` — a different account than the request named — and may serve its avatar. No registered handle can collide (local names can't contain `.`), so the reach is malformed input only; `slice` is the shape the `endsWith` guard above already implies:

```suggestion
    const localName = domainOrAddress.slice(0, -LENS_EXTENSION.length);
```

The guards below behave identically for every existing case (empty, oversized, byte-vs-char length, max-valid). A pin beside the lens `noData` cases keeps it from regressing — it reddens on the current line:

```ts
it('queries the text before the last .lens, not the text before the first one', async () => {
  mockedAxios.mockResolvedValue({ status: 200, data: { data: { account: null } } });

  await lens('a.lensb.lens');

  expect(mockedAxios.mock.calls[0][0].data.variables.request.username.localName).toBe('a.lensb');
});
```

src/resolvers/image/ens.ts
wa0x6e: Residual within the 4xx branch this thread already settled, confirmed by reading `ccipReadFetch` directly (`base-provider.js:895-897`): ethers collapses the *entire* 400-499 range into the same `SERVER_ERROR` / `"response not found during CCIP fetch: " + errorMessage` shape, and the real status code never reaches `err` — the `throwError` params here are only `{ url, errorMessage }`, no `status`. So a gateway 401/403/429 reads exactly like `greg.cb.id`'s no-record 404 above, falls through to the `metadata.ens.domains` fallback, and (muted by `isRoutineMiss`'s 404 check) degrades to blockie with no signal anywhere.

No clean narrowing exists from inside this catch. `errorMessage` is the one candidate field on `err`, and it was already `"unknown error"` in the live capture above — this gateway doesn't send anything to key on, so a string match here would either match nothing or risk exactly the failure direction you flagged (a no-record answer starting to report). Recovering the real status would mean overriding `ccipReadFetch` on the provider to capture it before ethers discards it, which is bigger than this catch and needs its own live verification, not something to fold in here.

Not a regression — this exact case was already fully silent before this branch (every resolver failure was `false`, zero signal), and it's the same "fail quiet on CCIP ambiguity" call already made above, just a sharper instance of it. Flagging so it's a named, known gap rather than a rediscovered one if `.cb.id` avatars go quiet during a gateway incident — and so the PR body's "only the gateway's own not-found answer" line is read as "only non-5xx," not as "only genuine not-found". Worth mirroring in the test name too: the no-record case's title currently claims the wording uniquely means no record, when it means any gateway 4xx.

src/resolvers/image/ens.ts
tony8713: Taken your preferred shape. The pre-call guard is out and the length check now sits in the `getText` catch, so a long-label name whose own resolver is onchain resolves exactly as it does today.

Confirmed the population it was costing, and the string it now matches, live through `rpc.snapshot.org/1`: `vitalik.eth`, `nick.eth`, `ens.eth` and `brantly.eth` all answer `supportsWildcard: false` and serve their avatar record, and `<'a'x64>.cb.id` walks to `cb.id`'s wildcard resolver and throws the plain `Error('invalid DNS encoded entry; length exceeds 63 bytes')` out of `getText`, with `code` and `reason` both undefined.

`isValidName` is live again under this shape, and pinned rather than argued: dropping it reddens `answers false for a name namehash cannot take`, whose assertion is that the provider is never asked. Two cases sit on the moved check, one for the long label rejected in the catch and one for a long label still served when nothing dns-encodes it, so the pre-call guard cannot come back without something going red.

src/resolvers/image/lens.ts
tony8713: Fixed as written, with your pin. `slice` is the shape the `endsWith` above already implies, and the four guards below behave identically for every case they cover.

Putting `split(LENS_EXTENSION)[0]` back reddens the new case on its own and nothing else in the file, which is the reach you described: malformed input answered under a name it did not carry.

src/resolvers/image/ens.ts
tony8713: Agreed, and the naming is what changes rather than the code. The stub is `CCIP_GATEWAY_4XX` now and its case reads `reads any 4xx from the offchain gateway as no record`, so nothing claims that wording is unique to a missing record, and the PR body line says the same.

Read `ccipReadFetch` the same way you did: the `throwError` call carries `{ url, errorMessage }` and no status, and `errorMessage` was already `unknown error` in the live `greg.cb.id` capture, so there is nothing on `err` that tells a 401 from a 404. Leaving it as the named gap rather than reaching for an override of `ccipReadFetch`, which is its own change with its own live verification.

tony8713:
All three addressed on top of `bab7f5b`, one commit each, replies inline.

- `ca5047a` takes your preferred shape on the trade you left me: the DNS label check moves out of the pre-call guard and into the `getText` catch, so a long-label name whose own resolver is onchain resolves as it does today, and `isValidName` goes back to guarding `namehash`.
- `790b007` cuts a lens local name at the last `.lens` rather than the first, with your pin.
- `55ca45f` renames the CCIP case for the range it covers, and the PR body line says the same.

The `index.ts` thread is the one still open, since the ball is with you there rather than with me.

wa0x6e (changes requested):
utAck, but waiting release of https://github.com/snapshot-labs/snapshot.js/pull/1234 , and avoid the getProvider type casting

src/resolvers/image/ens.ts
```diff
@@ -13,19 +14,25 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
...
 export default async function resolve(nameOrAddress: string) {
-  // getProvider returns `any`. Without the annotation the calls below are unchecked.
-  const provider: StaticJsonRpcProvider = getProvider(1);
   const ensName = await castToEnsName(nameOrAddress);
 
-  if (!ensName) return false;
+  if (!ensName || !isValidName(ensName)) return false;
 
+  // getProvider returns `any`. Without the annotation the calls below are unchecked.
+  const provider: StaticJsonRpcProvider = getProvider(1);
```
wa0x6e: To be removed after release of https://github.com/snapshot-labs/snapshot.js/pull/1234

tony8713: Released snapshot.js v0.17.0 contains the conditional provider return type in its published declarations. Updated to ^0.17.0, removed the cast, and made our wrapper generic so Starknet literals keep their provider type.

wa0x6e (approved):
utAck

wa0x6e (approved):

wa0x6e (approved):
utAck

PR #553, opened 2026-08-16
fix(helpers): bound the shared image fetch with a whole-call deadline

tony8713:
`fetchHttpImage` is the shared fetch behind every image resolver, and the only thing bounding it was the axios `timeout` in `axiosDefaultParams`. That is an idle timeout on the socket rather than a budget for the call: an upstream that keeps sending, however slowly, resets it every time it writes, so a response body that starts and never ends holds the resolver open with nothing underneath it. `src/api.ts` fans the resolver chain out with `Promise.all` and the failure contract in `src/resolvers/image/index.ts` turns a throw into `false`, and neither of those reaches a promise that simply never settles.

This puts the call inside the `withDeadline` from #534, the way #541 and #542 did for their own call sites, so the budget covers the transfer rather than the request.

## The budget changed

It is now 5s rather than the shared 10s, and `withDeadline` takes the number as an optional second argument so a call site can say what it is bounding. The default is unchanged and no other call site passes one.

5s is also not the 5s that went. The axios timeout was per request, re-armed by every write on the socket, and never covered the body at all. This one covers the whole call: connect, headers, redirects and the body read.

## And the client changed

axios was the only HTTP client here when the repo started, before Node had one of its own, and the buffer response was never the reason it stayed. Two things follow from dropping it.

It answers a signal with a `Cancel` of its own rather than rejecting with the signal's reason, and that object carries neither an `Error` prototype nor the name `isSilencedError` reads, so an abort arrived at a caller as a resolver failure rather than as one of our own deadlines. `fetch` raises a `DOMException` named `AbortError`, which the classifier already matches, so the translation the deadline would otherwise need is gone rather than added.

`fetch` also resolves a non-2xx where axios raised one. `user-cover`, `space-cover`, `space-logo` and `space-cover-sx` are registered with `resize: false`, so what this returns for them reaches sharp in `src/api.ts` with no guard between (#495), and an error page would arrive there as an image. The status has to become a throw, and `httpError` carries it in both the places `isSilencedError` reads, which is what keeps a 504 and a 429 muted the way axios's `response.status` did.

axios stays in the repo: `graphQlCall` and the starknet image resolver still use it.

Refs #536, #206.

### DIFF
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
@@ -1,9 +1,8 @@
 import http from 'http';
 import https from 'https';
 import { getAddress } from '@ethersproject/address';
-import axios from 'axios';
 import { isStarknetAddress } from './address';
-import { withDeadline } from '../utils';
+import { httpError, withDeadline } from '../utils';
 
 export const axiosDefaultParams = {
   httpAgent: new http.Agent({ keepAlive: true }),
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
utAck

PR #554, opened 2026-08-17
fix(provider): keep one provider, the snapshot.js one

tony8713:
`src/helpers/provider.ts` exports two different things that both
mean "provider": `provider`, which wraps `snapshot.utils.getProvider` and points at
`BROVIDER_URL`, and `getProvider`, a hand-rolled `StaticJsonRpcProvider` with its own
cache, a hardcoded `rpc.snapshot.org` host and a longer per-request timeout. Every
address resolver imports the first one aliased to `getProvider`, and
`src/resolvers/image/ens.ts` imports the second one under that same name. So
`getProvider` means two different things depending on which file you are reading.

Drop the hand-rolled one and keep the snapshot.js wrapper, renamed to `getProvider`
so no call site has to alias it.

## This is not a pure rename

**The ENS image resolver changes both endpoint and budget.** `src/resolvers/image/ens.ts`
was the only user of the hand-rolled provider. Its import line does not change at all,
but it now resolves to the shared one, so it moves from a hardcoded `rpc.snapshot.org/1`
at a 20s per-request timeout to `BROVIDER_URL` at 5s. That timeout bounds one JSON-RPC
request rather than the whole resolution, which here is the two `eth_call`s behind
`getResolver` and `getText('avatar')`. Exceeding it is silent: the resolver answers
`false` and the avatar route serves its fallback under the short cache header.

**`spaceId` and `unstoppableDomains` start honouring `BROVIDER_URL`.** Both passed
`{ timeout: 5e3 }`, which was already the wrapper's default, and because a partial
options object replaced the defaults rather than merging into them, passing it was also
what made those two fall back to snapshot.js's own default host. Dropping the argument
fixes that, puts every call site on the same shape, and lets the options parameter go.
Worth confirming against the deployed `BROVIDER_URL` before merge: if it is
`https://rpc.snapshot.org`, as `.env.example` and `test/.env.test` both set it, this
one moves nothing.

Caching is unaffected. snapshot.js memoizes per network, URL and timeout, so the
hand-rolled cache goes away without a per-request construction taking its place.

The wrapper keeps snapshot.js's `any` return. Two of the eight call sites are Starknet,
where that same function hands back a starknet.js `RpcProvider` rather than a
`StaticJsonRpcProvider`, so no one concrete return type is true for all of them.
`src/resolvers/image/ens.ts` annotates its own local instead, which is what keeps its
ethers calls checked.

### DIFF
--- a/src/helpers/provider.ts
+++ b/src/helpers/provider.ts
@@ -3,22 +3,12 @@ import snapshot from '@snapshot-labs/snapshot.js';
 import { Address } from './types';
 
 const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';
-const providers: Record<string, StaticJsonRpcProvider> = {};
 
-export function provider(
-  network: string,
-  providerOptions: { broviderUrl?: string; timeout?: number } = { broviderUrl, timeout: 5e3 }
+export function getProvider(
+  network: string | number,
+  providerOptions: { broviderUrl?: string; timeout?: number } = {}
 ) {
-  return snapshot.utils.getProvider(network, providerOptions);
-}
-
-export function getProvider(network: number): StaticJsonRpcProvider {
-  if (!providers[`_${network}`])
-    providers[`_${network}`] = new StaticJsonRpcProvider(
-      { url: `https://rpc.snapshot.org/${network}`, timeout: 20e3, allowGzip: true },
-      network
-    );
-  return providers[`_${network}`];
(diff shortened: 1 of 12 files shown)


tony8713:
Base note: #552's branch lives on a fork (chai3-bot/stamp), so GitHub will not let me target it and this is opened against master instead. The file list and commit list therefore include #552's commits as well. Mine are the three on top: 7eaa457, 3bc8132, 30e8407. Once #552 merges this collapses to just those.

wa0x6e (approved):
utAck

PR #555, opened 2026-08-17
fix(resolvers): stop the shibarium DNS cache from holding the process open

tony8713:
`yarn test` prints its summary and then the node process stays up. `package.json` already
worked around that on two of the three test scripts: `test:integration` and `test:e2e` both
pass `--forceExit`, and they are also the two that pass `--runInBand`, which is where it
bites. `test` inherits the same problem wherever jest picks a single worker.

What holds the event loop open is the DNS cache inside `@webinterop/dns-connect`. Its
`InMemoryCache` evicts each entry with a plain `setTimeout`, and the DNS module writes the
TLD registration status with a one-day TTL on every `resolve()`. That timer is ref'd, so one
Shibarium name resolution keeps the process alive for a day after the work is done.

This is not only a test problem. `lookupAddresses`, `resolveNames` and `getOwner` each build
their own `DNSConnect`, so every request reaching a Shibarium resolver arms its own timers
over a cache instance nothing will ever read again, and the pending timer keeps that cache
reachable until it fires.

So the cache is replaced with one that expires on read rather than on a timer. It answers the
same and holds nothing open. The client is built in one place now instead of three, and the
`--forceExit` flags go with it.

Not addressed here: those call sites each build a fresh client, so the cache it carries can
only ever serve hits inside a single call. Sharing one client per forwarder domain would be a
real improvement, but it is a behaviour change and belongs on its own.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "start": "node dist/src/index.js",
     "test": "jest",
     "test:coverage": "jest --coverage",
-    "test:integration": "jest --runInBand --forceExit test/integration",
-    "test:e2e": "jest --runInBand --forceExit test/e2e/"
+    "test:integration": "jest --runInBand test/integration",
+    "test:e2e": "jest --runInBand test/e2e/"
   },
   "dependencies": {
     "@adraffy/ens-normalize": "^1.10.0",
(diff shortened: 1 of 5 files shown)


wa0x6e (approved):
utAck

PR #556, opened 2026-08-17
feat(cache): share one cache module between the DNS and address stores

tony8713:
Depends on #555

#555 is in master, so the Shibarium DNS cache expires on read now instead of holding a timer.
It is still built inside `dnsConnect()` though: every call gets a fresh client with a fresh
cache, so nothing that cache learns outlives the request that filled it and the TLD
registration lookup starts from cold every time. The memory store below is what makes that
cache survive across requests rather than being created and discarded inside each call, which
is the whole point of it and easier to see now the timer fix has landed.

The address resolvers cache is the same job again, written straight against redis in
`resolvers/address/cache.ts`.

So there is one `src/cache` module now: a `CacheStore` interface with a memory store and a
redis store behind it, and a policy per kind instead of one policy for everything.

The address cache moves onto the redis store with its behaviour held still: same
`address-resolvers` prefix, same TTL, same empty string for an address that resolves to no
name.

The DNS cache moves onto the memory store, kept at module scope per forwarder domain.
dns-connect keys its entries by name and network alone, so mainnet and testnet need separate
stores or they answer each other's lookups.

Memory rather than redis for that one. dns-connect's provider interface is async so redis
would fit, but the library awaits the cache unguarded and a rejected read propagates straight
out of `resolve()`, which would let a redis blip fail a Shibarium resolution that would
otherwise have succeeded. What redis would add is sharing between instances, and the repeats
that sharing would serve are largely absorbed by the address resolvers cache sitting in front
of Shibarium.

Two things the module is explicit about.

**The memory store is bounded.** Expiring on read is what keeps it timer-free, but it never
reaches an entry that nothing asks for again. Per call that did not matter, since the whole
cache died with the call. At module scope it does, so the store takes a maximum entry count
and drops whatever has gone unread longest.

**Empty answers are a per-kind choice.** The address routes go on caching an address with no
name for the full TTL. DNS does not cache one at all, so a name that failed to resolve is
tried again on the next request rather than held for the record's TTL.

Not addressed here: the address routes cache an empty answer for twelve hours, so one
transient upstream timeout pins "no name" for half a day and the recovered upstream is not
consulted again until the key expires. That is a production behaviour change and wants its
own review.

### DIFF
--- a/src/cache/index.ts
+++ b/src/cache/index.ts
@@ -1,0 +1,3 @@
+export { CachePolicy, CacheStore } from './types';
+export { MemoryStore, MemoryStoreOptions } from './stores/memory';
+export { RedisStore, RedisStoreOptions } from './stores/redis';
--- a/src/cache/stores/memory.ts
+++ b/src/cache/stores/memory.ts
@@ -1,0 +1,64 @@
+import { CachePolicy, CacheStore } from '../types';
+
+export type MemoryStoreOptions = CachePolicy & {
+  maxEntries: number;
+};
+
+type Entry = { value: string; expiresAt: number };
+
+export class MemoryStore implements CacheStore {
+  private entries = new Map<string, Entry>();
+
+  constructor(private readonly options: MemoryStoreOptions) {}
+
+  async get(key: string): Promise<string | undefined> {
+    const entry = this.entries.get(key);
+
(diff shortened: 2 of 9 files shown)


wa0x6e (approved):
utAck

PR #568, opened 2026-08-18
refactor(ens): batch reverse lookup through the Universal Resolver

### LINKED ISSUE #558: refactor(ens): batch reverse lookup through the Universal Resolver (1/3)
Part 1 of 3 of moving ENS resolution onto the Universal Resolver via viem, picking up #483 with the batching it was missing. Reverse first: `lookupAddresses` takes up to 50 addresses per request, so it is where the RPC volume is.

## Today

`src/resolvers/address/ens.ts:lookupAddresses`:
1. `getNames(address[])` on the ReverseRecords contract — one call for the batch, on-chain records only.
2. For every address it misses, `provider.lookupAddress()` (ethers) — one `registry.resolver()` RPC each, more if a reverse resolver exists (CCIP names). Most addresses have no primary name, so this is ~1 RPC per address per uncached lookup.

## Change

- Add `"viem": "^2.55.11"` as a direct dependency (same range `@snapshot-labs/snapshot.js@^0.15.2` already pulls, dedupes to one copy — see review on #483).
- New `src/resolvers/address/universalResolver.ts`: viem public client on `${BROVIDER_URL}/1` with `batch: { multicall: { batchSize, wait: 0 } }`, plus the `settle()` / `reverseLookup()` helpers from #483 (`src/addressResolvers/universalResolver.ts` on branch `migrate-ens-universal-resolver`; paths have since moved to `src/resolvers/address/`).
- `lookupAddresses` → `reverseLookup(normalizedAddresses)`. Delete the ReverseRecords call and the `lookupAddress` fallback.

With multicall batching, on-chain hits and `ResolverNotFound` misses settle in one `aggregate3`; only entries whose resolver reverts `OffchainLookup` pay a per-name CCIP round trip. The Universal Resolver forward-verifies the reverse claim on chain, so no separate check.

## Acceptance

- Unit test with a mocked transport: N on-chain addresses → **one** `eth_call`. Assert the count; that is the point of the change.
- Unit test: a batch containing one `OffchainLookup` entry → gateway fetch for that entry only, the rest still in the single batch.
- Existing integration fixtures keep passing: `sdntestens.eth`, `jesse.base.eth` (off-chain resolver), address whose name expired → nothing.
- No change to the `lookupAddresses` contract or to `resolvers/address/index.ts`.

## Decisions to make in the PR

- Transport timeout: repo providers use 5s; #483 used 20s.
- `batchSize`: viem's default (1024 bytes of calldata) chunks 50 reverse calls into several `aggregate3`s. Set it explicitly and let the count test prove it.

Why #483 was closed: its correctness goals had already landed (#430, #478) and it did no batching, so 50 addresses meant 50 RPCs. This issue is the batching answer.

tony8713:
ENS reverse resolution currently uses a legacy batch helper that cannot follow offchain names, then fans out through per-address fallbacks.

This replaces both paths with one viem Universal Resolver client. Onchain lookups share a Multicall3 request, while CCIP entries settle independently so one offchain lookup does not split or fail the rest.

Closes #558

### DIFF
--- a/src/resolvers/address/cache.ts
+++ b/src/resolvers/address/cache.ts
@@ -1,3 +1,4 @@
+import { capture } from '@snapshot-labs/snapshot-sentry';
 import { RedisStore } from '../../cache';
 import constants from '../../constants.json';
 import { addressResolversCacheHitCount } from '../../helpers/metrics';
@@ -38,7 +39,7 @@ export default async function cache(
     const cacheable = Object.fromEntries(
       Object.entries(results).filter(([key]) => !nonCacheable.has(key))
     );
-    await setCache(cacheable);
+    setCache(cacheable).catch(capture);
 
     return { ...cache, ...results };
   }
--- a/src/resolvers/address/cache.ts
+++ b/src/resolvers/address/cache.ts
@@ -3,6 +3,13 @@ import constants from '../../constants.json';
 import { addressResolversCacheHitCount } from '../../helpers/metrics';
 
 export const KEY_PREFIX = 'address-resolvers';
+export const NON_CACHEABLE = Symbol('non-cacheable');
+export type CacheResult = Record<string, string> & { [NON_CACHEABLE]?: string[] };
+
+export function markNonCacheable(result: Record<string, string>, keys: string[]): CacheResult {
+  if (keys.length > 0) Object.defineProperty(result, NON_CACHEABLE, { value: keys });
+  return result;
+}
 
 const store = new RedisStore({ prefix: KEY_PREFIX, maxTtl: constants.ttl, cacheEmpty: true });
 
@@ -14,7 +21,10 @@ export function setCache(payload: Record<string, string>): Promise<void> {
   return store.setMany(payload);
 }
 
-export default async function cache(input: string[], callback) {
+export default async function cache(
+  input: string[],
+  callback: (input: string[]) => Promise<CacheResult>
+) {
   const cache = await getCache(input);
   const cachedKeys = Object.keys(cache);
   const uncachedInputs = input.filter(a => !cachedKeys.includes(a));
@@ -24,7 +34,11 @@ export default async function cache(input: string[], callback) {
 
   if (uncachedInputs.length > 0) {
     const results = await callback(uncachedInputs);
-    setCache(results);
+    const nonCacheable = new Set(results[NON_CACHEABLE] || []);
+    const cacheable = Object.fromEntries(
+      Object.entries(results).filter(([key]) => !nonCacheable.has(key))
+    );
+    await setCache(cacheable);
 
     return { ...cache, ...results };
   }
--- a/src/resolvers/address/ens.ts
+++ b/src/resolvers/address/ens.ts
@@ -1,6 +1,7 @@
 import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
+import { markNonCacheable } from './cache';
 import { reverseLookup } from './universalResolver';
 import constants from '../../constants.json';
 import { isEvmAddress } from '../../helpers/address';
@@ -38,16 +39,29 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
 
   const { values, errors } = await reverseLookup(normalizedAddresses);
 
-  if (errors.length > 0 && Object.keys(values).length === 0) {
-    throw errors[0];
+  if (errors.length === normalizedAddresses.length) {
+    throw (errors.find(({ error }) => !isSilencedError(error)) || errors[0]).error;
   }
 
-  const validNames = normalizeEns(normalizedAddresses.map(address => values[address] || ''));
+  errors.forEach(({ address, error }) => {
+    if (!isSilencedError(error)) {
+      capture(error, {
+        tags: { provider: NAME },
+        contexts: { input: { lookupAddresses: [address] } }
+      });
+    }
+  });
 
-  return Object.fromEntries(
+  const validNames = normalizeEns(normalizedAddresses.map(address => values[address] || ''));
+  const result = Object.fromEntries(
     normalizedAddresses
       .map((address, index) => [address, validNames[index]])
       .filter((_, index) => !!validNames[index])
+  );
+
+  return markNonCacheable(
+    result,
+    errors.map(({ address }) => address)
   );
 }
 
--- a/src/resolvers/address/ens.ts
+++ b/src/resolvers/address/ens.ts
@@ -1,7 +1,7 @@
 import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
-import snapshot from '@snapshot-labs/snapshot.js';
+import { reverseLookup } from './universalResolver';
 import constants from '../../constants.json';
 import { isEvmAddress } from '../../helpers/address';
 import { isSilencedError, isTransportFailure } from '../../helpers/errors';
@@ -32,35 +32,17 @@ function normalizeHandles(names: Handle[]): Handle[] {
 }
 
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
-  const abi = ['function getNames(address[] addresses) view returns (string[] r)'];
   const normalizedAddresses = normalizeAddresses(addresses);
 
   if (normalizedAddresses.length === 0) return {};
 
-  let reverseRecords: string[] = [];
-  try {
-    reverseRecords = await snapshot.utils.call(
-      provider,
-      abi,
-      ['0x3671aE578E63FdF66ad4F3E12CC0c0d71Ac7510C', 'getNames', [normalizedAddresses]],
-      { blockTag: 'latest' }
-    );
-  } catch (err: any) {
-    if (err?.code !== 'CALL_EXCEPTION') throw err;
-  }
-  const validNames = normalizeEns(reverseRecords);
+  const { values, errors } = await reverseLookup(normalizedAddresses);
 
-  // The batch contract only reads on-chain reverse records. Names served by
-  // off-chain resolvers (CCIP-Read) need provider.lookupAddress, which follows
-  // the OffchainLookup flow via the ENS UniversalResolver.
-  const missing = normalizedAddresses.map((_, i) => i).filter(i => !validNames[i]);
-  const lookups = await Promise.allSettled(
-    missing.map(idx => provider.lookupAddress(normalizedAddresses[idx]))
-  );
-  const fallbackNames = normalizeEns(lookups.map(r => (r.status === 'fulfilled' && r.value) || ''));
-  missing.forEach((idx, j) => {
-    if (fallbackNames[j]) validNames[idx] = fallbackNames[j];
-  });
+  if (errors.length > 0 && Object.keys(values).length === 0) {
+    throw errors[0];
+  }
+
+  const validNames = normalizeEns(normalizedAddresses.map(address => values[address] || ''));
 
   return Object.fromEntries(
     normalizedAddresses
--- a/src/resolvers/address/ens.ts
+++ b/src/resolvers/address/ens.ts
@@ -2,7 +2,7 @@ import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import { markNonCacheable } from './cache';
-import { reverseLookup } from './universalResolver';
+import { isSilencedReverseError, reverseLookup } from './universalResolver';
 import constants from '../../constants.json';
 import { isEvmAddress } from '../../helpers/address';
 import { isSilencedError, isTransportFailure } from '../../helpers/errors';
@@ -32,6 +32,10 @@ function normalizeHandles(names: Handle[]): Handle[] {
   return normalizeEns(names).filter(h => h);
 }
 
+function isSilencedLookupError(error: unknown): boolean {
+  return isSilencedError(error) || isSilencedReverseError(error);
+}
+
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
   const normalizedAddresses = normalizeAddresses(addresses);
 
@@ -40,11 +44,13 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
   const { values, errors } = await reverseLookup(normalizedAddresses);
 
   if (errors.length === normalizedAddresses.length) {
-    throw (errors.find(({ error }) => !isSilencedError(error)) || errors[0]).error;
+    const actionable = errors.find(({ error }) => !isSilencedLookupError(error));
+    if (actionable) throw actionable.error;
+    return markNonCacheable({}, normalizedAddresses);
   }
 
   errors.forEach(({ address, error }) => {
-    if (!isSilencedError(error)) {
+    if (!isSilencedLookupError(error)) {
       capture(error, {
         tags: { provider: NAME },
         contexts: { input: { lookupAddresses: [address] } }
--- a/src/resolvers/address/universalResolver.ts
+++ b/src/resolvers/address/universalResolver.ts
@@ -13,7 +13,6 @@ import { getProviderOptions } from '../../helpers/provider';
 
 const rpcUrl = `${getProviderOptions().broviderUrl}/${mainnet.id}`;
 const EMPTY_REVERSE_ERRORS = new Set([
-  'ResolverError',
   'ResolverNotContract',
   'ResolverNotFound',
   'ReverseAddressMismatch',
--- a/src/resolvers/address/universalResolver.ts
+++ b/src/resolvers/address/universalResolver.ts
@@ -1,6 +1,8 @@
 import {
+  BaseError,
   ccipRequest,
   CcipRequestParameters,
+  ContractFunctionRevertedError,
   createPublicClient,
   http,
   Address as ViemAddress
@@ -10,6 +12,24 @@ import { withDeadline } from '../../helpers/deadline';
 import { getProviderOptions } from '../../helpers/provider';
 
 const rpcUrl = `${getProviderOptions().broviderUrl}/${mainnet.id}`;
+const EMPTY_REVERSE_ERRORS = new Set([
+  'ResolverError',
+  'ResolverNotContract',
+  'ResolverNotFound',
+  'ReverseAddressMismatch',
+  'UnsupportedResolverProfile'
+]);
+
+function isEmptyReverseError(error: unknown): boolean {
+  if (!(error instanceof BaseError)) return false;
+
+  const reverted = error.walk(cause => cause instanceof ContractFunctionRevertedError);
+
+  return (
+    reverted instanceof ContractFunctionRevertedError &&
+    EMPTY_REVERSE_ERRORS.has(reverted.data?.errorName || '')
+  );
+}
 const client = createPublicClient({
   chain: mainnet,
   batch: { multicall: { batchSize: 16 * 1024 } },
@@ -35,7 +55,7 @@ export type BatchResult = { values: Record<string, string>; errors: BatchError[]
 
 export async function reverseLookup(addresses: string[]): Promise<BatchResult> {
   const settled = await Promise.allSettled(
-    addresses.map(address => client.getEnsName({ address: address as ViemAddress }))
+    addresses.map(address => client.getEnsName({ address: address as ViemAddress, strict: true }))
   );
   const values: Record<string, string> = {};
   const errors: BatchError[] = [];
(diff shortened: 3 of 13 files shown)


src/resolvers/address/universalResolver.ts
```diff
@@ -0,0 +1,29 @@
+import { createPublicClient, http, Address as ViemAddress } from 'viem';
+import { mainnet } from 'viem/chains';
+
+const rpcUrl = `${process.env.BROVIDER_URL || 'https://rpc.snapshot.org'}/1`;
+const client = createPublicClient({
+  chain: mainnet,
+  batch: { multicall: { batchSize: 64 * 1024, wait: 0 } },
+  transport: http(rpcUrl, { retryCount: 0, timeout: 5e3 })
```
wa0x6e: The 5s transport timeout only bounds the JSON-RPC leg. viem's default `ccipRequest` calls bare `fetch` with no `AbortSignal` unless the client's `ccipRead.request` option supplies one — so one stalled offchain gateway (e.g. a `*.base.eth` Basename, common in real traffic) holds the entire `lookupAddresses` batch open (`Promise.allSettled` waits for all 50) up to undici's default header timeout (~300s). The old ethers fallback path capped this at 120s, so the PR regresses it to effectively unbounded.

This is the same gap snapshot.js's own viem client closes with its `ccipReadWithDeadline` wrapper; the fix below reuses this repo's existing `withDeadline` helper (default 10s per gateway request), and the resulting `AbortError` is already special-cased by `isSilencedError`. `tsc --noEmit` and the existing unit suites pass unmodified with this applied.

```suggestion
import {
  ccipRequest,
  CcipRequestParameters,
  createPublicClient,
  http,
  Address as ViemAddress
} from 'viem';
import { mainnet } from 'viem/chains';
import { withDeadline } from '../../helpers/deadline';

const rpcUrl = `${process.env.BROVIDER_URL || 'https://rpc.snapshot.org'}/1`;
const client = createPublicClient({
  chain: mainnet,
  batch: { multicall: { batchSize: 64 * 1024, wait: 0 } },
  transport: http(rpcUrl, { retryCount: 0, timeout: 5e3 }),
  // The transport timeout only covers the JSON-RPC leg; CCIP-Read gateway
  // requests use their own fetch, which viem otherwise leaves unbounded.
  ccipRead: {
    request: (parameters: CcipRequestParameters) =>
      withDeadline(signal => ccipRequest({ ...parameters, requestOptions: { signal } }))
  }
```

src/resolvers/address/ens.ts
```diff
@@ -32,35 +32,17 @@ function normalizeHandles(names: Handle[]): Handle[] {
...
-      abi,
-      ['0x3671aE578E63FdF66ad4F3E12CC0c0d71Ac7510C', 'getNames', [normalizedAddresses]],
-      { blockTag: 'latest' }
-    );
-  } catch (err: any) {
-    if (err?.code !== 'CALL_EXCEPTION') throw err;
+  const { values, errors } = await reverseLookup(normalizedAddresses);
+
+  if (errors.length > 0 && Object.keys(values).length === 0) {
+    throw errors[0];
```
wa0x6e: This throws the raw viem error chain (`ContractFunctionExecutionError` → `CallExecutionError` → `TimeoutError`/`HttpRequestError`), and `isSilencedError` (`src/helpers/errors.ts`) doesn't match it — it only inspects `error`/`error.error`/`error.cause` one level deep for `code`/`status`, and its message list doesn't include viem's wording. The old ethers path threw `code: 'TIMEOUT'`, which the codes list silenced, so every transient RPC blip (timeout, connection refused/reset) on this path now reaches Sentry where it used to be muted.

Reproduced both shapes locally: a transport timeout (`BROVIDER_URL=http://10.255.255.1`) gives a 3-level chain ending in `TimeoutError`; a connection-refused (`http://127.0.0.1:19999`) gives a 4-level chain ending in `Error{code:'ECONNREFUSED'}` — both return `false` from `isSilencedError` today. Message matching alone can't rescue the refused/reset case (the `.details` cascade breaks at the non-viem `TypeError('fetch failed')` node), so the fix walks the cause chain with viem's own `BaseError.walk()`.

The fix lives in `src/helpers/errors.ts`, outside this PR's diff, so no suggestion block — patch below. Verified: both shapes silenced after the fix, negative controls (real 401, contract revert) still captured, `tsc --noEmit` clean, 37 existing tests across 4 suites pass.

```diff
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
@@ -16,6 +16,12 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
   // An abort is always one of our own deadlines, and each transport words it differently.
   if (error.name === 'AbortError') return true;
 
+  // viem wraps transport failures 3-4 levels deep (ContractFunctionExecutionError ->
+  // CallExecutionError -> HttpRequestError/TimeoutError -> the raw cause), and none
+  // of the wrapper levels carry `code`/`status`. `walk()` is viem's own BaseError
+  // method for reaching the innermost cause in one call instead of a manual loop.
+  const root = typeof error.walk === 'function' ? error.walk() : undefined;
+
   const messages = [
     'invalid token ID',
     'is not supported',
@@ -25,6 +31,8 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
     // visible as they may signal a real problem (e.g. FORMERR 1 = malformed query);
     // NXDOMAIN (3) never reaches here (dns-connect returns it as an empty result).
     'Received error status from DNS server: 2.',
+    // viem's TimeoutError, surfaced instead of ethers' code: 'TIMEOUT'.
+    'The request timed out.',
     ...(additionalMessages || [])
   ];
   const codes = [
@@ -33,16 +41,18 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
     error.code,
     error.status,
     error.response?.status,
-    error.cause?.code
+    error.cause?.code,
+    root?.code
   ];
   return (
     messages.some(
       m =>
         error.message?.includes(m) ||
         error.error?.message?.includes(m) ||
-        error.cause?.message?.includes(m)
+        error.cause?.message?.includes(m) ||
+        root?.message?.includes(m)
     ) ||
-    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET', 504, 429].some(c =>
+    ['TIMEOUT', 'ECONNABORTED', 'ETIMEDOUT', 'ECONNRESET', 'ECONNREFUSED', 504, 429].some(c =>
       codes.some(v => String(v ?? '').includes(String(c)))
     )
   );
```

src/resolvers/address/ens.ts
```diff
@@ -32,35 +32,17 @@ function normalizeHandles(names: Handle[]): Handle[] {
...
-  const missing = normalizedAddresses.map((_, i) => i).filter(i => !validNames[i]);
-  const lookups = await Promise.allSettled(
-    missing.map(idx => provider.lookupAddress(normalizedAddresses[idx]))
-  );
-  const fallbackNames = normalizeEns(lookups.map(r => (r.status === 'fulfilled' && r.value) || ''));
-  missing.forEach((idx, j) => {
-    if (fallbackNames[j]) validNames[idx] = fallbackNames[j];
-  });
+
+  const validNames = normalizeEns(normalizedAddresses.map(address => values[address] || ''));
```
wa0x6e: `Object.keys(values).length === 0` isn't "every lookup failed" — `values` only holds addresses that *have* a name. Two consequences: a batch where nobody has an ENS name plus one transient blip takes the throw path unnecessarily (noise-level — output is `{}` either way), and, worse, a batch where one address resolves silently discards every other failure: `errors` is dropped, the affected addresses fall through to `''`, and `cacheEmpty: true` pins "no ENS name" for them for 12h (ttl 43200) with nothing in Sentry. The PR's own unit test "keeps a successful sibling when an offchain gateway request rejects" produces exactly that shape — 1 success + rejections → nothing reported.

The old CCIP fallback was equally silent on per-address rejections, but the legacy `getNames()` batch errors used to throw reliably; this folds them into the silent bucket whenever any single address succeeds.

Suggestion: make the predicate a true "every lookup failed" and run surviving partial failures through the repo's capture-and-continue precedent (`unstoppableDomains.ts`'s `resolveNames`). Verified locally: `tsc --noEmit` clean, unit suite passes with this applied.

```suggestion
  if (errors.length === normalizedAddresses.length) {
    throw errors[0];
  }

  // Partial failures: index.ts only sees the resolved values below, so once at
  // least one lookup succeeds any remaining errors would otherwise vanish silently.
  errors.forEach(err => {
    if (!isSilencedError(err)) {
      capture(err, { input: { addresses: normalizedAddresses } });
    }
  });

  const validNames = normalizeEns(normalizedAddresses.map(address => values[address] || ''));
```

src/resolvers/address/universalResolver.ts
```diff
@@ -0,0 +1,29 @@
+import { createPublicClient, http, Address as ViemAddress } from 'viem';
+import { mainnet } from 'viem/chains';
+
+const rpcUrl = `${process.env.BROVIDER_URL || 'https://rpc.snapshot.org'}/1`;
+const client = createPublicClient({
+  chain: mainnet,
+  batch: { multicall: { batchSize: 64 * 1024, wait: 0 } },
```
wa0x6e: `batchSize: 64 * 1024` is a byte budget over the scheduler's pending calldata, and the scheduler is keyed by client + block tag only (viem `call.ts:463`) — with this module-level client and a `0` wait (a `setTimeout(0)` macrotask, not a same-tick barrier), **concurrent** `lookupAddresses` requests merge into one `aggregate3`. Reproduced live: 5 independent 50-address requests produced 2 `eth_call`s, one carrying 224 calls spanning 4+ logical requests. At ~324 bytes per `reverseWithGateways` call, 64KB only chunks at ~202 calls — so with `retryCount: 0` and the 5s timeout, one transient failure of a merged call takes down several unrelated requests at once (a 200-call batch did succeed against the production RPC, so this is blast-radius sizing, not a hard cap being hit).

Suggestion halves the merge ceiling (~202 → ~101 calls) while keeping 2× margin over the 50-address requirement — the pinned "50 in one eth_call" test still passes — and documents the floor/ceiling coupling with `MAX_LOOKUP_ADDRESSES`. It also drops `wait: 0`, which is viem's documented default (`createClient.ts`: "@default 0") — dead config:

```suggestion
// batchSize must stay well above the 50-address MAX_LOOKUP_ADDRESSES floor (so one
// lookupAddresses call always fits in a single eth_call) but well below viem's chunking
// point for this call shape (~200 calls at ~324 bytes each). The scheduler batches by
// client + block tag only, so concurrent lookupAddresses requests can share one
// aggregate3 batch; capping batchSize keeps that merged batch from growing unbounded.
const client = createPublicClient({
  chain: mainnet,
  batch: { multicall: { batchSize: 32 * 1024 } },
```

(Overlaps the CCIP-timeout suggestion on lines 1–8 above — if you apply both, the merge is trivial: keep that suggestion's imports/`ccipRead` block and this one's comment + `batchSize` line.)

test/unit/resolvers/address/universalResolver.test.ts
```diff
@@ -0,0 +1,181 @@
...
+} from 'viem';
+import { reverseLookup } from '../../../../src/resolvers/address/universalResolver';
+
+const reverseAbi = parseAbi([
+  'function reverseWithGateways(bytes reverseName, uint256 coinType, string[] gateways) view returns (string resolvedName, address resolver, address reverseResolver)'
+]);
+const UNIVERSAL_RESOLVER = '0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe' as Address;
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000' as Address;
+const ADDRESSES = Array.from(
+  { length: 50 },
```
wa0x6e: Minor: `{ length: 50 }` hardcodes `MAX_LOOKUP_ADDRESSES` — if the cap in `resolvers/address/index.ts` changes, this "maximum address batch" test silently tests something else. The constant is exported and already imported by a sibling unit test (`starknet-batch-guard.test.ts`), so no cycle / new pattern (two disjoint lines, hence a plain comment):

```diff
--- a/test/unit/resolvers/address/universalResolver.test.ts
+++ b/test/unit/resolvers/address/universalResolver.test.ts
+import { MAX_LOOKUP_ADDRESSES } from '../../../../src/resolvers/address';
 import { reverseLookup } from '../../../../src/resolvers/address/universalResolver';
 ...
 const ADDRESSES = Array.from(
-  { length: 50 },
+  { length: MAX_LOOKUP_ADDRESSES },
   (_, index) => `0x${(index + 1).toString(16).padStart(40, '0')}` as Address
 );
```

src/resolvers/address/universalResolver.ts
```diff
@@ -0,0 +1,29 @@
+import { createPublicClient, http, Address as ViemAddress } from 'viem';
+import { mainnet } from 'viem/chains';
+
+const rpcUrl = `${process.env.BROVIDER_URL || 'https://rpc.snapshot.org'}/1`;
```
wa0x6e: Minor dedup: this line re-derives the same `BROVIDER_URL` fallback that `src/helpers/provider.ts:5` already computes, and the hardcoded `/1` duplicates the network id — two sources of truth for the RPC endpoint; changing the default in one place silently diverges. `helpers/provider.ts` imports nothing from `resolvers/`, so this is cycle-free (the fix spans a file outside this PR's diff, hence a plain comment):

```diff
--- a/src/helpers/provider.ts
+++ b/src/helpers/provider.ts
-const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';
+export const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';

--- a/src/resolvers/address/universalResolver.ts
+++ b/src/resolvers/address/universalResolver.ts
 import { mainnet } from 'viem/chains';
+import { broviderUrl } from '../../helpers/provider';

-const rpcUrl = `${process.env.BROVIDER_URL || 'https://rpc.snapshot.org'}/1`;
+const rpcUrl = `${broviderUrl}/${mainnet.id}`;
```

src/resolvers/address/universalResolver.ts
tony8713: Confirmed. The RPC timeout did not reach gateway fetches. The client now runs the CCIP request under withDeadline while preserving request options and composing any existing signal. That deadline covers viem’s full URL sequence, rather than granting a fresh budget to each URL; the stalled-gateway test reaches AbortError and keeps a successful sibling.

src/resolvers/address/ens.ts
tony8713: Confirmed. I used cycle-safe traversal across error and cause nodes instead of inspecting only the innermost walk() result. Deep viem timeouts and refused connections are muted, while wrapped 401/403 and contract failures remain visible; the tests cover both sides.

src/resolvers/address/ens.ts
tony8713: Confirmed, and fixed the cache consequence as well. All-failed now uses the rejection count and prefers a reportable error when failures are mixed. Partial failures capture only reportable errors with provider/input context, and rejected addresses are carried as non-cacheable while successful and genuinely empty siblings remain cacheable.

src/resolvers/address/universalResolver.ts
tony8713: Confirmed the shared scheduler behavior, but the current calldata is 292 bytes, so 64 KiB admits 224 calls and 32 KiB would still admit about 112. I used 16 KiB instead: one 50-address request remains one eth_call, while concurrent maximum requests split with a 56-call ceiling. The redundant wait: 0 is gone.

test/unit/resolvers/address/universalResolver.test.ts
tony8713: Applied. The maximum-batch test now uses MAX_LOOKUP_ADDRESSES. The referenced Starknet test does not currently import that constant, but this barrel import is cycle-free and keeps the test coupled to the public cap.

src/resolvers/address/universalResolver.ts
tony8713: Applied. provider.ts now exports the canonical broviderUrl, and the viem endpoint derives its network suffix from mainnet.id, with no resolver import cycle.

src/resolvers/address/universalResolver.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const EMPTY_REVERSE_ERRORS = new Set([
+  'ResolverError',
+  'ResolverNotContract',
+  'ResolverNotFound',
+  'ReverseAddressMismatch',
+  'UnsupportedResolverProfile'
+]);
+
+function isEmptyReverseError(error: unknown): boolean {
```
wa0x6e, line 30: The deadline fix holds, but the "AbortError is already special-cased by `isSilencedError`" part (my earlier comment, and the `root.name === 'AbortError'` assertion in the unit test) does not hold on the production path.

`getEnsName` passes `gatewayUrls = ['x-batch-gateway:true']`, so the UR's `OffchainLookup` goes through viem's `localBatchGatewayRequest`, which *catches* every sub-fetch error and re-encodes it — `HttpError(status, msg)` for HTTP non-2xx, plain `Error(msg)` for everything else (`utils/ens/localBatchGatewayRequest.js`, `encodeError`). UR's `resolveBatchCallback` then reverts that data raw (`FLAG_BATCH_ERROR`). The unit test only reaches a real `AbortError` because its fixture reverts with a direct `https://gateway.test/{data}` URL.

Live probe against mainnet with `0x2211…7DA9` (jesse.base.eth, Coinbase gateway), overriding `fetch` for the gateway URL only:

| gateway behaviour | surfaced revert (`ContractFunctionRevertedError.data`) | `isSilencedError` |
|---|---|---|
| hang → our 10 s abort | `Error("This operation was aborted")` | **false** |
| HTTP 503 | `HttpError(503, "HTTP request failed.")` | **false** |
| `TypeError('fetch failed')` | `Error("HTTP request failed.")` | **false** |

Cacheability is right (all three land in `errors` → non-cacheable), but every transient gateway blip for a `*.base.eth` / `*.cb.id` holder is now a Sentry event — per address through the `forEach` in `ens.ts`, or through the all-failed throw for a single-address request. The old fallback swallowed these silently, so this is new noise rather than new data loss. A batch-gateway classifier next to `isEmptyReverseError` (treat `HttpError` with 5xx/429, and `Error("This operation was aborted" | "HTTP request failed.")` as transient) would keep those muted while leaving `Error(...)` from a genuinely broken resolver loud. The unit test should also get a fixture that goes through `x-batch-gateway:true` so it models the real path.

src/resolvers/address/cache.ts
```diff
@@ -24,7 +34,11 @@ export default async function cache(input: string[], callback) {
 
   if (uncachedInputs.length > 0) {
     const results = await callback(uncachedInputs);
-    setCache(results);
+    const nonCacheable = new Set(results[NON_CACHEABLE] || []);
+    const cacheable = Object.fromEntries(
+      Object.entries(results).filter(([key]) => !nonCacheable.has(key))
+    );
+    await setCache(cacheable);
```
wa0x6e: Now awaited where master fired-and-forgot (`setCache(results);`). A rejected `multi().exec()` (`src/cache/stores/redis.ts:47` has no catch) fails the whole request after every resolver already succeeded, and every response pays one extra Redis RTT. Awaiting was the right instinct (the old floating promise was an unhandled-rejection risk), but the cache is an optimisation — a failed write should not fail the response:

```diff
--- a/src/resolvers/address/cache.ts
+++ b/src/resolvers/address/cache.ts
+import { capture } from '@snapshot-labs/snapshot-sentry';
 import { RedisStore } from '../../cache';
 ...
-    await setCache(cacheable);
+    // The cache is an optimisation; a failed write must not fail the response.
+    setCache(cacheable).catch(err => capture(err));
```

Not for this PR: `getCache` on line 29 has the same shape (a rejecting read fails the request before any lookup), and `helpers/redis.ts` uses node-redis' default offline queue, so a *disconnected* Redis hangs commands until reconnect rather than rejecting — `.catch` never fires. True fail-open needs `disableOfflineQueue: true` on the client plus a `.catch` on the read; that touches every Redis user in the process, so it's a separate change.

wa0x6e (commented):
Re-verified all six earlier threads from scratch (code + viem 2.55.13 source + live mainnet probes, not the replies): each has a real fix — transport timeout/ECONNREFUSED now silenced, 401/403 stay loud, all-failed predicate is correct, failed addresses are non-cacheable, batchSize math checks out (292 B/call → 56 per eth_call), constants deduped. `tsc`, `eslint`, 139 unit/helper tests pass.

Two new things below, one of them on the production CCIP path that the unit tests don't model.

wa0x6e (changes requested):
see comments

src/resolvers/address/universalResolver.ts
tony8713: Confirmed on the live jesse.base.eth reverse path: a forced Coinbase 503 surfaced as HttpError(503, "HTTP request failed.") and a stalled request aborted after the 10s deadline as Error("This operation was aborted"), and neither was silenced. The classifier now requires an OffchainLookupError in the cause chain and mutes only 429/5xx or viem’s two exact transient messages, while a direct Universal Resolver HttpError stays actionable. The tests now route through x-batch-gateway:true and verify the locally encoded callback data for both deadline and HTTP failures.

src/resolvers/address/cache.ts
tony8713: Confirmed by forcing RedisStore.setMany to reject: the existing await rejected an otherwise successful lookup response. The write now runs without delaying the response and attaches catch(capture); a controlled pending-write test proves the result settles first and the later rejection is reported. I left the read and offline-queue scope unchanged.

PR #569, opened 2026-08-19
refactor(ens): read avatar through universal resolver

### LINKED ISSUE #560: refactor(ens): read the avatar text record through the Universal Resolver client (3/3)
Part 3 of 3, optional. Avatar text record through the same client as parts 1 and 2, so ENS has one CCIP-aware path.

## Today

`src/resolvers/image/ens.ts`: `provider.getResolver(name)` then `resolver.getText('avatar')` — two ethers RPCs, plus the `metadata.ens.domains` fallback URL.

## Change

`getTextRecord(name, 'avatar')` on the viem client (`getEnsText`, from #483) — one Universal Resolver call, CCIP handled. Keep the `metadata.ens.domains` fallback.

Check first whether `@snapshot-labs/snapshot.js` ≥0.16 exposes an ENS text-record helper (the #483 close comment says it took over the avatar record). If it does and it is CCIP-aware, use that and skip the local client for this one.

## Acceptance

- Existing image integration fixtures unchanged.
- Fewer RPCs per avatar resolution than today; assert with a mocked transport if the viem path is taken.

Only reason to do this: single client, one path. Not needed for the RPC savings in parts 1 and 2.

tony8713:
## Summary

ENS avatar resolution previously discovered an ethers resolver before reading its text record, leaving image lookup on a separate two-call path.

Use snapshot.js' CCIP-aware Universal Resolver helper for the avatar record. ENS-shaped inputs are normalized before lookup, the fallback name remains one encoded path segment, and the existing metadata fallback remains. Stamp's provider helper keeps its 5 second default while accepting a numeric timeout override.

The Universal Resolver result does not distinguish a missing resolver from a missing avatar record, so names without a resolver now also reach the metadata fallback. This extra request is accepted in exchange for removing resolver discovery and keeping one Universal Resolver path.

Closes #560

### DIFF
--- a/src/helpers/provider.ts
+++ b/src/helpers/provider.ts
@@ -2,10 +2,20 @@ import { StaticJsonRpcProvider } from '@ethersproject/providers';
 import snapshot from '@snapshot-labs/snapshot.js';
 import { Address } from './types';
 
-const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';
+const DEFAULT_PROVIDER_TIMEOUT = 5e3;
 
-export function getProvider(network: string | number) {
-  return snapshot.utils.getProvider(network, { broviderUrl, timeout: 5e3 });
+export function getProviderOptions(timeout = DEFAULT_PROVIDER_TIMEOUT): {
+  broviderUrl: string;
+  timeout: number;
+} {
+  return {
+    broviderUrl: process.env.BROVIDER_URL || 'https://rpc.snapshot.org',
+    timeout
+  };
+}
+
+export function getProvider(network: string | number, timeout = DEFAULT_PROVIDER_TIMEOUT) {
+  return snapshot.utils.getProvider(network, getProviderOptions(timeout));
 }
 
 /**
--- a/src/resolvers/image/ens.ts
+++ b/src/resolvers/image/ens.ts
@@ -1,8 +1,8 @@
 import { ens_normalize } from '@adraffy/ens-normalize';
 import { isAddress } from '@ethersproject/address';
-import { StaticJsonRpcProvider } from '@ethersproject/providers';
+import snapshot from '@snapshot-labs/snapshot.js';
 import { fetchHttpImage } from '../../helpers/http';
-import { getProvider } from '../../helpers/provider';
+import { getProviderOptions } from '../../helpers/provider';
 import { lookupAddresses } from '../address';
 
 async function castToEnsName(nameOrAddress: string): Promise<string | undefined> {
@@ -20,20 +20,14 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
 }
 
 export default async function resolve(nameOrAddress: string) {
-  // getProvider returns `any`. Without the annotation the calls below are unchecked.
-  const provider: StaticJsonRpcProvider = getProvider(1);
   const ensName = await castToEnsName(nameOrAddress);
 
   if (!ensName) return false;
 
-  const ensResolver = await provider.getResolver(ensName);
-
-  if (!ensResolver) {
-    return false;
-  }
-
-  let url = await ensResolver.getText('avatar');
-  url = url?.startsWith('http') ? url : `https://metadata.ens.domains/mainnet/avatar/${ensName}`;
+  let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
+  url = url?.startsWith('http')
+    ? url
+    : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;
 
   return await fetchHttpImage(url);
 }
--- a/test/unit/resolvers/image/ens.test.ts
+++ b/test/unit/resolvers/image/ens.test.ts
@@ -1,41 +1,109 @@
-import { getProvider } from '../../../../src/helpers/provider';
+import snapshot from '@snapshot-labs/snapshot.js';
+import { fetchHttpImage } from '../../../../src/helpers/http';
 import resolve from '../../../../src/resolvers/image/ens';
 
-jest.mock('../../../../src/helpers/provider', () => ({ getProvider: jest.fn() }));
+jest.mock('../../../../src/helpers/http', () => ({ fetchHttpImage: jest.fn() }));
 jest.mock('../../../../src/resolvers/address', () => ({ lookupAddresses: jest.fn() }));
 
 const STARKNET_ADDRESS = '0x0779ba6e4e227947acbbdfb978a292c401339027eeb3d768f5d12cd2e818265a';
-const INVALID_NAMES = [STARKNET_ADDRESS, 'nodot', 'a..b'];
+const INVALID_NAMES = [STARKNET_ADDRESS, 'nodot', 'a..b', '../avatar/vitalik.eth'];
 const VALID_NAMES = [
   ['vitalik.eth', 'vitalik.eth'],
   ['foo.xyz', 'foo.xyz'],
   ['ⓥⓘⓣⓐⓛⓘⓚ.eth', 'vitalik.eth']
 ] as const;
 
-const getResolver = jest.fn();
+const mockTransportRequest = jest.fn();
 
-beforeEach(() => {
-  getResolver.mockReset();
-  (getProvider as jest.Mock).mockReturnValue({ getResolver });
+jest.mock('viem', () => {
+  const viem = jest.requireActual('viem');
+  return {
+    ...viem,
+    http: jest.fn(() => viem.custom({ request: mockTransportRequest }))
+  };
 });
 
-describe('ens image resolver', () => {
+const mockedFetchHttpImage = jest.mocked(fetchHttpImage);
+const originalBroviderUrl = process.env.BROVIDER_URL;
+
+beforeEach(() => {
+  process.env.BROVIDER_URL = 'https://custom.rpc';
+  mockTransportRequest.mockReset();
+  mockedFetchHttpImage.mockReset();
+});
+
+afterEach(() => {
+  jest.restoreAllMocks();
+  if (originalBroviderUrl === undefined) delete process.env.BROVIDER_URL;
+  else process.env.BROVIDER_URL = originalBroviderUrl;
+});
+
+describe('resolvers/image/ens', () => {
   it.each(INVALID_NAMES)('skips the provider for %s', async input => {
+    const getEnsTextRecord = jest.spyOn(snapshot.utils, 'getEnsTextRecord');
+
     await expect(resolve(input)).resolves.toBe(false);
-    expect(getResolver).not.toHaveBeenCalled();
+    expect(getEnsTextRecord).not.toHaveBeenCalled();
+    expect(mockTransportRequest).not.toHaveBeenCalled();
+    expect(mockedFetchHttpImage).not.toHaveBeenCalled();
   });
 
   it.each(VALID_NAMES)('asks the provider for %s', async (input, normalized) => {
-    getResolver.mockResolvedValue(null);
+    const getEnsTextRecord = jest.spyOn(snapshot.utils, 'getEnsTextRecord').mockResolvedValue(null);
+    const image = Buffer.from('avatar');
+    mockedFetchHttpImage.mockResolvedValue(image);
 
-    await expect(resolve(input)).resolves.toBe(false);
-    expect(getResolver).toHaveBeenCalledWith(normalized);
+    await expect(resolve(input)).resolves.toBe(image);
+    expect(getEnsTextRecord).toHaveBeenCalledWith(normalized, 'avatar', '1', {
+      broviderUrl: 'https://custom.rpc',
+      timeout: 5e3
+    });
+  });
+
+  it('keeps a normalized name in one fallback path segment', async () => {
+    const image = Buffer.from('avatar');
+    jest.spyOn(snapshot.utils, 'getEnsTextRecord').mockResolvedValue(null);
+    mockedFetchHttpImage.mockResolvedValue(image);
+
+    await expect(resolve('🌈.eth')).resolves.toBe(image);
+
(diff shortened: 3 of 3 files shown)


src/resolvers/image/ens.ts
```diff
@@ -13,19 +13,11 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
...
 
-  const ensResolver = await provider.getResolver(ensName);
-
-  if (!ensResolver) {
-    return false;
-  }
-
-  let url = await ensResolver.getText('avatar');
+  let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
   url = url?.startsWith('http') ? url : `https://metadata.ens.domains/mainnet/avatar/${ensName}`;
```
wa0x6e: Dropping the `if (!ensResolver) return false` guard changes what happens to names that fail ENS normalization. Previously `provider.getResolver(name)` returned `null` for them (ethers' `namehash` throws inside, and `_getResolver` swallows it), so `resolve()` returned `false`. `snapshot.utils.getEnsTextRecord` also swallows the `ensNormalize` throw and returns `null`, but that is indistinguishable from "no avatar record" here, so the raw string now reaches the fallback URL uninterpolated-escaped:

`GET /avatar/..%2Favatar%2Fvitalik.eth` → Express decodes the route param to `../avatar/vitalik.eth`, `parseQuery` only lowercases it, `ens` runs on every `/avatar/:id` (default resolver list in `src/constants.json`), `ensNormalize` throws `Disallowed codepoint: 0x2F` → `null` → the request goes to `https://metadata.ens.domains/mainnet/avatar/../avatar/vitalik.eth`, which `fetch`'s URL parser collapses to `/mainnet/avatar/vitalik.eth` (200 image/jpeg). The host is fixed, so this is caller-controlled path selection on a trusted third party rather than SSRF — but it is unauthenticated and needs no special params. `0x3F` (`?`) and `0x23` (`#`) are rejected by normalization too, so they cannot reach the query/fragment either once the guard is back.

Normalizing up front restores the old rejection and lets the fallback be built from the normalized name (`encodeURIComponent` keeps it a single path segment). `ensNormalize` comes from `@ethersproject/hash`, already a direct dependency (`^5.8.0`), and is the same function snapshot.js uses internally, so there is no second RPC round-trip and no revision skew.

This also requires one added import at the top of the file:

```ts
import { ensNormalize } from '@ethersproject/hash';
```

```suggestion
  let normalizedName: string;
  try {
    normalizedName = ensNormalize(ensName);
  } catch {
    return false;
  }

  let url = await snapshot.utils.getEnsTextRecord(
    normalizedName,
    'avatar',
    '1',
    getProviderOptions()
  );
  url = url?.startsWith('http')
    ? url
    : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(normalizedName)}`;
```

Verified locally: `tsc --noEmit`, `eslint`, `prettier --check` and the new unit test all pass with this applied; `ensNormalize` leaves `example.eth` / `nick.eth` untouched, lowercases `VITALIK.eth`, keeps `🌈.eth` (percent-encoded in the URL, as `fetch` would have done anyway), and rejects the `/`, `?`, `#` cases.

Worth adding as a regression test — verified red without the fix, green with it:

```ts
  it('returns false for a name that fails ENS normalization', async () => {
    await expect(resolve('../avatar/vitalik.eth')).resolves.toBe(false);

    expect(mockTransportRequest).not.toHaveBeenCalled();
    expect(mockedFetchHttpImage).not.toHaveBeenCalled();
  });
```

src/resolvers/image/ens.ts
```diff
@@ -13,19 +13,11 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
...
   if (!ensName) return false;
 
-  const ensResolver = await provider.getResolver(ensName);
-
-  if (!ensResolver) {
-    return false;
-  }
-
-  let url = await ensResolver.getText('avatar');
+  let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
```
wa0x6e, line 27: `getEnsTextRecord` returns `null` both when a name has no resolver and when its resolver has no avatar record — the library's own comment is explicit about it ("unresolvable names and resolver-level failures resolve to null"). The removed `if (!ensResolver) return false` used to short-circuit the first case before any network call; now both fall through to the `metadata.ens.domains` fetch.

The final answer is unchanged (the 404 throws, `withFailureContract` catches it silently, `false` comes back), so this is a cost item rather than a correctness one. It is worth calling out explicitly because `ens` runs on **every** `/avatar/:id` base-cache-miss in parallel with the other five resolvers (`src/constants.json`, `src/api.ts`), not only after they fail — and the same route accepts Farcaster/Lens/Starknet/space identifiers that never had an ENS resolver to begin with. Those all now spend a fallback request that they previously skipped for free.

For scale, from Sentry (last 7 days): 94,550 `/avatar/:id` requests, and 4,670 outbound `metadata.ens.domains` calls already under the current code at 1.1–1.9s each. That 4,670 is the untouched "resolver exists, no http avatar" case; the added no-resolver share is not measurable from existing telemetry, since that path emits no span or log today (by design — the silent catch in `withFailureContract`). A canary comparing `metadata.ens.domains` span counts before/after would give the real delta.

No code change is being proposed — distinguishing the two `null` cases again means a second RPC round-trip, which is exactly what this PR set out to remove. But since issue #560's acceptance criterion is "fewer RPCs per avatar resolution," the matching tradeoff belongs in the PR description, e.g.:

> Note: because the Universal Resolver call cannot distinguish "no resolver" from "resolver with no avatar record," names with no ENS resolver at all now also hit the `metadata.ens.domains` fallback (previously short-circuited for free) — an accepted tradeoff for dropping the second RPC call.

src/helpers/provider.ts
```diff
@@ -2,10 +2,15 @@ import { StaticJsonRpcProvider } from '@ethersproject/providers';
 import snapshot from '@snapshot-labs/snapshot.js';
 import { Address } from './types';
 
-const broviderUrl = process.env.BROVIDER_URL || 'https://rpc.snapshot.org';
+export function getProviderOptions() {
```
wa0x6e: `getEnsTextRecord`'s options parameter is typed `any` (`@snapshot-labs/snapshot.js/dist/src/utils/ens.d.ts`: `options?: any`), and snapshot.js's `normalizeOptions` reads only `broviderUrl` and `timeout`, silently defaulting to `https://rpc.snapshot.org` / 25s for anything it doesn't recognise. So a typo here — `broviderURL`, say — would route production traffic off `BROVIDER_URL` with no compile error and no runtime signal.

An explicit return type closes that hole at the definition site, where it also covers the existing `getProvider` caller:

```suggestion
export function getProviderOptions(): { broviderUrl: string; timeout: number } {
```

Verified with `tsc --strict`: with the annotation, a `broviderURL` typo fails as `TS2561: Object literal may only specify known properties, but 'broviderURL' does not exist in type '{ broviderUrl: string; timeout: number; }'. Did you mean to write 'broviderUrl'?` — without it, the same typo typechecks clean today.

wa0x6e (changes requested):

src/resolvers/image/ens.ts
tony8713: Rebased onto master, which already normalizes before lookup. I kept that guard, encoded the fallback segment, and added the reported ../avatar/vitalik.eth case.

src/resolvers/image/ens.ts
tony8713: The PR description now records the extra fallback request for no-resolver names as the tradeoff for dropping resolver discovery.

src/helpers/provider.ts
tony8713: getProviderOptions now has the explicit return type. getProviderOptions and getProvider also accept the requested numeric timeout override.

src/resolvers/image/ens.ts
```diff
@@ -20,20 +20,14 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
...
 
-  const ensResolver = await provider.getResolver(ensName);
-
-  if (!ensResolver) {
-    return false;
-  }
-
-  let url = await ensResolver.getText('avatar');
-  url = url?.startsWith('http') ? url : `https://metadata.ens.domains/mainnet/avatar/${ensName}`;
+  let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
```
wa0x6e, line 27: Follow-up on the fallback hardening: the `ens_normalize` guard in `castToEnsName` only runs on the branch where the caller typed a name. The address branch returns at the `isAddress` check, so whatever `lookupAddresses` answers with reaches `getEnsTextRecord` and the fallback URL unvalidated.

That value is not necessarily an ENS name. The aggregator (`src/resolvers/address/index.ts`) takes the first non-empty answer across nine resolvers with `Snapshot` first, and the Snapshot resolver returns `user.name` straight from the hub. The profile schema in snapshot.js types `name` as `{ type: string, maxLength: 32 }` with no `pattern`, so any 32 characters are allowed. `normalizeEns` inside `resolvers/address/ens.ts` never sees these — it only filters that resolver's own reverse records.

Sampling the 862 most recent hub users that have a display name set:

| shape | share | examples |
|---|---|---|
| no `.` at all | 83.4% | `Kamruzzaman`, `Maria`, `KG` |
| has `.`, fails `ens_normalize` | 0.5% | `CLT #1`, `%%LEVE7`, `Mr. Cee` |
| valid after normalization | 16.1% | `jkdrq.eth`, `hawas7.base.eth` |

So for a large share of `/avatar/<address>` cache misses, the resolver currently spends a Universal Resolver `eth_call` plus a `metadata.ens.domains` request on a string that cannot be an ENS name. Both are guaranteed to fail — `metadata.ens.domains` answers `400 {"details":["Name contains dangerous or invalid characters"]}` for the `%`/`#` cases and `404` for the rest (verified against the live service). Validating before the lookup returns early and skips both.

It also puts the guard on the boundary rather than only on the sink: `encodeURIComponent` keeps a hostile display name such as `../../foo` inside one path segment, but with the check applied to both branches that name never reaches the URL builder at all.

```diff
 async function castToEnsName(nameOrAddress: string): Promise<string | undefined> {
-  if (isAddress(nameOrAddress)) {
-    return (await lookupAddresses([nameOrAddress]))[nameOrAddress];
-  }
+  const name = isAddress(nameOrAddress)
+    ? (await lookupAddresses([nameOrAddress]))[nameOrAddress]
+    : nameOrAddress;
 
-  if (!nameOrAddress.includes('.')) return undefined;
+  // A reverse lookup answers with any handle the address resolvers know, including
+  // Snapshot display names, so it needs the same check as a name typed by the caller.
+  if (!name?.includes('.')) return undefined;
 
   try {
-    return ens_normalize(nameOrAddress);
+    return ens_normalize(name);
   } catch {
     return undefined;
   }
 }
```

Note the `?.` — `lookupAddresses` returns no entry for an address it cannot resolve, so `name` is `undefined` there, which the old address branch passed through to the `!ensName` check.

Covering tests, verified red against the current head (5 failures) and green with the change, alongside the existing 189:

```ts
const ADDRESS = '0x1234567890123456789012345678901234567890';

// Reverse lookups answer with any handle the address resolvers know; a Snapshot
// display name has no character rules beyond a 32 character limit.
const NON_ENS_REVERSE_RECORDS = ['Kamruzzaman', 'CLT #1', '%%LEVE7', '../../foo'];

it.each(NON_ENS_REVERSE_RECORDS)('skips the provider for the reverse record %s', async name => {
  mockedLookupAddresses.mockResolvedValueOnce({ [ADDRESS]: name });
  const getEnsTextRecord = jest.spyOn(snapshot.utils, 'getEnsTextRecord');

  await expect(resolve(ADDRESS)).resolves.toBe(false);
  expect(getEnsTextRecord).not.toHaveBeenCalled();
  expect(mockTransportRequest).not.toHaveBeenCalled();
  expect(mockedFetchHttpImage).not.toHaveBeenCalled();
});

it('normalizes a reverse record before the text lookup', async () => {
  mockedLookupAddresses.mockResolvedValueOnce({ [ADDRESS]: 'VITALIK.eth' });
  const getEnsTextRecord = jest.spyOn(snapshot.utils, 'getEnsTextRecord').mockResolvedValue(null);
  const image = Buffer.from('avatar');
  mockedFetchHttpImage.mockResolvedValue(image);

  await expect(resolve(ADDRESS)).resolves.toBe(image);
  expect(getEnsTextRecord).toHaveBeenCalledWith('vitalik.eth', 'avatar', '1', expect.anything());
  expect(mockedFetchHttpImage).toHaveBeenCalledWith(
    'https://metadata.ens.domains/mainnet/avatar/vitalik.eth'
  );
});
```

They need `import { lookupAddresses } from '../../../../src/resolvers/address';` and `const mockedLookupAddresses = jest.mocked(lookupAddresses);` — the module is already mocked in this file.

Two things this deliberately does not fix. Handles from the non-ENS resolvers (`foo.lens`, `foo.crypto`, `foo.stark`) still pass, since they are syntactically valid names and filtering by TLD would break DNS-imported ENS names. And the no-resolver case from the PR description is unchanged — that one genuinely needs a second RPC to detect.

Since `castToEnsName` is untouched by this PR, this is a reasonable follow-up rather than a merge blocker.

wa0x6e (changes requested):
see comment

src/resolvers/image/ens.ts
tony8713: Agreed. This predates the PR change, so I am leaving the thread unresolved for follow-up rather than expanding #569.

wa0x6e (approved):
utAck

src/resolvers/image/ens.ts
tony8713: Tracked in #584 with the reproduced address path and focused acceptance criteria. Keeping #569 unchanged as requested.

PR #570, opened 2026-08-19
perf: batch Basename resolver calls

### LINKED ISSUE #561: perf(basename): batch name()/addr() through multicall
## Today

`src/resolvers/address/basename.ts`:
- `lookupAddresses` (up to 50 addresses per request): one `snapshot.utils.call(RESOLVER, 'name', [reverseNode])` **per address** on Base → 50 RPCs per uncached request (`collect()` at L42-55).
- `resolveNames` (≤5): one `addr(namehash)` call per handle.

Same contract, same function, only the argument varies — exactly what `batchContractCalls` (Multicaller) already does for gwei and Unstoppable Domains. Multicall is available on 8453 (`0xca11…76ca11` in snapshot.js `networks.json`).

## Change

- `lookupAddresses`: `batchContractCalls(NETWORK, provider, ABI, nodes, fill(RESOLVER), 'name')`, then map results back by node → address, `normalizeBasename` each. One RPC.
- `resolveNames`: same with `addr` (one RPC), keep the `EMPTY_ADDRESS` filter + `getAddress`.
- `getAvatar` unchanged (single-item path: `name()` then `text()`, sequential by nature).
- `collect()` goes away if nothing else uses it.

Note: `snapshot.utils.multicall` is `aggregate` (all-or-nothing). `name()`/`addr()` on the Basenames L2 resolver return empty/zero for unknown nodes rather than reverting, so this is safe — verify once against mainnet with a mixed batch (known name + random address) before relying on it.

## Acceptance

- Unit test with mocked provider: N addresses → **one** `eth_call`.
- Existing integration fixtures unchanged (`test/integration/resolvers/address/basename.test.ts`).
- Mixed batch: one address with a Basename, one without → first resolves, second absent, no throw.

tony8713:
## Summary

Basename reverse and forward lookup previously issued one `eth_call` per valid input. For a non-empty resolver batch of N inputs, that meant N requests.

The resolver now batches those reads through Base Multicall, reducing the same batch to 1 `eth_call` and saving N - 1 requests. For example, 2 inputs go from 2 requests to 1. Input validation, name normalization, and empty-result filtering remain unchanged.

Closes #561

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "zod": "^3.25.76"
   },
   "devDependencies": {
+    "@ethersproject/abi": "^5.7.0",
     "@snapshot-labs/eslint-config": "^0.1.1",
     "@snapshot-labs/prettier-config": "^0.1.0",
     "@types/express": "^4.17.11",
--- a/src/resolvers/address/basename.ts
+++ b/src/resolvers/address/basename.ts
@@ -4,7 +4,7 @@ import { namehash } from '@ethersproject/hash';
 import snapshot from '@snapshot-labs/snapshot.js';
 import { EMPTY_ADDRESS, isEvmAddress } from '../../helpers/address';
 import { getUrl } from '../../helpers/http';
-import { getProvider } from '../../helpers/provider';
+import { batchContractCalls, getProvider } from '../../helpers/provider';
 import { Address, Handle } from '../../helpers/types';
 
 export const NAME = 'Basename';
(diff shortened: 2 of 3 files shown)


wa0x6e:
**RPC savings from this PR** (per uncached request):

- `lookupAddresses`: up to 50 addresses/request, previously one `eth_call` per address → now 1 multicall. Saves up to **49 RPCs per request** (N−1 for N valid addresses, ~98% at full batch).
- `resolveNames`: ≤5 handles/request → 1 multicall, saves up to 4.
- `getAvatar`: unchanged (single-item path, nothing to batch).

Latency measured live against Base (`rpc.snapshot.box/8453`, 3 runs each, mixed known/unknown nodes):

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| one 50-entry `aggregate` | 2070ms | 754ms | 696ms |
| 50 parallel `eth_call`s | 7791ms | 2099ms | 1368ms |

Batched call was faster in every run — the batching also removes the tail-latency contention of 50 concurrent connections hitting the same endpoint.

wa0x6e (approved):
utAck

PR #571, opened 2026-08-19
perf(lookupDomains): batch ENS registration lookups

### LINKED ISSUE #562: perf(lookupDomains): batch registration lookups for hashed labels
## Today

`src/resolvers/lookupDomains/ens.ts`: after the `account(id)` query, every domain whose name still carries a hashed label (`[<hash>].eth`, unindexed labelName) triggers its own `registration(id)` subgraph query in `fetchDomainData` (L14-41, fanned out at L81). Unbounded per address: an address holding many such names costs one subgraph query each, on the same request.

## Change

Collect the hashes first, one query:

```graphql
query Registrations($ids: [String!]!) {
  registrations(where: { id_in: $ids }) { id domain { labelName } }
}
```

then substitute `[hash]` → `labelName` in place. Zero hashes → no query at all (already the case per-domain, keep it).

Keep the strict `data.` access (see the comment at L33-34): a missing envelope must still throw rather than serve raw `[hash]` labels.

## Acceptance

- Unit test with mocked `graphQlCall`: address with K hashed-label domains → exactly **2** calls (`account` + `registrations`), not 1+K.
- Same output as today for the existing lookupDomains fixtures.
- Domain whose registration is missing keeps its `[hash]` name (current behaviour).

tony8713:
Batch ENS hashed-label registration resolution into one subgraph query sized for all requested IDs. Responses are joined by registration ID, while missing registrations preserve the hashed name and addresses without hashed labels skip the registration query.

Closes #562

### DIFF
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -14,9 +14,7 @@ type Domain = {
 
 type Registration = {
   id: string;
-  domain?: {
-    labelName?: string;
-  };
+  domain?: Pick<Domain, 'labelName'>;
 };
 
 function getLabelHash(domain: Domain) {
@@ -48,10 +46,7 @@ async function fetchDomainData(domains: Domain[], chainId: string): Promise<Doma
     }
   );
   const labelNames = new Map(
-    (data.registrations || []).map(registration => [
-      registration.id,
-      registration.domain?.labelName
-    ])
+    data.registrations.map(registration => [registration.id, registration.domain?.labelName])
   );
 
   return domains.map(domain => {
@@ -60,7 +55,7 @@ async function fetchDomainData(domains: Domain[], chainId: string): Promise<Doma
 
     return {
       ...domain,
-      name: labelName ? domain.name.replace(`[${hash}]`, labelName) : domain.name
+      name: labelName ? domain.name.replace(`[${hash}]`, () => labelName) : domain.name
     };
   });
 }
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -21,12 +21,12 @@ function getLabelHash(domain: Domain) {
   return domain.name.match(/\[(.*?)\]/)?.[1];
 }
 
-async function fetchDomainData(domains: Domain[], chainId: string): Promise<Domain[]> {
+async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Handle[]> {
   const hashes = [
     ...new Set(domains.map(getLabelHash).filter((hash): hash is string => Boolean(hash)))
   ];
 
-  if (!hashes.length) return domains;
+  if (!hashes.length) return domains.map(domain => domain.name);
 
   const {
     data: { data }
@@ -53,10 +53,7 @@ async function fetchDomainData(domains: Domain[], chainId: string): Promise<Doma
     const hash = getLabelHash(domain);
     const labelName = hash ? labelNames.get(`0x${hash}`) : undefined;
 
-    return {
-      ...domain,
-      name: labelName ? domain.name.replace(`[${hash}]`, () => labelName) : domain.name
-    };
+    return labelName ? domain.name.replace(`[${hash}]`, () => labelName) : domain.name;
   });
 }
 
@@ -97,5 +94,5 @@ export default async function lookupDomains(
       !domain.name.endsWith('.addr.reverse')
   );
 
-  return (await fetchDomainData(domains, chainId)).map(domain => domain.name);
+  return fetchDomainNames(domains, chainId);
 }
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -12,33 +12,57 @@ type Domain = {
   expiryDate?: number;
 };
 
-async function fetchDomainData(domain: Domain, chainId: string): Promise<Domain> {
-  const hash = domain.name.match(/\[(.*?)\]/)?.[1];
+type Registration = {
+  id: string;
+  domain?: {
+    labelName?: string;
+  };
+};
 
-  if (!hash) return domain;
+function getLabelHash(domain: Domain) {
+  return domain.name.match(/\[(.*?)\]/)?.[1];
+}
+
+async function fetchDomainData(domains: Domain[], chainId: string): Promise<Domain[]> {
+  const hashes = [
+    ...new Set(domains.map(getLabelHash).filter((hash): hash is string => Boolean(hash)))
+  ];
+
+  if (!hashes.length) return domains;
 
   const {
     data: { data }
-  } = await graphQlCall(
+  } = await graphQlCall<{ registrations: Registration[] }>(
     constants.ensSubgraph[chainId],
-    `query Registration($id: String!) {
-      registration(id: $id) {
+    `query Registrations($ids: [String!]!, $first: Int!) {
+      registrations(first: $first, where: { id_in: $ids }) {
+        id
         domain {
-          name
           labelName
         }
       }
     }`,
-    { id: `0x${hash}` }
+    {
+      ids: hashes.map(hash => `0x${hash}`),
+      first: hashes.length
+    }
   );
-  // Not `data?.`: swallowing a missing envelope here serves the raw `[hash]`
-  // label as though the subgraph had simply not indexed it.
-  const labelName = data.registration?.domain?.labelName;
+  const labelNames = new Map(
+    (data.registrations || []).map(registration => [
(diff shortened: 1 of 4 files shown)


src/resolvers/lookupDomains/ens.ts
```diff
@@ -12,33 +12,57 @@ type Domain = {
...
+      ids: hashes.map(hash => `0x${hash}`),
+      first: hashes.length
+    }
+  );
+  const labelNames = new Map(
+    (data.registrations || []).map(registration => [
+      registration.id,
+      registration.domain?.labelName
+    ])
   );
```
wa0x6e, line 50: `data.registrations || []` can't fire in practice — `registrations` is `[Registration!]!` on both configured subgraphs (introspected), `graphQlCall` already throws on a missing envelope / `errors[]`, and the declared generic makes the `[]` branch unreachable for TS. It does sit exactly where the deleted "not `data?.`" comment used to guard the must-throw invariant #562 asked to keep, so if a stub or a looser endpoint ever returns `null` here we'd silently serve `[hash]` names instead of failing loudly.

```suggestion
  // Not `|| []`: registrations is a non-null list per the subgraph schema, and
  // graphQlCall already throws on a missing/error envelope, so a fallback here
  // would only hide a real bug behind raw `[hash]` labels.
  const labelNames = new Map(
    data.registrations.map(registration => [
      registration.id,
      registration.domain?.labelName
    ])
  );
```

src/resolvers/lookupDomains/ens.ts
```diff
@@ -12,33 +12,57 @@ type Domain = {
...
+        id
         domain {
-          name
           labelName
         }
       }
     }`,
-    { id: `0x${hash}` }
+    {
+      ids: hashes.map(hash => `0x${hash}`),
```
wa0x6e, line 44: Nit: the `0x${hash}` registration id is derived here and again at L59 (`labelNames.get(`0x${hash}`)`). If one side ever gets normalized (lowercase, zero-pad) and the other doesn't, the map lookup just misses and every label silently falls back to `[hash]` — no error. One helper keeps them in lockstep:

```ts
function toRegistrationId(hash: string) {
  return `0x${hash}`;
}
```

then `ids: hashes.map(toRegistrationId)` here and `labelNames.get(toRegistrationId(hash))` at L59.

src/resolvers/lookupDomains/ens.ts
```diff
@@ -12,33 +12,57 @@ type Domain = {
...
-    ...domain,
-    name: labelName ? domain.name.replace(`[${hash}]`, labelName) : domain.name
-  };
+  return domains.map(domain => {
+    const hash = getLabelHash(domain);
+    const labelName = hash ? labelNames.get(`0x${hash}`) : undefined;
+
+    return {
+      ...domain,
+      name: labelName ? domain.name.replace(`[${hash}]`, labelName) : domain.name
```
wa0x6e: Pre-existing (the line just moved), but cheap to fix while it's in the hunk: string-pattern `replace` still interprets `$$`, `$&`, `` $` ``, `$'` in the replacement. Real mainnet labels hit this — `$$ell` → `$ell.eth`, `$'''` → `.eth''.eth`, and a label `$&` splices the raw `[hash]` back in (`labelName_contains:"$$"` and `"$'"` both return live domains; ≥1000 names start with `$`). A replacer function makes `labelName` literal:

```suggestion
      name: labelName ? domain.name.replace(`[${hash}]`, () => labelName) : domain.name
```

wa0x6e (changes requested):
Batching logic looks right — I diffed this branch's resolver against master on live mainnet/sepolia addresses (incl. a 39-name account with 23 hashed labels and a 133-name account) and the output is byte-identical, and the hashed-label decode works end to end through the API (`global.[b50225…].eth` → `global.aragonid.eth`).

One ask before merge: nothing outside the mocked unit test exercises the `registrations` query against a real subgraph. `test/integration/resolvers/lookupDomains.test.ts` runs the ENS path live, but no case asserts a decoded label — a regression that served raw `[hash]` names (or a query that graph-node rejected) would still pass `toContain('.eth')` / `toContain('boorger.eth')`. Please add one integration case that pins a decoded name, e.g.:

```ts
it('decodes hashed parent labels on subdomains', async () => {
  const result = await lookupDomains('0x279489452dd8035f82326c8036f81d7bd1c65e6c');

  expect(result).toContain('global.aragonid.eth');
});
```

(`0x2794…` owns `global.[b50225e6…].eth`; the parent registration resolves to `aragonid` — verified live. `0xbd74a131c039430b520303bd2f76ef8b9f326454` → `fathom.aragonid.eth` works as an alternate if that owner changes.)

src/resolvers/lookupDomains/ens.ts
tony8713: Confirmed with a null-list stub: the fallback resolved to the raw hashed name. Removed the fallback in 68e44ac and added `rejects when the registration list is null`, so the regression test preserves the invariant without a code comment.

src/resolvers/lookupDomains/ens.ts
tony8713: Keeping this inline. `rg` finds only these two registration-ID constructions, both consume the same unmodified regex capture in this function, and there is no normalization path today; a helper would add an abstraction without fixing a current defect.

src/resolvers/lookupDomains/ens.ts
tony8713: Confirmed the cited JavaScript replacement semantics and live labels containing those tokens. Switched to a replacer callback in 68e44ac; the existing batched-label test now uses `$&`, and reverting to string replacement makes it fail.

tony8713:
Addressed the requested live coverage in 68e44ac: the integration suite now asserts `global.aragonid.eth` for `0x279489452dd8035f82326c8036f81d7bd1c65e6c`. I independently confirmed the account response contains `global.[b50225…].eth` and the registration maps that ID to `aragonid`; removing `id` from the live query now fails this test while the mocked unit tests remain green.

src/resolvers/lookupDomains/ens.ts
```diff
@@ -12,33 +12,49 @@ type Domain = {
...
+  id: string;
+  domain?: Pick<Domain, 'labelName'>;
+};
+
+function getLabelHash(domain: Domain) {
+  return domain.name.match(/\[(.*?)\]/)?.[1];
+}
 
-  if (!hash) return domain;
+async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Handle[]> {
```
wa0x6e, line 24: Suggestion: `fetchDomainNames` mixes two responsibilities — the batched registrations fetch and the domains→names mapping (it always does the mapping, but only fetches when hashed labels are present). Splitting network from pure mapping keeps each piece one job:

```ts
async function fetchLabelNames(
  hashes: string[],
  chainId: string
): Promise<Map<string, string | undefined>> {
  if (!hashes.length) return new Map();

  const {
    data: { data }
  } = await graphQlCall<{ registrations: Registration[] }>(
    constants.ensSubgraph[chainId],
    `query Registrations($ids: [String!]!, $first: Int!) {
      registrations(first: $first, where: { id_in: $ids }) {
        id
        domain {
          labelName
        }
      }
    }`,
    { ids: hashes.map(hash => `0x${hash}`), first: hashes.length }
  );

  return new Map(
    data.registrations.map(registration => [registration.id, registration.domain?.labelName])
  );
}

function decodeName(domain: Domain, labelNames: Map<string, string | undefined>): Handle {
  const hash = getLabelHash(domain);
  const labelName = hash ? labelNames.get(`0x${hash}`) : undefined;

  return labelName ? domain.name.replace(`[${hash}]`, () => labelName) : domain.name;
}
```

with `lookupDomains` orchestrating:

```ts
const hashes = [
  ...new Set(domains.map(getLabelHash).filter((hash): hash is string => Boolean(hash)))
];
const labelNames = await fetchLabelNames(hashes, chainId);

return domains.map(domain => decodeName(domain, labelNames));
```

`fetchLabelNames` = network only (hashes in, lookup map out, no-op on empty), `decodeName` = pure string patch, `lookupDomains` = orchestration. Same call sequence and output — existing tests pass unchanged. Naming/structure hygiene only, no behavior change; fine to skip if we want to freeze churn on this PR.

tony8713: Confirmed this is structural only, not a defect. `fetchDomainNames` is private with one caller and its contract is already one operation: `Domain[]` to `Handle[]`, including the no-hash fast path. Splitting it would add two helpers and expose the intermediate registration map to `lookupDomains` without changing behavior, so I am freezing the churn here.

wa0x6e (approved):
utAck

PR #572, opened 2026-08-19
fix(image): skip invalid user hub queries

### LINKED ISSUE #563: fix(image): skip the hub query for user avatars with non-address ids
## Today

`src/resolvers/image/snapshot.ts:103-124`:

```ts
function normalizeEvmAddress(value: string) {
  try { return getAddress(value.trim()); }
  catch { return value; }          // falls back to the raw input
}
```

Single call site (L121), shared by all five offchain resolvers from `createPropertyResolver`:

- `resolveUserAvatar` / `resolveUserCover` — entity `user`, id is an address. `GET /avatar/<id>` runs `resolveUserAvatar` on every request (`constants.resolvers.avatar`), and `<id>` is the raw URL chunk. Non-address ids (`vitalik.eth`, `foo.lens`, `foo.stark`, a starknet `0x…64hex`, junk) → `getAddress` throws → raw string → hub `user(id: "vitalik.eth")`. Hub users are keyed by address, so that query can never match: one wasted hub round trip per avatar request whose id is not an EVM address.
- `resolveSpaceAvatar` / `resolveSpaceCover` / `resolveSpaceLogo` on `s` / `s-tn` — entity `space`, id is a slug (`ens.eth`). The raw fallback is what makes these work.

One function, two behaviours, one is wrong for users.

## Change

Split by entity, local to the file:

```ts
// user ids are addresses; anything else can't match a hub user
function toUserId(value: string): string | null {
  return isEvmAddress(value) ? getAddress(value) : null;
}

// space ids are slugs (`ens.eth`) or, on some networks, addresses — checksum when it parses
function toSpaceId(value: string): string {
  try { return getAddress(value.trim()); } catch { return value; }
}
```

```ts
const id = entity === 'user' ? toUserId(address) : toSpaceId(address);
if (!id) return false;
value = await getOffchainProperty(..., id, entity, property);
```

Check first whether the hub has starknet users; if so `toUserId` also accepts `isStarknetAddress` (returning the value as-is, like the onchain branch does).

## Acceptance

- Unit: `resolveUserAvatar('vitalik.eth')`, `('foo.lens')`, `('<starknet address>')` (unless hub supports it) → `graphQlCall` never called, returns `false`.
- Unit: `resolveUserAvatar('<evm address>')` still queries with the checksummed address.
- Unit: `resolveSpaceAvatar('ens.eth')` still queries with `ens.eth`; `resolveSpaceAvatar('<evm address>')` with the checksummed form.
- Existing image integration fixtures unchanged.

tony8713:
## Summary

- stop invalid user ids before they reach Snapshot Hub
- preserve EVM checksumming, Starknet user ids, and offchain space ids

Fixes #563

### DIFF
--- a/src/resolvers/image/snapshot.ts
+++ b/src/resolvers/image/snapshot.ts
@@ -1,4 +1,5 @@
 import { getAddress } from '@ethersproject/address';
+import snapshot from '@snapshot-labs/snapshot.js';
 import { defaultOffchainNetwork, offchainNetworks } from '../../constants.json';
 import { graphQlCall } from '../../helpers/graphql';
 import { fetchHttpImage, getUrl, spaceIds } from '../../helpers/http';
@@ -101,7 +102,15 @@ async function getOnchainProperty(
   return spaces?.map(space => space.metadata?.[property]).filter(Boolean)[0];
 }
 
-function normalizeEvmAddress(value: string) {
+function normalizeUserId(value: string): string | null {
+  try {
+    return getAddress(value.trim());
+  } catch {
+    return value.startsWith('0x') && snapshot.utils.isStarknetAddress(value) ? value : null;
+  }
+}
+
+function normalizeSpaceId(value: string) {
   try {
     return getAddress(value.trim());
   } catch {
@@ -117,9 +126,12 @@ function createPropertyResolver(entity: Entity, property: Property) {
     if (!Object.keys(API_URLS).includes(networkId)) return false;
 
     if (offchainNetworks.includes(networkId) || entity === 'user') {
+      const id = entity === 'user' ? normalizeUserId(address) : normalizeSpaceId(address);
+      if (!id) return false;
+
       value = await getOffchainProperty(
         offchainNetworks.includes(networkId) ? networkId : defaultOffchainNetwork,
-        normalizeEvmAddress(address),
+        id,
         entity,
         property
       );
--- a/src/resolvers/image/snapshot.ts
+++ b/src/resolvers/image/snapshot.ts
@@ -103,10 +103,11 @@ async function getOnchainProperty(
 }
 
 function normalizeUserId(value: string): string | null {
+  const trimmed = value.trim();
   try {
-    return getAddress(value.trim());
+    return getAddress(trimmed);
   } catch {
-    return value.startsWith('0x') && snapshot.utils.isStarknetAddress(value) ? value : null;
+    return trimmed.startsWith('0x') && snapshot.utils.isStarknetAddress(trimmed) ? trimmed : null;
   }
 }
 
--- a/test/unit/resolvers/image/noData.test.ts
+++ b/test/unit/resolvers/image/noData.test.ts
@@ -1,7 +1,11 @@
 import axios from 'axios';
 import nodeFetch from 'node-fetch';
 import farcaster from '../../../../src/resolvers/image/farcaster';
-import { resolveSpaceAvatar, resolveSpaceLogo } from '../../../../src/resolvers/image/snapshot';
+import {
+  resolveSpaceAvatar,
+  resolveSpaceLogo,
+  resolveUserAvatar
+} from '../../../../src/resolvers/image/snapshot';
 import { resolveAvatar as resolveSxAvatar } from '../../../../src/resolvers/image/space-sx';
 import starknet from '../../../../src/resolvers/image/starknet';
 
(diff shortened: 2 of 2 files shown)


src/resolvers/image/snapshot.ts
```diff
@@ -101,7 +102,15 @@ async function getOnchainProperty(
...
 }
 
-function normalizeEvmAddress(value: string) {
+function normalizeUserId(value: string): string | null {
+  try {
+    return getAddress(value.trim());
+  } catch {
+    return value.startsWith('0x') && snapshot.utils.isStarknetAddress(value) ? value : null;
+  }
+}
```
wa0x6e, line 112: `snapshot.utils.isStarknetAddress` is a felt-range check only — `0x1`, `0xdead`, 44-hex junk (e.g. the test file's own `NOT_AN_ADDRESS`) all return `true`, so they still reach the hub (`user(id:"0x1")` → live round trip, hub felt-pads it). The repo already has the strict predicate for this in `src/helpers/address.ts:14` (`/^0x[a-fA-F0-9]{64}$/` + felt check, with a comment naming exactly this trap) — used by `helpers/http.ts`, `resolvers/address/starknet.ts`, `normalizeAddresses`.

The unpadded-Starknet acceptance isn't needed by any client: sx `apps/ui/src/helpers/utils.ts:113` `formatAddress` (used by `getStampUrl`, the single builder for every `avatar`/`user-cover`/`space*` URL) runs non-42-length ids through starknet.js `validateAndParseAddress`, which always emits `0x` + 64 hex zero-padded lowercase — pinned by sx's own test (`utils.test.ts:309-312`: `0x4c5dda…` → `avatar/0x00000000000000000000000004c5dda…`). So the strict helper is the right one here.

```suggestion
function normalizeUserId(value: string): string | null {
  try {
    return getAddress(value.trim());
  } catch {
    return isStarknetAddress(value) ? value : null;
  }
}
```

Plus, outside this hunk:

```diff
 import { getAddress } from '@ethersproject/address';
-import snapshot from '@snapshot-labs/snapshot.js';
 import { defaultOffchainNetwork, offchainNetworks } from '../../constants.json';
+import { isStarknetAddress } from '../../helpers/address';
```

and in `test/unit/resolvers/image/noData.test.ts`: move `UNPADDED_STARKNET_ADDRESS` from the "normalizes a valid %s user id" rows into the "answers false … without asking" list, and add `'0x1'` and `NOT_AN_ADDRESS` there too. Verified locally: prettier clean, `noData.test.ts` 24/24, `tsc --noEmit` clean.

src/resolvers/image/snapshot.ts
```diff
@@ -101,7 +102,15 @@ async function getOnchainProperty(
   return spaces?.map(space => space.metadata?.[property]).filter(Boolean)[0];
 }
 
-function normalizeEvmAddress(value: string) {
+function normalizeUserId(value: string): string | null {
+  try {
+    return getAddress(value.trim());
+  } catch {
+    return value.startsWith('0x') && snapshot.utils.isStarknetAddress(value) ? value : null;
```
wa0x6e: Trimming is asymmetric: the EVM branch uses `value.trim()`, the Starknet branch tests/returns the raw `value`. A leading space is reachable — Express decodes `%20` in the path segment into `req.params.id`, and `parseQuery` only lowercases, never trims — so `/avatar/%200x07ff…` is rejected while `/avatar/%200xef83…` is accepted. Trim once and use it on both branches (on top of the helper-reuse suggestion above):

```ts
function normalizeUserId(value: string): string | null {
  const trimmed = value.trim();
  try {
    return getAddress(trimmed);
  } catch {
    return isStarknetAddress(trimmed) ? trimmed : null;
  }
}
```

wa0x6e (changes requested):
see comments

src/resolvers/image/snapshot.ts
tony8713: I did not apply the strict helper because it narrows Hub's accepted ID contract. Live Hub canonicalizes `0x1`, and canonicalizes `0x269cb810bc22316f8e6a53d164398e83f8255f7e980621f6083e5f7bad74fa0` to `0x0269cb810bc22316f8e6a53d164398e83f8255f7e980621f6083e5f7bad74fa0` with `djizus.stark` and a real IPFS avatar, so a valid direct Stamp request would regress to fallback. This branch mirrors Hub's felt validation, while the `0x` guard still skips decimal `1` and names.

src/resolvers/image/snapshot.ts
tony8713: Confirmed. Express makes the leading space reachable and `parseQuery` does not trim, so `770ee68` trims once before EVM and Starknet validation and returns the trimmed Starknet ID. Removing that trim makes all three spaced EVM, padded Starknet, and unpadded Starknet rows fail.

wa0x6e (approved):
utAck

PR #573, opened 2026-08-19
fix: isolate invalid Space ID labels

### LINKED ISSUE #564: fix(spaceId): one invalid .bnb label rejects the whole resolveNames batch
## Today

`src/resolvers/address/spaceId.ts:83-98`:

```ts
const normalizedHandles = normalizeHandles(handles);   // endsWith('.bnb') only
...
const namehashes = normalizedHandles.map(namehash);    // outside any try
```

Ethers' `namehash` throws on labels ENS normalization rejects. Verified with `@ethersproject/hash@5.8`:

- `foo!.bnb` → `Disallowed codepoint: 0x21`
- `a..bnb`, `.bnb` → `invalid ENS name; empty component`
- `foo_bar.bnb` → `underscore only allowed at start`

One such handle in the batch rejects the whole `resolveNames` promise:

1. every valid `.bnb` handle alongside it resolves to nothing for that request;
2. `resolvers/address/index.ts` catches the rejection, and the message is not in `isSilencedError`, so it is `capture()`d to Sentry — user-supplied junk becomes an error report.

`lookupAddresses` is unaffected: its namehash input is `<addr>.addr.reverse`, always valid.

## Change

Compute the hash per handle, drop the ones that throw, and continue with the rest:

```ts
const pairs = normalizedHandles.flatMap(handle => {
  try { return [[handle, namehash(handle)] as const]; } catch { return []; }
});
```

Then feed `pairs.map(([, h]) => h)` to `resolveNameHashes` and map back through `pairs`. No Sentry capture for the dropped ones: an invalid label is input, not a fault. Same shape gwei already uses (`tokenId()` returns `''` on failure and is filtered).

## Acceptance

- Unit: `resolveNames(['foo!.bnb', '<valid>.bnb'])` → the valid one resolves, no throw, `capture` not called.
- Unit: batch of only invalid `.bnb` handles → `{}` without any RPC.
- Existing spaceId integration fixtures unchanged.

tony8713:
## Summary

Invalid, noncanonical, or unsupported `.bnb` names could disrupt Space ID batches or produce missing or misleading results for otherwise valid requests.

Validate and hash Space ID names consistently, isolate invalid inputs, and discard empty records while preserving valid sibling and reverse-resolution results.

Closes #564

### DIFF
--- a/src/resolvers/address/spaceId.ts
+++ b/src/resolvers/address/spaceId.ts
@@ -82,16 +82,24 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
 }
 
 export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Address>> {
-  const normalizedHandles = normalizeHandles(handles);
+  const pairs = normalizeHandles(handles).flatMap(handle => {
+    try {
+      return [[handle, namehash(handle)] as const];
+    } catch {
+      return [];
+    }
+  });
 
-  if (normalizedHandles.length === 0) return {};
+  if (pairs.length === 0) return {};
 
-  const namehashes = normalizedHandles.map(namehash);
-  const addresses: Record<string, Address> = await resolveNameHashes(namehashes, 'addr');
+  const addresses: Record<string, Address> = await resolveNameHashes(
+    pairs.map(([, hash]) => hash),
+    'addr'
+  );
   const results = {};
 
   Object.entries(addresses).forEach(([hash, addr]) => {
-    const handle = normalizedHandles[namehashes.indexOf(hash)];
+    const handle = pairs.find(([, pairHash]) => pairHash === hash)?.[0];
     if (handle) results[handle] = addr;
   });
 
--- a/src/resolvers/address/spaceId.ts
+++ b/src/resolvers/address/spaceId.ts
@@ -1,5 +1,4 @@
-import { ens_normalize } from '@adraffy/ens-normalize';
-import { namehash } from '@ethersproject/hash';
+import { namehash, normalize } from 'viem/ens';
 import { EMPTY_ADDRESS, isEvmAddress } from '../../helpers/address';
 import { batchContractCalls, getProvider } from '../../helpers/provider';
 import { Address, Handle } from '../../helpers/types';
@@ -68,8 +67,8 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
 
   if (normalizedAddresses.length === 0) return {};
 
-  const reverseNamehashes = normalizedAddresses.map(addr => {
-    return namehash(`${addr.slice(2)}.addr.reverse`);
+  const reverseNamehashes: string[] = normalizedAddresses.map(addr => {
+    return namehash(normalize(`${addr.slice(2)}.addr.reverse`));
   });
   const names: Record<string, Handle> = await resolveNameHashes(reverseNamehashes, 'name');
   const results = {};
@@ -85,7 +84,7 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
 export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Address>> {
   const pairs = normalizeHandles(handles).flatMap(handle => {
     try {
-      if (ens_normalize(handle) !== handle) return [];
+      if (normalize(handle) !== handle) return [];
       return [[handle, namehash(handle)] as const];
     } catch {
       return [];
--- a/src/resolvers/address/spaceId.ts
+++ b/src/resolvers/address/spaceId.ts
@@ -1,3 +1,4 @@
+import { ens_normalize } from '@adraffy/ens-normalize';
 import { namehash } from '@ethersproject/hash';
 import { EMPTY_ADDRESS, isEvmAddress } from '../../helpers/address';
 import { batchContractCalls, getProvider } from '../../helpers/provider';
@@ -84,6 +85,7 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
 export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Address>> {
   const pairs = normalizeHandles(handles).flatMap(handle => {
     try {
+      if (ens_normalize(handle) !== handle) return [];
       return [[handle, namehash(handle)] as const];
     } catch {
       return [];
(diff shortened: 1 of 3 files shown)


test/unit/resolvers/address.test.ts
```diff
@@ -125,3 +127,51 @@ describe('address resolvers - resolver failures', () => {
...
+    jest.spyOn(provider, 'batchContractCalls').mockRejectedValueOnce(error);
+
+    await expect(resolveNames([HANDLE])).resolves.toEqual({});
+    expect(capture).toHaveBeenCalledTimes(1);
+    expect(capture).toHaveBeenCalledWith(error, {
+      tags: { provider: 'Space ID' },
+      contexts: { input: { resolveNames: [HANDLE] } }
+    });
+  });
+});
```
wa0x6e, line 203: The rewritten map-back (`pairs.find(...)` at `spaceId.ts:102`) is never exercised with more than one surviving pair — every scenario here ends with ≤1 `[handle, hash]` by the time the map-back runs, so swapping it for `pairs[0][0]` keeps all tests green (integration fixture doesn't catch it either: its "mix" case only gets one resolver back). One test with two valid handles, addresses returned out of input order, closes the gap:

```suggestion

  it('maps each valid handle back to its own address', async () => {
    const HANDLE_B = 'another.bnb';
    const HASH_B = namehash(HANDLE_B);
    const RESOLVER_B = '0x5555555555555555555555555555555555555555';
    const RESOLVED_ADDRESS_B = '0x330bc93D88C0aF11f1159eA89a885d5ADd3A7Cf7';

    const batch = jest
      .spyOn(provider, 'batchContractCalls')
      .mockResolvedValueOnce({ [HASH]: RESOLVER, [HASH_B]: RESOLVER_B })
      // Returned out of input order, to prove the map-back keys off the hash, not position.
      .mockResolvedValueOnce({ [HASH_B]: RESOLVED_ADDRESS_B, [HASH]: RESOLVED_ADDRESS });

    await expect(resolveNames([INVALID_HANDLES[0], HANDLE, HANDLE_B])).resolves.toEqual({
      [HANDLE]: RESOLVED_ADDRESS,
      [HANDLE_B]: RESOLVED_ADDRESS_B
    });
    expect(batch).toHaveBeenCalledTimes(2);
    expect(batch.mock.calls[0][3]).toEqual([HASH, HASH_B]);
    expect(capture).not.toHaveBeenCalled();
  });
});
```

Verified: passes on current code, fails with the `pairs[0][0]` mutation.

src/resolvers/address/spaceId.ts
```diff
@@ -82,16 +82,24 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
 }
 
 export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Address>> {
-  const normalizedHandles = normalizeHandles(handles);
+  const pairs = normalizeHandles(handles).flatMap(handle => {
+    try {
+      return [[handle, namehash(handle)] as const];
```
wa0x6e, line 88: Pre-existing, not introduced here — but this is exactly the validation seam, and the sibling idiom closes it in one line. `namehash` only *throws* on labels ENS normalization rejects; labels it silently rewrites (fullwidth `ｆｏｏ.bnb`, NFD `é.bnb`) hash identically to the canonical name, so they resolve to that name's owner and get cached under the raw key. `ens.ts:19` and `basename.ts:38` both guard with `ens_normalize(x) === x`; spaceId never did.

```suggestion
      if (ens_normalize(handle) !== handle) return [];
      return [[handle, namehash(handle)] as const];
```

plus `import { ens_normalize } from '@adraffy/ens-normalize';` at the top (already a direct dep). `ens_normalize` throws on the same four invalid labels `namehash` does, so it's safe inside the existing `try`. Verified: `namehash('ｆｏｏ.bnb') === namehash('foo.bnb')`; with the guard, `ens_normalize('ｆｏｏ.bnb')` → `'foo.bnb'` ≠ input → dropped; all 19 unit tests still pass (also with `'ｆｏｏ.bnb'` added to `INVALID_HANDLES`, which would be a cheap regression case).

src/resolvers/address/spaceId.ts
```diff
@@ -82,16 +82,24 @@ export async function lookupAddresses(addresses: Address[]): Promise<Record<Addr
...
+  const addresses: Record<string, Address> = await resolveNameHashes(
+    pairs.map(([, hash]) => hash),
+    'addr'
+  );
   const results = {};
 
   Object.entries(addresses).forEach(([hash, addr]) => {
-    const handle = normalizedHandles[namehashes.indexOf(hash)];
+    const handle = pairs.find(([, pairHash]) => pairHash === hash)?.[0];
     if (handle) results[handle] = addr;
```
wa0x6e: Pre-existing, one line, since this loop is being touched: `resolveNameHashes` strips a zero *resolver*, but not a zero `addr()` result — `Multicaller.execute()` sets whatever the contract returned, and nothing downstream in `index.ts` drops a zero-address *value* for `resolveNames`. A `.bnb` name with a resolver but no addr record comes back to the API as `0x000…0`. `basename.ts:89` and `gwei.ts:70` both guard this; spaceId is the outlier (`EMPTY_ADDRESS` is already imported here).

```suggestion
    if (handle && addr !== EMPTY_ADDRESS) results[handle] = addr;
```

wa0x6e (changes requested):
see reviews

test/unit/resolvers/address.test.ts
tony8713: Confirmed the unit file alone misses this mutation, but the related set does not. With the exact `pairs[0][0]` mutation, `test/integration/resolvers/address/spaceId.test.ts` fails its existing mixed-domain case: it returns the valid address under `inexistent-domain-for-test.bnb` instead of `boorger.bnb`, while the other 18 related suites pass. I did not add duplicate coverage.

src/resolvers/address/spaceId.ts
tony8713: Confirmed. `ens_normalize("ｆｏｏ.bnb")` is `foo.bnb`, and both inputs have the same namehash. Added the exact-normalization guard and fullwidth regression case; removing the guard makes the no-RPC assertion fail while the other 18 related suites pass.

src/resolvers/address/spaceId.ts
tony8713: Confirmed through the shared `resolveNames` path: a zero `addr()` result was returned as `{ "boorger.bnb": EMPTY_ADDRESS }`. Added the value guard and regression test; removing the guard fails only that test across the related set.

wa0x6e (approved):
utAck

PR #574, opened 2026-08-19
fix(image): skip getResolver for non-ENS inputs

### LINKED ISSUE #565: fix(image): skip getResolver for input that is not ENS-shaped
## Today

`src/resolvers/image/ens.ts` runs on every `GET /avatar/<id>` (`constants.resolvers.avatar`). `<id>` is the raw, lowercased URL chunk.

```ts
async function castToEnsName(nameOrAddress) {
  if (isAddress(nameOrAddress)) return (await lookupAddresses([nameOrAddress]))[nameOrAddress];
  return nameOrAddress;   // anything else, as-is
}
...
const ensResolver = await provider.getResolver(ensName);   // mainnet RPC
```

Anything that is not an EVM address goes straight to `getResolver()`: a starknet address (`0x` + 64 hex), a string with no dot, a name ENS normalization rejects (`a..b`, `foo!`). Each is one wasted registry `eth_call`, and `avatar` is the hottest image type.

The sibling-TLD case (`foo.lens`, `foo.bnb`, …) is the same call site but a different gate; it goes with the ENS/UD TLD gate in a separate issue. This one is about shape.

## Change

In `castToEnsName`, for the non-address branch, return `undefined` unless the input is ENS-shaped:

```ts
function isEnsShaped(name: string): boolean {
  if (isStarknetAddress(name)) return false;
  try { return ens_normalize(name) === name; } catch { return false; }
}
```

Mirrors `normalizeEns` in `resolvers/address/ens.ts`. Not `.eth`-only: DNS names (`foo.xyz`) stay allowed, ENS serves them.

`ens_normalize` and `isStarknetAddress` are already imported elsewhere in the repo; no new dependency.

## Acceptance

- Unit: `resolve('<starknet address>')`, `resolve('nodot')`, `resolve('a..b')` → `getResolver` never called, returns `false`.
- Unit: `resolve('vitalik.eth')` and `resolve('foo.xyz')` still reach `getResolver`.
- Existing image/ens integration fixtures unchanged.

Independent of #560 (which changes what the call is, not whether it happens).

tony8713:
## Summary

Non-address avatar inputs went to the mainnet ENS resolver even when they could not be an ENS name.

`ens_normalize` accepts single-label strings, so require a dot and successful normalization before the resolver call. Valid equivalents are passed through in normalized form, while EVM addresses keep their reverse lookup path.

Closes #565

### DIFF
--- a/src/resolvers/image/ens.ts
+++ b/src/resolvers/image/ens.ts
@@ -1,3 +1,4 @@
+import { ens_normalize } from '@adraffy/ens-normalize';
 import { isAddress } from '@ethersproject/address';
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
 import { fetchHttpImage } from '../../helpers/http';
@@ -9,7 +10,13 @@ async function castToEnsName(nameOrAddress: string): Promise<string | undefined>
     return (await lookupAddresses([nameOrAddress]))[nameOrAddress];
   }
 
-  return nameOrAddress;
+  if (!nameOrAddress.includes('.')) return undefined;
+
+  try {
+    return ens_normalize(nameOrAddress);
+  } catch {
+    return undefined;
+  }
 }
 
 export default async function resolve(nameOrAddress: string) {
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
tAck

PR #580, opened 2026-08-20
fix(resolvers): skip sibling TLD names

### LINKED ISSUE #566: fix(resolvers): skip sibling-TLD names in ENS, Unstoppable Domains and image/ens
## Today

Five resolvers own a TLD nobody else can serve: `.lens` (lens), `.bnb` (Space ID), `.stark` (Starknet), `.gwei` (GNS), `.shib` (Shibarium), plus `.base.eth` (Basename, L2). The two open-ended resolvers still accept those names and query for them:

- `src/resolvers/address/ens.ts:27` — `normalizeHandles` is `ens_normalize(name) === name` only. `foo.lens` etc. pass → subgraph query, then `provider.resolveName()` per unresolved handle: `getResolver` walks `foo.lens` → `resolver('foo.lens')`, `resolver('lens')` — 2 mainnet RPCs, always a miss. `.base.eth` also resolves here via CCIP (registry + resolver + gateway HTTP), duplicating the Basename resolver in the same batch.
- `src/resolvers/address/unstoppableDomains.ts:26` — `^[.a-z0-9-]+$` passes any dotted name. The UD lib's `checkDomain` only rejects `eth|luxe|xyz|kred` (`@unstoppabledomains/resolution` `Uns.js:723`), so `foo.lens/.bnb/.stark/.gwei/.shib` → `namehash` → Polygon `ownerOf` eth_call, guaranteed revert.
- `src/resolvers/image/ens.ts:6-20` — avatar: a `foo.lens` id, or an address whose primary handle from `lookupAddresses` comes back as `foo.lens`/`foo.bnb`, goes to `provider.getResolver('foo.lens')`.

## Change

One list, one helper, in `src/helpers/address.ts`:

```ts
// TLDs served by one dedicated resolver. Open-ended resolvers (ENS, UD) skip them:
// a `.lens` name can't exist on ENS mainnet or UNS Polygon; each miss is a wasted RPC.
export const OWNED_TLDS = ['.lens', '.bnb', '.stark', '.gwei', '.shib', '.base.eth'];
export const hasOwnedTld = (h: Handle) => OWNED_TLDS.some(tld => h.endsWith(tld));
```

Consumers:
- `address/ens.ts` `normalizeHandles`: `.filter(h => h && !hasOwnedTld(h))`
- `address/unstoppableDomains.ts` `normalizeHandles`: add `&& !hasOwnedTld(h)`
- `image/ens.ts` `castToEnsName`: return `undefined` when `hasOwnedTld(name)` — both branches (raw name, and the handle from `lookupAddresses`)

Literal list, not per-resolver `TLDS` exports aggregated in the index: six entries that change once a year; aggregation adds a type and a reduce for nothing.

## Not in the list, on purpose

- `.eth` — ENS owns it; UD already rejects it in-lib ("is not supported", silenced).
- DNS TLDs (`foo.xyz`, `foo.com`) — ENS serves them via DNSSEC; UD sells some. Can't cheaply tell which; leave.
- UD-only TLDs (`.crypto`, `.nft`, `.x`, …) still reach ENS. Long, moving list; add the top few if traffic shows it.
- `.base.eth` **is** in the list even though ENS mainnet can resolve it via CCIP: the Basename resolver answers from L2 in the same batch, so excluding it drops 2 RPCs + 1 gateway hit and no result. If belt-and-braces via ENS is wanted, remove it from the list — say so on the PR.

Global `normalizeHandles` lowercases first, so `endsWith` on the lowercase list is enough.

## Acceptance

- Unit: ENS `resolveNames(['foo.lens'])` → neither `graphQlCall` nor `provider.resolveName` called; UD `resolveNames(['foo.lens'])` → `snapshot.utils.call` not called; `image/ens` `resolve('foo.lens')` → `getResolver` not called.
- Unit: `resolveNames(['vitalik.eth', 'foo.xyz'])` unchanged on both.
- Existing integration fixtures unchanged (`jesse.base.eth` reverse lookup stays: `lookupAddresses` is not gated, only `resolveNames`).

Related: #565 (image/ens shape gate, same call site, different check); #559 makes ENS misses near-free once batched, the UD and image/ens halves are not covered by anything else.

tony8713:
Open-ended ENS and Unstoppable Domains lookups currently accept terminal names owned by dedicated sibling resolvers. The ENS image resolver can also send those names to mainnet after direct input or reverse lookup.

This introduces a shared, case-insensitive terminal-suffix check and applies it before those address and avatar requests. Embedded suffix text continues through the existing resolver, and `.base.eth` remains eligible for ENS because CCIP can be the only successful route for some names.

Closes #566

### DIFF
--- a/src/helpers/address.ts
+++ b/src/helpers/address.ts
@@ -6,6 +6,13 @@ import { Address, Handle } from './types';
 
 export const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000';
 
+const OWNED_TLDS = ['.lens', '.bnb', '.stark', '.gwei', '.shib'];
+
+export function hasOwnedTld(handle: Handle): boolean {
+  const normalizedHandle = handle.toLowerCase();
+  return OWNED_TLDS.some(tld => normalizedHandle.endsWith(tld));
+}
+
 export function isEvmAddress(address: Address): boolean {
   return /^0x[a-fA-F0-9]{40}$/.test(address);
 }
--- a/src/resolvers/address/ens.ts
+++ b/src/resolvers/address/ens.ts
@@ -3,7 +3,7 @@ import { getAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
 import constants from '../../constants.json';
-import { isEvmAddress } from '../../helpers/address';
+import { hasOwnedTld, isEvmAddress } from '../../helpers/address';
 import { isSilencedError, isTransportFailure } from '../../helpers/errors';
 import { graphQlCall } from '../../helpers/graphql';
 import { getProvider } from '../../helpers/provider';
@@ -28,7 +28,7 @@ function normalizeAddresses(addresses: Address[]): Address[] {
 }
 
 function normalizeHandles(names: Handle[]): Handle[] {
-  return normalizeEns(names).filter(h => h);
+  return normalizeEns(names).filter(h => h && !hasOwnedTld(h));
 }
 
 export async function lookupAddresses(addresses: Address[]): Promise<Record<Address, Handle>> {
--- a/src/resolvers/address/unstoppableDomains.ts
+++ b/src/resolvers/address/unstoppableDomains.ts
@@ -1,7 +1,7 @@
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
 import Resolution, { NamingServiceName } from '@unstoppabledomains/resolution';
-import { isEvmAddress } from '../../helpers/address';
+import { hasOwnedTld, isEvmAddress } from '../../helpers/address';
 import { isSilencedError, isTransportFailure } from '../../helpers/errors';
 import { withoutEmptyValues } from '../../helpers/object';
 import { batchContractCalls, getProvider } from '../../helpers/provider';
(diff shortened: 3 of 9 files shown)


tony8713:
Rebased onto master (`df5f1cf`) to clear the conflict; head is now `3fa652a`.

The conflict was with #565 (`fix(image): validate reverse-resolved ENS names`), which landed on master in the same function. It already routes both branches of `castToEnsName` through one `ens_normalize` call, so this branch no longer needs its own `normalizeEnsName` helper: the sibling-TLD check is now a single line on the normalized name. That is strictly better here, because the check now runs after normalization on the reverse-lookup path too, so `foo.le<ZWSP>ns` is caught.

Two tests from this branch were dropped as superseded rather than resolved. Both asserted the pre-#565 behaviour that `castToEnsName` returns the raw reverse-lookup name, which is no longer true: `keeps existing invalid reverse-name handling` expected `a..b` to reach the provider, and `keeps resolving ENS names returned by reverse lookup` duplicated what master`s `normalizes a valid reverse result before lookup and fallback` now covers. The sibling-TLD cases they were sitting next to are kept.

Test is red on the 9 Starknet cases, which are the exhausted upstream Alchemy quota and not this branch: same 9 names as the master control run on the identical commit ([32388732172](https://github.com/snapshot-labs/stamp/actions/runs/32388732172)), and 10 `Monthly capacity limit exceeded` lines in the log. Lint is green.

ChaituVR (approved):
utAck

Issue #585, opened 2026-08-20
fix(image): starknet resolver forwards over-prime addresses to RPC


wa0x6e:
Sentry: [STAMP-74](https://snapshot-labs.sentry.io/issues/STAMP-74)

`GET /avatar/s-tn:0x2121212121212121212121212121212121212121212121212121212121212121` throws:

```
RpcError: RPC: starknet_call ... -32602: Invalid params: {"reason":"The maximum field value was exceeded."}
    at getImage (src/resolvers/image/starknet.ts:26)
```

## Root cause

`0x2121…21` is 64 hex chars but ≈ 2^253, above the Starknet felt prime (~2^251). The image resolver validates with a local regex-only helper:

```ts
// src/resolvers/image/starknet.ts
function normalizeAddress(address: string): string | null {
  return /^(0x)?[0-9a-fA-F]{64}$/.test(address) ? address : null;
}
```

so the over-prime value is forwarded to `provider.getStarkProfile()` and the RPC rejects it. `helpers/address.ts` already has `isStarknetAddress()` which adds the felt-range check (`snapshot.utils.isStarknetAddress`), and `resolvers/address/starknet.ts` uses it — the image resolver drifted.

## Fix

Replace the local `normalizeAddress` with the shared `isStarknetAddress` helper:

```ts
const address = isStarkDomain(domainOrAddress)
  ? await getStarknetAddress(domainOrAddress)
  : isStarknetAddress(domainOrAddress) ? domainOrAddress : null;
```

Note: this also stops accepting un-prefixed 64-hex input (`(0x)?`), which the router/`http.ts` already requires `0x` for.

Unrelated but spotted in the same file: provider is hardcoded to mainnet (`getProvider('0x534e5f4d41494e')`) regardless of `s-tn` network.

wa0x6e:
Duplicate of #505 — same root cause (regex-only felt check in the image starknet resolver). Sentry: STAMP-74.

PR #588, opened 2026-08-20
fix(image): validate Starknet profile addresses

### LINKED ISSUE #505: Any 64-char hex value is treated as a Starknet address on the avatar path
`normalizeAddress` in `src/resolvers/starknet.ts:15` gates on shape only:

```ts
if (!address.match(/^(0x)?[0-9a-fA-F]{64}$/)) throw new Error('Invalid starknet address');
```

A Starknet address is a field element, so it is below `L2_ADDRESS_UPPER_BOUND` (2^251 - 256). Only 3.125% of 32-byte hex values are. Transaction hashes, namehashes and proposal ids all match that regex and go straight to `provider.getStarkProfile(address)` on line 33 — the three most recent hub proposal ids are all out of range.

Nothing upstream filters. `parseQuery` lowercases the raw path segment and validates nothing, and `starknet` is in `constants.resolvers.avatar`, so every public `GET /avatar/:id` with a 64-hex id reaches this.

## Reproduction

```js
const snapshot = require('@snapshot-labs/snapshot.js');
const provider = snapshot.utils.getProvider('0x534e5f4d41494e');

// an ethereum transaction hash - matches normalizeAddress's regex
provider
  .getStarkProfile('0x88df016429689c079f3b2f6ad39fa052532c56795b733da78a91ebe6a713944b')
  .catch(e => console.log(e.name, '|', e.message.split('\n').pop().trim()));
```

```
LibraryError | -32602: Invalid params: {"reason":"The maximum field value was exceeded."}
```

Same thing through the live route, which answers 200 with a blockie:

```
curl -sI "https://cdn.stamp.fyi/avatar/0x88df016429689c079f3b2f6ad39fa052532c56795b733da78a91ebe6a713944b?resolver=starknet"
```

## What actually happens

The value is compiled into calldata and a real `starknet_call` goes out to Starknet mainnet through `rpc.snapshot.org` — one per request on a warm process, verified by relaying the provider through a logging proxy. The node rejects it and starknet.js raises the `LibraryError` above, with no `code`, `status` or `response` on the error.

That throw lands in `resolve()`'s `catch { return false; }`, and nothing on the image path calls `capture()`, so this is not Sentry noise. The request falls through to the blockie fallback and the failure is invisible. What it costs is a wasted mainnet round trip on the shared brovider for every one of these. Worth contrasting with the sibling path: #491 adds `isResolvableAddress` to `src/addressResolvers/starknet.ts` precisely because there the same input *would* reach `capture()` and read as an outage.

The ~3% that are valid felts don't throw at all — they come back as an empty profile after the same round trip.

Two more things fall out of the same line:

- The `0x` is optional, so a bare 64-digit string is accepted and read as **decimal**. `/avatar/1111…1111` (64 ones) makes a live call for an entirely unrelated address.
- Exactly 64 hex digits is required, so a Starknet address written without leading zeros is rejected. `th0rgal.stark` is `0xa00373a00352aa367058555149b573322910d54fcdf3a926e3e56d0dcb4b0c` (62 digits) and `getStarkProfile` resolves it to a real picture, but `/avatar/` on that address returns a blockie.

## Suggested fix

Keep the shape check, add the range check with `snapshot.utils.isStarknetAddress` — already a dependency, and the function the hub itself gates with, so it can't drift from what the hub accepts:

```ts
function normalizeAddress(address: string): string {
  if (!/^0x[0-9a-fA-F]{64}$/.test(address) || !snapshot.utils.isStarknetAddress(address)) {
    throw new Error('Invalid starknet address');
  }

  return address;
}
```

The shape half has to stay. `snapshot.utils.isStarknetAddress` is a felt check, not a shape check, and a 40-hex EVM address is a perfectly good felt — it returns `true`. Used alone here it would fire a `starknet_call` for every EVM avatar request, which is nearly all of them.

Same reason the exported `isStarknetAddress` in `src/addressResolvers/utils.ts` must not be swapped for it: `normalizeAddresses` uses it to decide whether to skip `getAddress()`, so EVM addresses would come back lowercased instead of checksummed.

#504 does the equivalent for `normalizeAddresses`. #491 touches this file but only to swap `isStarknetDomain` for the shared `isStarkDomain` — `normalizeAddress` is unchanged there, so this line survives both.

tony8713:
The Starknet image resolver took any 64-digit hex string for an address, so over-prime values such as transaction hashes and proposal ids reached `getStarkProfile` and came back as RPC errors that `isSilencedError` does not silence, which means they were captured rather than merely wasteful. At the same time an address written without its leading zeros was rejected outright and fell through to the blockie. `getAddressFromStarkName` returns addresses in exactly that unpadded form, so the resolver disagreed with itself: the domain branch fed `getStarkProfile` a shape the address branch refused.

Shape is not the invariant here, the address bound is. `helpers/address.ts` gains a predicate that accepts an unpadded id, excludes EVM addresses because they are valid felts too, and caps the input at the 32 bytes an address occupies so one identity cannot be spelled an unbounded number of ways. Both image resolvers now use it, `image/starknet.ts` for the behaviour and `image/snapshot.ts` to drop an inline copy of the same rule.

The trade-off is that short values such as `0x1` now reach `getStarkProfile`, which answers with the default identicon that `resolve` already maps to no image; the 64-digit rule blocked those, at the cost of every address that arrives unpadded.

Domain resolution is unchanged, and so is the strict `isStarknetAddress` that `helpers/http.ts` and the Starknet address resolver depend on to keep EVM addresses out.

Closes #505

#585 was closed as a duplicate.

### DIFF
--- a/src/resolvers/image/starknet.ts
+++ b/src/resolvers/image/starknet.ts
@@ -1,5 +1,5 @@
 import axios from 'axios';
-import { isStarkDomain, isStarknetAddress } from '../../helpers/address';
+import { isStarkDomain, isStarknetFelt } from '../../helpers/address';
 import { axiosDefaultParams, fetchHttpImage, getUrl } from '../../helpers/http';
 import { getProvider } from '../../helpers/provider';
 
@@ -15,7 +15,7 @@ async function getStarknetAddress(domain: string): Promise<string | null> {
 async function getImage(domainOrAddress: string): Promise<string | null> {
   const address = isStarkDomain(domainOrAddress)
     ? await getStarknetAddress(domainOrAddress)
-    : isStarknetAddress(domainOrAddress)
+    : isStarknetFelt(domainOrAddress)
       ? domainOrAddress
       : null;
 
--- a/src/resolvers/image/starknet.ts
+++ b/src/resolvers/image/starknet.ts
@@ -1,14 +1,10 @@
 import axios from 'axios';
-import { isStarkDomain } from '../../helpers/address';
+import { isStarkDomain, isStarknetAddress } from '../../helpers/address';
 import { axiosDefaultParams, fetchHttpImage, getUrl } from '../../helpers/http';
 import { getProvider } from '../../helpers/provider';
 
 const DEFAULT_IMG_URL = 'https://starknet.id/api/identicons/0';
 const provider = getProvider('0x534e5f4d41494e');
-
-function normalizeAddress(address: string): string | null {
-  return /^(0x)?[0-9a-fA-F]{64}$/.test(address) ? address : null;
-}
 
 async function getStarknetAddress(domain: string): Promise<string | null> {
   const address = await provider.getAddressFromStarkName(domain);
@@ -19,7 +15,9 @@ async function getStarknetAddress(domain: string): Promise<string | null> {
 async function getImage(domainOrAddress: string): Promise<string | null> {
   const address = isStarkDomain(domainOrAddress)
     ? await getStarknetAddress(domainOrAddress)
-    : normalizeAddress(domainOrAddress);
+    : isStarknetAddress(domainOrAddress)
+      ? domainOrAddress
+      : null;
 
   if (!address) return null;
 
--- a/src/helpers/address.ts
+++ b/src/helpers/address.ts
@@ -15,6 +15,14 @@ export function isStarknetAddress(address: Address): boolean {
   return /^0x[a-fA-F0-9]{64}$/.test(address) && snapshot.utils.isStarknetAddress(address);
 }
 
+export function isStarknetFelt(address: Address): boolean {
+  return (
+    /^0x[a-fA-F0-9]+$/.test(address) &&
+    !isEvmAddress(address) &&
+    snapshot.utils.isStarknetAddress(address)
+  );
+}
+
 // StarknetID's encoder skips characters outside its alphabet instead of rejecting them, so
 // `a!b.stark` encodes exactly like `ab.stark` and would resolve to that owner. Every label has
 // to match, subdomains included. The alphabet is starknet.js's `basicAlphabet` plus the two
--- a/src/helpers/address.ts
+++ b/src/helpers/address.ts
@@ -17,7 +17,7 @@ export function isStarknetAddress(address: Address): boolean {
 
 export function isStarknetFelt(address: Address): boolean {
   return (
-    /^0x[a-fA-F0-9]+$/.test(address) &&
+    /^0x[a-fA-F0-9]{1,64}$/.test(address) &&
     !isEvmAddress(address) &&
     snapshot.utils.isStarknetAddress(address)
   );
--- a/src/resolvers/image/snapshot.ts
+++ b/src/resolvers/image/snapshot.ts
@@ -1,6 +1,6 @@
 import { getAddress } from '@ethersproject/address';
-import snapshot from '@snapshot-labs/snapshot.js';
 import { defaultOffchainNetwork, offchainNetworks } from '../../constants.json';
+import { isStarknetFelt } from '../../helpers/address';
 import { graphQlCall } from '../../helpers/graphql';
 import { fetchHttpImage, getUrl, spaceIds } from '../../helpers/http';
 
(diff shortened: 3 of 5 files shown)


src/resolvers/image/starknet.ts
```diff
@@ -19,7 +15,9 @@ async function getStarknetAddress(domain: string): Promise<string | null> {
 async function getImage(domainOrAddress: string): Promise<string | null> {
   const address = isStarkDomain(domainOrAddress)
     ? await getStarknetAddress(domainOrAddress)
-    : normalizeAddress(domainOrAddress);
+    : isStarknetAddress(domainOrAddress)
```
wa0x6e: `isStarknetAddress` requires exactly 64 hex digits, so an unpadded Starknet address still falls through to the blockie — #505's third bullet. `th0rgal.stark`'s owner is `0xa00373a00352aa367058555149b573322910d54fcdf3a926e3e56d0dcb4b0c` (62 digits); `provider.getStarkProfile` resolves it to the same profile as the padded form (verified live), but this branch returns `null` for it.

`normalizeUserId` in `src/resolvers/image/snapshot.ts:105` already accepts that shape (`startsWith('0x') && snapshot.utils.isStarknetAddress(trimmed)`, added in 48e18b9), so the two image resolvers now disagree on what a Starknet id is. The shared `isStarknetAddress` can't be loosened — `helpers/http.ts` and `resolvers/address/starknet.ts` rely on the exact-64 shape to keep EVM addresses out, and `test/integration/helpers/address.test.ts` locks that in.

Rather than a second inline copy of the predicate, could this land as a shared helper in `helpers/address.ts` used by both call sites? Something like:

```ts
// Unpadded felt accepted: Snapshot user ids and StarknetID owners are not zero-padded.
// EVM addresses are valid felts too, so they are excluded explicitly.
export function isStarknetFelt(address: Address): boolean {
  return address.startsWith('0x') && !isEvmAddress(address) && snapshot.utils.isStarknetAddress(address);
}
```

then `isStarknetFelt(domainOrAddress) ? domainOrAddress : null` here and `isStarknetFelt(trimmed) ? trimmed : null` in `normalizeUserId`. Trade-off to be aware of: it lets short felts like `0x1` through to `getStarkProfile`, which the 64-digit rule blocked. If unpadded input is deliberately out of scope for this PR, a line in the description saying so would settle #505's remaining bullet.

wa0x6e (changes requested):
Requesting changes for the unpadded-address handling — see comment on `src/resolvers/image/starknet.ts:18`.

src/resolvers/image/starknet.ts
tony8713: Verified and implemented, with your shape. `getAddressFromStarkName('th0rgal.stark')` returns `0xa00373a0...4b0c` unpadded and `getStarkProfile` gives it the same profile as the padded form, so `getImage` was refusing on the address branch the exact shape its own domain branch hands to `getStarkProfile` two lines below. And the exact-64 rule was never a near miss: 64 significant hex digits is at least 2^252, over the felt bound, so it could only ever match a value that was already zero-padded.

`isStarknetFelt` is in `helpers/address.ts` and used at both call sites; `normalizeUserId` loses its inline copy, and `isStarknetAddress` is untouched so `helpers/http.ts` and `resolvers/address/starknet.ts` keep the exact-64 shape. One deviation from your snippet: a regex for the `0x` test rather than `startsWith`, so off-type input cannot throw, matching what `address.test.ts` already locks in for `isStarknetAddress`.

Took the trade-off you flagged. `0x1` reaches `getStarkProfile` now, which answers with `starknet.id/api/identicons/0`, and `resolve()` already maps that to no image, so the cost is a round trip on malformed input. It is in the description.

Against mainnet the unpadded address now returns th0rgal's real avatar where it returned `false` before, while the proposal id and an EVM address still return `false` without touching the network. The red Test check is the 9 Starknet cases master is failing too, from the Alchemy monthly quota behind rpc.snapshot.org.

tony8713: Follow-up, since it is a second deviation from your snippet. `a4eb5df` caps the predicate at 64 hex digits, so `0x` plus the 32 bytes an address occupies. Your version had no upper bound and neither did mine at first; a review pass then measured five different zero-paddings of one address resolving to the same avatar under five cache keys, each paying an RPC call, an image fetch, a resize and two S3 PUTs. The cap admits every real id, padded or not, and rejects only padding wider than an address can be.

Also worth correcting from my note above: I said the two resolvers disagreed, which is true, but the `image/snapshot.ts` half is pure de-duplication rather than a behaviour fix. `helpers/api.ts:29` lowercases the path segment before any resolver sees it, so the input where the old inline expression and the shared helper differ cannot arrive over HTTP.

wa0x6e (approved):
tAck

PR #589, opened 2026-08-20
fix(image): validate reverse-resolved ENS names

### LINKED ISSUE #584: fix(image): validate reverse-resolved names before ENS avatar lookup
## Problem

`src/resolvers/image/ens.ts` validates a name supplied by the caller before the ENS avatar lookup, but the EVM-address branch returns the result of `lookupAddresses` without the same validation.

`src/resolvers/address/index.ts` prefers the Snapshot resolver, and `src/resolvers/address/snapshot.ts` returns the hub profile `user.name`. The profile schema allows any string up to 32 characters. As a result, an avatar request for an address can feed a display name such as `Kamruzzaman`, `CLT #1`, or `../../foo` into `getEnsTextRecord` and the metadata fallback even though it cannot be an ENS name. A valid reverse result such as `VITALIK.eth` also reaches the lookup without normalization.

This was reproduced through the production address branch with focused unit cases: invalid reverse results entered the provider or fallback path, and the uppercase ENS result was passed through unchanged.

## Change

Resolve an address to a candidate name, then apply the same dot and `ens_normalize` gate to both address-derived and caller-supplied names before the Universal Resolver call or metadata fallback.

Affected files:

- `src/resolvers/image/ens.ts`
- `test/unit/resolvers/image/ens.test.ts`

## Acceptance

- A reverse result with no dot, or one rejected by `ens_normalize`, returns `false` without a provider, transport, or image fetch.
- A valid reverse result is normalized before the text-record lookup and fallback URL.
- Caller-supplied ENS names, direct HTTP avatar records, and metadata fallback behavior remain unchanged.

Follow-up from #569.

tony8713:
Closes #584.

Snapshot profile display names can win the reverse lookup for address avatar requests. Those values bypassed the ENS name guard and were sent to the Universal Resolver and metadata fallback even when they could not be ENS names.

Resolve the address to a candidate first, then apply the same shape and ENS normalization checks used for caller-supplied names. Valid reverse names are normalized before both the text-record lookup and fallback request.

### DIFF
--- a/src/resolvers/image/ens.ts
+++ b/src/resolvers/image/ens.ts
@@ -6,14 +6,14 @@ import { getProviderOptions } from '../../helpers/provider';
 import { lookupAddresses } from '../address';
 
 async function castToEnsName(nameOrAddress: string): Promise<string | undefined> {
-  if (isAddress(nameOrAddress)) {
-    return (await lookupAddresses([nameOrAddress]))[nameOrAddress];
-  }
+  const name = isAddress(nameOrAddress)
+    ? (await lookupAddresses([nameOrAddress]))[nameOrAddress]
+    : nameOrAddress;
 
-  if (!nameOrAddress.includes('.')) return undefined;
+  if (!name?.includes('.')) return undefined;
 
   try {
-    return ens_normalize(nameOrAddress);
+    return ens_normalize(name);
   } catch {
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

PR #596, opened 2026-08-20
fix(basename): use one normalizer for validation and hashing

### LINKED ISSUE #576: fix(basename): one unnormalizable .base.eth label rejects the whole resolveNames batch
## Today

`src/resolvers/address/basename.ts:69-73`:

```ts
const pairs = handles
  .map(normalizeBasename)      // ens_normalize(x) === x gate, @adraffy/ens-normalize@1.10.1
  .filter(Boolean)
  .map(handle => [handle, namehash(handle)] as const);   // @ethersproject/hash@5.8 — outside any try
```

`@ethersproject/hash` vendors its own frozen snapshot of ens-normalize (`lib/ens-normalize/`, "near carbon-copy of the original source"), independent of the `@adraffy/ens-normalize@1.10.1` that `normalizeBasename` calls. The two disagree on newer codepoints. Probed against the installed versions — these all pass `ens_normalize(x) === x` but throw in `namehash(x)`:

- `🫨.base.eth`, `🫷.base.eth`, `🩷.base.eth`, `🫎.base.eth`, `🫏.base.eth`, `🐦‍🔥.base.eth` → `Disallowed codepoint: 0x1FAE8` (etc.)

Reproduced on the real module: `resolveNames(['🫨.base.eth', 'jesse.base.eth'])` throws `Disallowed codepoint: 0x1FAE8`, so the whole batch rejects — the valid sibling resolves to nothing and the rejection is `capture()`d to Sentry as a resolver failure. Same bug class as #564, one resolver over. `lookupAddresses` is unaffected (its `normalizeBasename` call is already inside a try/catch).

## Change

Same shape as #573: hash per handle, drop the ones that throw, continue with the rest.

```ts
const pairs = handles
  .map(normalizeBasename)
  .filter(Boolean)
  .flatMap(handle => {
    try { return [[handle, namehash(handle)] as const]; } catch { return []; }
  });
```

## Acceptance

- Unit: `resolveNames(['🫨.base.eth', '<valid>.base.eth'])` → valid one resolves, no throw, `capture` not called.
- Unit: batch of only such labels → `{}` without any RPC.
- Existing basename fixtures unchanged.

Found while reviewing #573.

tony8713:
`normalizeBasename` decides which handles are valid using `@adraffy/ens-normalize`, but the namehash came from `@ethersproject/hash`, which re-normalizes the name against its own vendored copy of an older ens-normalize. The two disagree on recently assigned codepoints, so a handle the gate accepts can make the hasher throw. In `resolveNames` that throw sits outside any try/catch and rejects the whole batch, so valid sibling handles resolve to nothing and the rejection is reported as a resolver failure. `getAvatar` throws on the same input.

Hash with viem's `namehash` instead, as the Space ID resolver already does. It hashes the labels it is given rather than applying a second, independent normalization, so `normalizeBasename` stays the single authority on what a valid Basename is, and the throw goes away at every call site in the module rather than at one of them.

On approach: #576 proposes catching the hasher's throw and dropping those handles. That keeps them unresolvable even though they are valid current ENS names, and does not cover `getAvatar`. Say the word if you would rather have the catch-and-drop shape.

Closes #576

### DIFF
--- a/src/resolvers/address/basename.ts
+++ b/src/resolvers/address/basename.ts
@@ -1,7 +1,7 @@
 import { ens_normalize } from '@adraffy/ens-normalize';
 import { getAddress } from '@ethersproject/address';
-import { namehash } from '@ethersproject/hash';
 import snapshot from '@snapshot-labs/snapshot.js';
+import { namehash } from 'viem/ens';
 import { EMPTY_ADDRESS, isEvmAddress } from '../../helpers/address';
 import { getUrl } from '../../helpers/http';
 import { batchContractCalls, getProvider } from '../../helpers/provider';
--- a/test/unit/resolvers/address/basename.test.ts
+++ b/test/unit/resolvers/address/basename.test.ts
@@ -1,18 +1,26 @@
 import { Interface } from '@ethersproject/abi';
-import { namehash } from '@ethersproject/hash';
 import { StaticJsonRpcProvider } from '@ethersproject/providers';
-import { lookupAddresses, resolveNames } from '../../../../src/resolvers/address/basename';
+import { namehash } from 'viem/ens';
+import {
+  getAvatar,
+  lookupAddresses,
+  resolveNames
+} from '../../../../src/resolvers/address/basename';
 
 const ADDRESS_WITH_NAME = '0x2211d1D0020DAEA8039E46Cf1367962070d77DA9';
 const SECOND_ADDRESS_WITH_NAME = '0x5b76f5B8fc9D700624F78208132f91AD4e61a1f0';
+const EMOJI_ADDRESS_WITH_NAME = '0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6';
 const ADDRESS_WITHOUT_NAME = '0x0C67A201b93cf58D4a5e8D4E970093f0FB4bb0D1';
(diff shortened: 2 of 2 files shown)


tony8713:
Two concrete data points on the approach, since they decide between the two shapes.

`🐦‍🔥.base.eth` is a live registered Basename. Against the real resolver this branch returns `0xCe46fd64D9e9161D8E694f7265b24A6d395E1dC2` for it; master throws on it, and the catch-and-drop shape from the issue resolves it to nothing. So the labels the old hasher rejects are not all unregistered, and dropping them loses real names.

The avatar path is broken for that same address today, and catch-and-drop does not reach it. `lookupAddresses` hands back `🐦‍🔥.base.eth` happily, then `getAvatar` hashes it and throws `Disallowed codepoint: 0x200D`, which `src/resolvers/image/basename.ts` turns into a captured resolver failure. On this branch it returns cleanly instead.

Left alone deliberately: `@ethersproject/hash` is now declared in package.json without anything importing it. I did not drop it here, since it would have to come back if you prefer the catch-and-drop shape, and it is still pulled in transitively either way. Happy to remove it in a follow-up.

wa0x6e (approved):
utAck

PR #597, opened 2026-08-20
fix(lookupDomains): bound the ENS account domains query at 1000

### LINKED ISSUE #575: fix(lookupDomains): paginate ENS account domains/wrappedDomains query
## Today

`src/resolvers/lookupDomains/ens.ts`: the `account(id)` query selects `domains { … }` and `wrappedDomains { … }` with no `first`, so graph-node's default page size (100) applies to each nested list. Addresses holding more than 100 names silently lose the overflow before any label decoding happens.

Measured live against `constants.ensSubgraph["1"]` for `0x9a75ed8e1e592c2e2b0d3eddee8404dcf326a8c5`:

| query | `domains` returned |
|---|---|
| no `first` (current) | 100 |
| `first: 1000` | 982 |
| `first: 5000` | rejected — graph-node max is 1000 |

882 of 982 domains (90%) dropped for that one address. `wrappedDomains` (76 here) is equally unbounded.

Sibling resolvers in the same directory (`shibarium.ts`, `unstoppableDomains.ts`) paginate exhaustively; `ens.ts` is the one holdout.

## Change

Minimum: add `first: 1000` to both nested lists — closes the gap for accounts up to 1000 per list.

```graphql
account(id: $id) {
  domains(first: 1000) { name expiryDate }
  wrappedDomains(first: 1000) { name expiryDate }
}
```

Full fix: cursor/skip pagination like the sibling resolvers, for accounts above 1000.

Note for #571's batching: `hashes.length` is currently bounded ≤200 by the 100+100 cap; raising the caps pushes that to ≤2000, past graph-node's `first` max of 1000 on the `registrations` query — the batch would need chunking (or its own pagination) once this lands.

## Acceptance

- Address above holding >100 names returns all of them (up to the chosen cap) from `lookupDomains`.
- Unit test with mocked `graphQlCall` asserting `first` is sent on both nested lists.

Found during review of #571.

tony8713:
`lookupDomains/ens` asked the ENS subgraph for `account(id) { domains, wrappedDomains }` with no `first`, so graph-node applied its default nested page size (100) to each list and the overflow was dropped before any label decoding happened. An address holding more than 100 names in either list silently lost the rest.

Both nested lists now carry `first: 1000`. That's the subgraph's own ceiling, not a chosen round number: `first: 5000` against this endpoint is rejected outright (`The 'first' argument must be between 0 and 1000`), so 1000 is the most it will ever hand back in one call. An account past that per list is treated as an edge case rather than something this resolver walks through with cursor pagination; this stays a single request. When either list comes back exactly at the cap, `capture` records the address and how many names it did return, so a truncated answer stays visible instead of merely plausible.

Closes #575

### DIFF
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -1,3 +1,4 @@
+import { capture } from '@snapshot-labs/snapshot-sentry';
 import constants from '../../constants.json';
 import { graphQlCall } from '../../helpers/graphql';
 import { Address, Handle } from '../../helpers/types';
@@ -5,6 +6,8 @@ import { Address, Handle } from '../../helpers/types';
 export const NAME = 'Ens';
 export const DEFAULT_CHAIN_ID = '1';
 export const CHAIN_IDS = Object.keys(constants.ensSubgraph);
+
+const ACCOUNT_DOMAINS_LIMIT = 1000;
 
 type Domain = {
   name: string;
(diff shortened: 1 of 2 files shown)


tony8713:
Heads up for whichever of this and #618 lands second: #618 fixes the crash from a nameless ENS domain row by guarding `domain.name` in this same filter before it's rewritten here. This branch's rewrite of the query still leaves `!domain.name.endsWith('.addr.reverse')` unguarded, so merging this after #618 (or rebasing it) needs to carry that guard forward or the null-name crash comes back.

tony8713:
Rebased onto master (#619 merged). Confirmed the guard did not carry over on its own: the first commit of this branch rewrites the same filter line #619 touches, so rebasing produced a real textual conflict there, and PR597's own side of that conflict had no null check. Resolved it by hand, keeping this branch's pagination logic plus `domain.name &&`. Mutation-tested afterward: removing the guard again reddens the new "drops a nameless row instead of throwing" unit test with the same `TypeError: Cannot read properties of null (reading 'endsWith')` #618 hit, and nothing else in the related test set. New head c4aa17f.

tony8713:
Correction to my previous comment: pushing a21524a. CI on c4aa17f caught something the null-name guard proof missed: rebasing had also silently reverted two label-lookup fixes master picked up after this branch forked (03b92e4, 66cdb98), by keeping this branch's own `registrations(id_in)` query over master's `domains(labelhash_in)`. The former misses hashed labels sitting at a subdomain position or on a DNS-imported name, which is exactly what `test/unit/helpers/graphqlEnvelope.test.ts` caught (`data.registrations is not iterable`, since that shared test now pins master's query shape). Re-resolved by keeping master's label-lookup code verbatim and layering only this branch's actual contribution, the account-level pagination, on top. Null-name guard still in place and still mutation-tested.

tony8713:
CI green on a21524a: Lint pass, Test `Test Suites: 1 skipped, 56 passed, 56 of 57 total` / `Tests: 4 skipped, 14 todo, 488 passed, 506 total`, zero failures including the two live-API and Starknet suites this box can't clear locally. Ready for re-review.

wa0x6e (approved):
utAck

PR #598, opened 2026-08-20
fix(image): reject non-image HTTP responses

### LINKED ISSUE #587: fix(image): non-image 200 answers reach sharp and get reported
Sentry: [STAMP-71](https://snapshot-labs.sentry.io/issues/STAMP-71) — 19 events since 2026-08-16 (when #543 turned capture on), two sharp messages grouped together:

```
Error: Input buffer contains unsupported image format
Error: Input buffer has corrupt header: glib: XML parse error: ... Opening and ending tag mismatch: html line 1 and div
```

both thrown from `toWebp` in `helpers/image.ts`, captured by `withResize` in `resolvers/image/index.ts`.

## Root cause

The avatar URL answers **HTTP 200 with `text/html`**. `fetchHttpImage` only checks `response.ok`, so the HTML body is handed to sharp, which rejects it. When the body starts with `<`, libvips routes it to the SVG loader and produces the "XML parse error ... html" variant — same cause.

Every input behind the 19 events, checked live:

| input | avatar record | upstream answer |
|---|---|---|
| `she256.eth` (9 events) | `https://drive.google.com/file/d/…/view?usp=sharing` | 200 `text/html`, Drive share page |
| `tinfoilwookie.eth` (5) | `https://ipfs.io/ipfs/Qme2JM…/4230` | gateway hangs / non-image |
| `inkstone.eth`, `southernallstars.eth`, `grendel.eth` | `https://opensea.io/assets/…` | 200 `text/html`, asset page |
| `¤¤¤.eth` | `https://app.frax.finance/static/media/….jpg` | host dead |
| space `dezukidao.eth` | `https://dezuki.com/_next/static/images/logo….png` | 200 `text/html`, 114-byte JS redirect stub |

User-supplied garbage, not an outage. Nothing to log further; the inputs are fully characterised.

## Fix

In `helpers/http.ts` `fetchHttpImage`, treat a non-`image/*` content type as no-data:

```ts
const type = response.headers.get('content-type') ?? '';
if (!type.startsWith('image/')) throw httpError(new URL(url).host, 404, `not an image: ${type}`);
```

404 is already a routine miss for `withFailureContract`, so this answers `false` without a report.

Optionally also silence sharp's `Input buffer contains unsupported image format` / `Input buffer has corrupt header` in `withResize` — covers a server that lies about content type, and the `fetchImageOrMetadata` path in `resolvers/image/starknet.ts` that bypasses `fetchHttpImage`.

## Acceptance

- A 200 `text/html` answer for an avatar URL returns `false` from the resolver without a `capture`.
- A real `image/*` answer still resizes as before.
- One unit case per branch: non-image content type, image content type.

### LINKED ISSUE #609: fix(http): data: profile pictures are rewritten to the IPFS gateway and rejected (STAMP-7E, STAMP-6P)
Sentry: [STAMP-7E](https://snapshot-labs.sentry.io/issues/STAMP-7E) — **51 occurrences**, first seen 2026-08-23 15:39:43 UTC, last seen 2026-08-24 09:41:52 UTC, `environment:production`, `handled:yes`, `provider:starknet`, 0 users, substatus `escalating`.
Sentry: [STAMP-6P](https://snapshot-labs.sentry.io/issues/STAMP-6P) — **48 occurrences**, first seen 2026-06-25 20:19:25 UTC, last seen 2026-08-23 15:24:15 UTC, same tags. Same defect seen through axios; the group split when the native-fetch migration changed the error shape. STAMP-6P is currently filed as #592 under a different diagnosis (see *Sweep* below).

## What is reported

```
Error: [snapshot.4everland.link] 
```

The message is empty because prod runs node v26 over HTTP/2, which has no reason phrase, so `response.statusText` is `''`. The status is **400** — carried on the error object, not in the message.

Ours are the first three frames:

```
src/api.ts:135              Promise.all(currentResolvers.map(r => resolvers[r](address, network, networkId)))  (index 5)
src/resolvers/image/index.ts:29   return await resolve(...args)
src/resolvers/image/starknet.ts:32  const fetched = await fetchImageOrMetadata(getUrl(img_url))
src/helpers/deadline.ts:13        return await fn(controller.signal)
src/helpers/http.ts:32            throw httpError(new URL(url).host, response.status, response.statusText)
src/helpers/errors.ts:8           return Object.assign(new Error(`[${source}] ${message}`), { status, response: { status } })
```

Sample event (`7e24b6165a5543aba0ad034412148425`, 2026-08-24 09:41:52 UTC, pod `stamp-646898499-z6ctd`):

- Request: `GET https://cdn.stamp.fyi/avatar/0x02ff9d1ccff043c05753dbf6c2c47be24947ae5ed429843cd8ec7c774454072d`
- Resolver input `args`: `["0x02ff9d1ccff043c05753dbf6c2c47be24947ae5ed429843cd8ec7c774454072d", "1", null]`
- Client `Bytespider` (crawler)

The last 30 events cover 11 distinct Starknet addresses, all `provider:starknet`, all the same host.

## Root cause

Not the upstream. We build an invalid URL and the gateway rejects it.

`provider.getStarkProfile(address).profilePicture` does not always return an `ipfs://` URL. For on-chain-metadata NFTs (Blobert, Realms) it returns a **`data:` URI**:

```
data:application/json;base64,eyJuYW1lIjoiQmxvYmVydCAjMzQ2NiIsImRlc2NyaXB0aW9uIjoi…
```

`src/resolvers/image/starknet.ts:32` passes it through `getUrl` (`src/helpers/http.ts:40`), which delegates to `snapshot.js`:

```js
// @snapshot-labs/snapshot.js — utils.getUrl
if (!uri.startsWith('ipfs://') && !uri.startsWith('ipns://') &&
    !uri.startsWith('https://') && !uri.startsWith('http://'))
  return `${ipfsGateway}/ipfs/${uri}`;   // <- anything else is assumed to be a bare CID
```

A `data:` URI is not a bare CID, so it is concatenated onto the gateway path:

```
https://snapshot.4everland.link/ipfs/data:application/json;base64,eyJuYW1l…
```

4everland answers **400** with `invalid path "/ipfs/data:application/json;base64,…"`. `src/helpers/http.ts:34` turns that into `httpError(host, 400, statusText)` and the resolver wrapper reports it: `isRoutineMiss` (`src/resolvers/image/index.ts:33`) only forgives 404 and `INVALID_ARGUMENT`, and `isSilencedError` (`src/helpers/errors.ts:10`) matches 504/429/timeout codes — 400 is in neither.

### Reproduced (2026-08-24, mainnet via `rpc.snapshot.org`)

```
getStarkProfile(0x02ff9d1c…) → cumberlord.stark
  profilePicture = data:application/json;base64,eyJuYW1lIjoiQmxvYmVydCAjMzQ2Ni…
getUrl(...)      → https://snapshot.4everland.link/ipfs/data:application/json;base64,…
fetch(...)       → 400 Bad Request, content-type text/plain
  body: invalid path "/ipfs/data:application/json;base64,eyJuYW1lIjoiQmxvYmVydCAjMzQ2Ni…"
```

All 11 addresses from the recent events resolve to `data:application/json;base64,…` profile pictures — `revilotc.stark`, `delegate.gislegob.stark`, `cheelax.stark`, `bal7hazar.stark`, `recipromancer.stark`, `romulus.stark`, `classicj.stark`, `bob0005.stark`, `fortunaregem.stark`, `fabiocatalao.stark`, `rmz.stark`. So does `hugooo.stark` (`0x0044315e…`), the sample address on STAMP-6P.

Note: `getStarkProfile` returns the URI with a trailing `` from the felt decode. Undici's base64 parser tolerates it; it is not the cause.

## Impact

No user-visible breakage: `api.ts` takes the first resolver that answers and falls back to a blockie, so these addresses get a generic avatar. **But their real avatar has never rendered** — every Starknet ID whose PFP is an on-chain-metadata NFT is a permanent miss, not a transient failure. That half is the actual bug; the Sentry noise is the symptom.

Rate: 51 events in ~18h, against ~65k HTTP requests/24h across the service (Prometheus, `http_request_duration_seconds_count{instance=~"stamp.fyi:.*"}`, 2026-08-24). A per-resolver denominator is not available — the per-resolver metric was reverted in 06b5810 — so the share of Starknet resolutions that fail this way is unmeasured. The events cluster in crawler bursts (Bytespider), so the count tracks crawl volume, not user traffic.

## Provenance

Not a regression. `getUrl` has mangled `data:` URIs for as long as the Starknet resolver has existed; STAMP-6P has been recording it since 2026-06-25.

What is new is that it is *reported*: `12a1e52 feat(resolvers): report resolver failures` added `capture()` to the resolver failure contract and first deployed 2026-08-23 15:38 (`app_start_time` on the first event). The same deploy carried the axios→fetch migration, which changed the error shape and split the group — STAMP-6P's last event is 15:24:15, STAMP-7E's first is 15:39:43, same day.

## Suggested fix

`data:` needs no gateway, and native fetch reads it directly. `src/helpers/http.ts:40`:

```ts
export function getUrl(url) {

  return snapshot.utils.getUrl(url, gateway);
}
```

Verified end-to-end with that change (2026-08-24):

```
fetch(data:application/json;…)     → 200, content-type application/json
  → metadata.image = data:image/svg+xml;base64,PHN2ZyB4bWxucz…
fetch(data:image/svg+xml;…)        → 200, content-type image/svg+xml, 2607 bytes
sharp(buf).resize(500,500).webp()  → 15112 bytes
```

So the fix makes these avatars render, it does not merely mute the error. `AbortSignal` works on a `data:` fetch, so `withDeadline` is unaffected.

Both content-type gates in #598 pass on that chain (`application/json` on the metadata branch, `image/svg+xml` on `readHttpImage`), but #598 rewrites the surrounding lines of `src/helpers/http.ts`, so expect a conflict depending on merge order.

Out of scope, unmeasured: the same `getUrl` branch mangles every other non-CID scheme it does not know (`ar://`, a bare path). No Sentry group currently points at one.

## Sweep

- `audit.sh -R stamp` over all 20 unresolved Sentry ids; positive control STAMP-75 returns TRACKED. STAMP-7E is UNFILED.
- Only STAMP-6P shares this root cause. It is filed as **#592** (`fix(image): broken third-party avatar URLs are reported as resolver errors`), which diagnoses it as a *third-party* URL failing and proposes silencing it. That diagnosis does not hold for this address: the URL is ours to build and we build it wrong.
- **#592's PR #595 (unreviewed) widens `isRoutineMiss` to treat any 4xx as a routine miss.** That mutes STAMP-7E and STAMP-6P without rendering a single avatar, and removes the only signal pointing at this bug. Worth landing this fix first, or at least not reading the silence as resolution.
- **PR #598 (reviewed)** restructures `fetchImageOrMetadata` and `src/helpers/http.ts`, but leaves `getUrl` untouched — it does not fix this.
- PR #604 (STAMP-74) touches the same resolver but the RPC call, not the image fetch.
- Checked the remaining unresolved ids (STAMP-7A, 7B, 7D, 79, 6Z, 6R, 6Q); none share this cause.

tony8713:
Image resolvers took any successful HTTP response as image bytes, so an HTML error page reached sharp and surfaced as an upstream failure rather than a routine miss. Rejected responses also left their stream open, and the Starknet path buffered the whole body before it looked at the declared type, so a response that never ends never reached the check at all.

A zero-length 200 was a third way through, because an empty Buffer is truthy: it passed the emptiness guards and reached sharp, or on the resolvers that do not resize, the request handler itself, where nothing catches it. The Starknet metadata branch handed `JSON.parse`'s result back untyped, so a body that was not the expected object left the resolver as a TypeError or a SyntaxError.

Reject a declared type that is not `image/*`, compared case insensitively, read the headers before consuming the body, and cancel the body on every path that does not read it. Reject a zero-length body, and read a metadata body as no data unless it carries a string `image`. Both rejections are routine misses rather than something to report, and an unusable metadata body is simply no data. A response that declares no type at all is still accepted so long as it has bytes, because real avatars are served that way. Valid images and Starknet JSON metadata keep their existing paths.

Folded in from #610, per review: a Starknet profile picture that is a `data:` URI rather than a bare CID was rewritten onto the IPFS gateway path and rejected there, so an on-chain-metadata NFT's avatar never rendered even though the resolver reported nothing wrong. `getUrl` now passes a `data:` URI through unchanged.

Also addresses three follow-up review findings. `readHttpImage` attributed a redirect's rejection to the host that was asked rather than the one that actually answered. The Starknet metadata gate matched only a bare `application/json` prefix, missing the `+json` structured suffix and `text/json` that real NFT metadata is served as, so it fell through to the image gate and was silently dropped. And a Starknet image unit test mocked `getAddressFromStarkName`, which no case in that file exercises.

Closes #587
Closes #609


tony8713:
The red Test job here is not this branch. `rpc.snapshot.org` is answering `starknet_chainId` with `429: Monthly capacity limit exceeded`, so the four Starknet image snapshots and the five Starknet address resolver cases fail. Re-running master's own Test workflow on df5f1cf just now gave exactly the same nine names, and this branch is 432 passed against master's 424, which is the eight tests it adds.

Worth knowing separately: that same quota means Starknet avatars are almost certainly failing in production right now.

src/helpers/http.ts
```diff
@@ -25,16 +17,28 @@ export function spaceIds(id: string): string[] | null {
...
-    if (!response.ok) throw httpError(new URL(url).host, response.status, response.statusText);
+  const type = response.headers.get('content-type');
+  if (type && !type.toLowerCase().startsWith('image/')) {
+    await response.body?.cancel();
+    throw httpError(new URL(url).host, 404, `not an image: ${type}`);
+  }
 
-    return Buffer.from(await response.arrayBuffer());
-  }, IMAGE_FETCH_BUDGET);
+  return Buffer.from(await response.arrayBuffer());
```
wa0x6e: The media-type gate covers "declared something that isn't an image", but there's no zero-length cell: a `200` with an empty body and either no `Content-Type` or an `image/*` one still returns as image bytes.

`Buffer.from(await response.arrayBuffer())` on a zero-byte body is an empty-but-*truthy* `Buffer`, so `withResize`'s `if (!input) return false` guard at `src/resolvers/image/index.ts:55` doesn't fire. Two outcomes downstream:

- `resize: true` resolvers — sharp throws `Input Buffer is empty`, and the catch at `src/resolvers/image/index.ts:61` calls `capture(err)` unconditionally, with no `isSilencedError`/`isRoutineMiss` filter (unlike `withFailureContract`). That's a Sentry event per hit, i.e. the same amplification this PR removes for the HTML case.
- `resize: false` resolvers (`user-cover`, `space-cover`, `space-logo`, `space-cover-sx`) — the empty Buffer never passes through `withResize` at all. `files.find(Boolean)` at `src/api.ts:142` picks it up and it reaches the unguarded `resize()` at `src/api.ts:154`, outside any try/catch in the async handler.

That second path is the reason this is worth fixing in the same change rather than as a follow-up: `express@^4.17.1` doesn't forward async rejections to error middleware, and there's no `unhandledRejection` handler in the repo or in `@snapshot-labs/snapshot-sentry`, so under Node's default the throw is an unhandled rejection. The trigger is a user-controlled avatar/cover URL answering `200` with no body — worth confirming that path yourself, since it's a bigger claim than the Sentry-noise one.

A length check next to the media-type check keeps it a routine miss, consistent with every other "not an image" case here:

```suggestion
  const buffer = Buffer.from(await response.arrayBuffer());
  if (buffer.length === 0) {
    throw httpError(new URL(url).host, 404, 'empty body');
  }

  return buffer;
```

src/resolvers/image/starknet.ts
```diff
@@ -25,17 +25,16 @@ async function getImage(domainOrAddress: string): Promise<string | null> {
...
-  if (contentType.includes('application/json')) {
-    return JSON.parse(data.toString('utf-8'));
-  }
-  return data;
+  return withDeadline(async signal => {
+    const response = await fetch(url, { signal });
+    const type = response.headers.get('content-type') ?? '';
+
+    if (response.ok && type.toLowerCase().startsWith('application/json')) {
+      return JSON.parse(await response.text());
```
wa0x6e: `JSON.parse` returns `any`, and it's silently widened here to the declared `Buffer | { image?: string }` — so `tsc` can't catch a body that doesn't match that shape. Three shapes escape `resolve()` as untyped throws, and since they carry no `status`, `code`, or `cause.code`, neither `isSilencedError` nor `isRoutineMiss` matches them, so `capture()` fires. That's the same class of upstream-garbage Sentry noise this PR removes for HTML bodies, on the sibling branch of the same function.

Measured against the resolver with a valid felt address:

```
body 'null'          => TypeError: Cannot read properties of null (reading 'image')
body '{"image":42}'  => TypeError: uri.startsWith is not a function   (from snapshot.utils.getUrl)
body 'not json'      => SyntaxError: Unexpected token 'o', ... is not valid JSON
body '[1,2,3]'       => resolves false   (safe)
body '"a string"'    => resolves false   (safe)
```

The third one is why the guard has to wrap the parse rather than just validate its result.

On the shape of the check — `typeof metadata?.image === 'string'` rather than a truthiness or null check, because that matches `getUrl`'s actual contract:

- falsy values (`""`, `null`, `undefined`) never reach it: `fetched.image ?` on line 48 short-circuits and returns `false` already;
- a truthy *non-string* (`42`, `{}`, `["a"]`) hits `uri.startsWith(...)` inside `getUrl` with no type check, and throws;
- a string that isn't a URL is already safe — `getUrl` prefixes anything unrecognized with the IPFS gateway, so it fetches, fails, and comes back as an `httpError` with a status, i.e. a routine miss. Worth keeping that degradation rather than rejecting it early.

```suggestion
      // The profile owner controls this URL and the JSON it serves: `null`, a number
      // under `image`, or a body that is not JSON at all must read as no-data here
      // rather than a TypeError/SyntaxError out of `resolve`.
      try {
        const metadata = JSON.parse(await response.text());
        return typeof metadata?.image === 'string' ? { image: metadata.image } : {};
      } catch {
        return {};
      }
```

Note this narrows the return to just `image`; `resolve()` only reads that field today, so nothing is lost, but returning `metadata` unchanged when the shape checks out would work equally well if you'd rather not drop the other fields.

test/unit/resolvers/image/starknet.test.ts
```diff
@@ -1,15 +1,30 @@
...
+
+beforeEach(() => {
+  mockGetStarkProfile.mockReset().mockResolvedValue({
+    profilePicture: 'https://example.com/avatar'
+  });
+});
+
+afterEach(() => {
+  jest.restoreAllMocks();
+});
```
wa0x6e, line 37: The new file-level `beforeEach` gives `mockGetStarkProfile` a default resolved profile, which now also applies to the four pre-existing address-validation tests above. Those stub no `fetch` of their own, and they exist to assert `expect(mockGetStarkProfile).not.toHaveBeenCalled()` — so with a working profile behind the gate, a regression gets past that line before the assertion runs.

Stubbing `isStarknetFelt` to `() => true` to simulate a gate regression:

```
master shape (no default)   ● expect(jest.fn()).not.toHaveBeenCalled()
                              Received number of calls: 1
                              1: "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7"

this branch (with default)  ● Received promise rejected instead of resolved
                              Rejected to value: [Error: [example.com] Not Found]
```

The suite still fails either way, so nothing is hidden — the cost is the lost diagnostic and live outbound HTTPS from `test/unit/`, multiplied by four per case by `jest.retryTimes(3)` in `test/setup-jest.ts:15`.

Pinning `fetch` in the same hook keeps the failure offline and on the right assertion. Two things fall out for free: `.mockReset()` is redundant (`clearMocks: true` is already set in `jest.config.js:10`, and it's the same construct `6ed20ab` removed from this file), and scoping the restore to the fetch spy stops `jest.restoreAllMocks()` from also restoring the `console.log` spy from `test/setup-jest.ts:13`, which currently makes this file print redis chatter that every other suite suppresses.

```suggestion
let fetchSpy: jest.SpyInstance;

beforeEach(() => {
  mockGetStarkProfile.mockResolvedValue({ profilePicture: 'https://example.com/avatar' });
  fetchSpy = jest.spyOn(global, 'fetch').mockRejectedValue(new Error('unexpected fetch'));
});

afterEach(() => {
  fetchSpy.mockRestore();
});
```

Verified: 8/8 still pass with this applied, and under the stubbed-open gate the three cases fail offline with `Rejected to value: [Error: unexpected fetch]`.

test/integration/helpers/http.test.ts
```diff
@@ -8,16 +8,66 @@ const BODY = Buffer.from('as much of an image as the fetch cares about');
...
 let missingUrl: string;
+let nonImageUrl: string;
+let mixedCaseUrl: string;
+let undeclaredUrl: string;
 let neverEndingUrl: string;
+let slowErrorUrl: string;
+let nonImageClosed!: Promise<void>;
+let resolveNonImageClosed!: () => void;
+let slowErrorClosed!: Promise<void>;
+let resolveSlowErrorClosed!: () => void;
```
wa0x6e, line 31: These two are declared with a definite-assignment `!` but only assigned inside their own `it()` body, while the `beforeAll` request handler reads the outer-scope bindings at *request* time via `streamForever(res, resolveNonImageClosed)`.

Any hit to `/not-an-image.png` or `/slow-error.png` outside its owning test — a retry, a reorder, a stray second request — calls `undefined()` inside the `res.on('close', ...)` handler. That surfaces as an uncaught throw inside the HTTP server rather than a test failure, which is a confusing thing to debug.

Initialising both to a no-op closes the hole without changing what the tests assert:

```suggestion
let resolveNonImageClosed: () => void = () => {};
let slowErrorClosed!: Promise<void>;
let resolveSlowErrorClosed: () => void = () => {};
```

wa0x6e (changes requested):

tony8713:
Not from this review, but worth flagging before either of these merges.

This branch was exporting `IMAGE_FETCH_BUDGET` from `src/helpers/http.ts` and importing it into `image/starknet.ts`. That is the ending you asked for on #457 on the 21st and then explicitly retracted on the 22nd, in https://github.com/snapshot-labs/stamp/pull/457#discussion_r3834873179:

> It told you to export `IMAGE_FETCH_BUDGET` and import it here and in `image/starknet.ts`. That is the wrong ending; take this one instead. [...] So: **delete `IMAGE_FETCH_BUDGET` and leave three bare `5e3`.**

#457's head has no such constant any more. I have taken the same ending here: `fetchHttpImage` and `fetchImageOrMetadata` each carry a bare `5e3`, and nothing is exported. Flagging it because the review here did not raise it, and whichever of the two lands second would otherwise reintroduce the shape you rejected.

The two still conflict independently of that. Both rewrite the image read in `src/helpers/http.ts` and `fetchImageOrMetadata` in `src/resolvers/image/starknet.ts`, and both delete `axiosDefaultParams`, so whichever lands second rebases onto the first.

test/unit/resolvers/image/starknet.test.ts
tony8713: Confirmed both halves before applying. With `isStarknetFelt` stubbed to `() => true`, the three `not.toHaveBeenCalled` cases give exactly what you measured: `Rejected to value: [Error: [example.com] Not Found]` on this branch against `Received number of calls: 1` on the master shape. That rejection is a real outbound 404 off example.com from `test/unit/`, so the live-HTTPS half holds too.

Applied your hook as written. The console half is visible in the run: before, this file printed `[redis] Redis connect / ready / end` that `http.test.ts` in the same run did not, and after it prints only the one `Connecting to Redis` line every suite gets from `setup-jest`'s own import.

test/integration/helpers/http.test.ts
tony8713: Reproduced. An `it` placed ahead of the owning tests that does nothing but fetch `/not-an-image.png` makes the file report `PASS ... 9 passed, 9 total`, and then the process dies with `TypeError: resolveClosed is not a function at ServerResponse.<anonymous> (test/integration/helpers/http.test.ts:28:9)`, thrown out of `ServerResponse.emit('close')` after the reporter has finished printing. Confusing is right.

Applied the no-op initialisers as written. `nonImageClosed` keeps its `!`, since nothing but its owning test reads it.

src/helpers/http.ts
tony8713: Reproduced the whole chain before applying, since you flagged the second half as the bigger claim.

The empty buffer is truthy so `!input` does not fire, and sharp on it throws `Input Buffer is empty`. For the `resize: false` half I stood up express 4.19.2 (what `^4.17.1` resolves to here) with an async handler that throws and an error middleware behind it: the middleware is never reached, the process logs `UNHANDLED REJECTION: Input Buffer is empty`, and the client request hangs with no response at all. `rg unhandledRejection` finds nothing in `src/`, `test/` or `@snapshot-labs/snapshot-sentry`, so that is Node's default.

Applied the length check as written. Two integration cases pin it, one route answering 200 with `image/png` and an empty body, one answering 200 with no `Content-Type` at all.

Mutation check at the pushed head: deleting the length check fails exactly those two out of the 27-file related set, `Tests: 2 failed, 229 passed` against a clean `231 passed`.

src/resolvers/image/starknet.ts
tony8713: Reproduced your table verbatim against the resolver, same five outcomes including the two safe ones, and confirmed the Sentry half by putting the three throws through `isSilencedError` and a copy of `isRoutineMiss`: all three come back false on both, so `capture` fires.

Applied the guard as written, including `typeof metadata?.image === 'string'` and the narrowed `{ image }` return, for the `getUrl` reasons you gave.

Two deviations worth calling out. The first is that I moved `await response.text()` out of the try and left only `JSON.parse` inside it. As written the catch also swallows the deadline abort: against a metadata body that opens and never closes, the suggestion resolves `{}` where reading outside the try throws `AbortError`, which is the one thing `isSilencedError` is built to recognise by name. So a stalled upstream stopped being distinguishable from one that answered with something unusable. There is a test on that now.

The second is that I dropped the three-line comment and put the shapes in the unit file instead: `null`, `{"image":42}`, `not json at all` and `{"name":"nft"}`, all asserting `false`. That is the same thing the comment was holding, except it holds in CI. Say the word if you would rather have both.

Mutation checks at the pushed head, against the 27-file related set. Reverting to the bare `JSON.parse` fails exactly the first three shape cases and leaves `{"name":"nft"}` and the abort case green, `Tests: 3 failed, 228 passed` against a clean `231 passed`. Moving the body read back inside the try fails only the abort case, on `Expected: "AbortError", Received: undefined`, because the resolver resolves `false` instead.

wa0x6e:
**Review-panel suggestion: fold #610 into this PR and close it as superseded.**

#610 (`fix(http): data: profile pictures are rewritten to the IPFS gateway and rejected`) touches `src/helpers/http.ts` and `test/unit/resolvers/image/starknet.test.ts` — both already in this PR's diff — so folding it in adds no new review surface (+18/−1, one commit).

The two changes are entangled rather than merely adjacent. Traced end to end using a `data:application/json` profile picture whose `image` is a `data:image/svg+xml` URI (the shape of the live input behind STAMP-7E, Realms World "Blobert #3466"):

```
#598 alone   -> httpError(snapshot.4everland.link, 400) => Sentry capture
#598 + #610  -> image returned (image/svg+xml)
```

`fetch()` handles `data:` URIs natively and derives the content type from the URI itself — measured: `data:image/svg+xml;base64,…` → `200 image/svg+xml`, `data:image/png;base64,…` → `200 image/png`. So once `getUrl` stops rewriting them onto the gateway path, both hops of the Starknet chain work and this PR's new `image/*` gate accepts the result.

Two practical consequences of keeping them apart:

- Whichever merges second hits a conflict in `test/unit/resolvers/image/starknet.test.ts`. Verified by rebasing this branch onto `master` + #610: `src/helpers/http.ts` auto-merges cleanly, the test file conflicts at commit 2 of 10.
- Any tightening of the metadata guard at `src/resolvers/image/starknet.ts:37` has to be written knowing whether `data:` survives `getUrl`. A guard requiring an absolute `http(s)`/`ipfs` URL would reject the exact value #610's new test asserts.

If you fold it in: add `Closes #609` alongside `Closes #587`, and port #610's new test onto this PR's mock setup.

Unrelated but worth recording: this branch rebases cleanly onto current `master` (786c16c) — 10/10 commits replayed, no conflicts, `tsc --noEmit` clean, and both touched test files green (25/25).

src/helpers/http.ts
```diff
@@ -17,21 +17,28 @@ export function spaceIds(id: string): string[] | null {
...
+  if (type && !type.toLowerCase().startsWith('image/')) {
+    await response.body?.cancel();
+    throw httpError(new URL(url).host, 404, `not an image: ${type}`);
+  }
 
-    if (!response.ok) throw httpError(new URL(url).host, response.status, response.statusText);
+  const buffer = Buffer.from(await response.arrayBuffer());
+  if (buffer.length === 0) {
+    throw httpError(new URL(url).host, 404, 'empty body');
+  }
```
wa0x6e, line 46: `readHttpImage` derives the host from the *requested* `url` in three places, but after a redirect the response comes from a different host — so a redirected miss reports the host that was asked, not the one that answered. It also re-parses `new URL(url)` once per branch.

Measured against a local redirecting server with Node 22's global fetch: after a 302, `response.url` is the final URL while the `url` parameter stays the originally-requested one.

Keeping the `url` parameter as a fallback is deliberate rather than dropping it — a hand-built `new Response(...)` has an empty `.url` (the constructor can't set it), and the unit tests here construct responses that way.

```suggestion
export async function readHttpImage(url: string, response: Response): Promise<Buffer> {
  // A redirected response answers from a different host than the one requested.
  const host = new URL(response.url || url).host;

  if (!response.ok) {
    await response.body?.cancel();
    throw httpError(host, response.status, response.statusText);
  }

  const type = response.headers.get('content-type');
  if (type && !type.toLowerCase().startsWith('image/')) {
    await response.body?.cancel();
    throw httpError(host, 404, `not an image: ${type}`);
  }

  const buffer = Buffer.from(await response.arrayBuffer());
  if (buffer.length === 0) {
    throw httpError(host, 404, 'empty body');
  }
```

src/resolvers/image/starknet.ts
```diff
@@ -23,12 +24,24 @@ async function getImage(domainOrAddress: string): Promise<string | null> {
...
-    response.headers.get('content-type')?.includes('application/json')
-      ? await response.json()
-      : Buffer.from(await response.arrayBuffer())
-  );
+async function fetchImageOrMetadata(url: string): Promise<Buffer | { image?: string }> {
+  return withDeadline(async signal => {
+    const response = await fetch(url, { signal });
+    const type = response.headers.get('content-type') ?? '';
+
+    if (response.ok && type.toLowerCase().startsWith('application/json')) {
```
wa0x6e: The JSON classifier uses a `application/json` prefix check, which misses the `+json` structured suffix (`application/ld+json`, `application/vnd.api+json`) and `text/json` — all shapes real NFT metadata is served as. Those bodies fall through to `readHttpImage` and are rejected as "not an image", which becomes a synthetic 404 and is swallowed as a routine miss, so the avatar silently disappears with nothing logged.

Splitting on `;` first also makes the check robust to a `charset` parameter without relying on prefix matching.

(The related concern about leading whitespace in the header is **not** an issue — measured with a server sending `Content-Type:    application/json`, Node's http layer strips the optional whitespace before `Headers.get()` returns it, so `.startsWith` vs `.includes` makes no difference there.)

```suggestion
    const type = (response.headers.get('content-type') ?? '').toLowerCase().split(';')[0].trim();
    const isJson = type === 'application/json' || type === 'text/json' || type.endsWith('+json');

    if (response.ok && isJson) {
```

test/unit/resolvers/image/starknet.test.ts
```diff
@@ -1,24 +1,32 @@
 const mockGetStarkProfile = jest.fn();
 
 jest.mock('../../../../src/helpers/provider', () => ({
-  getProvider: () => ({ getStarkProfile: mockGetStarkProfile })
+  getProvider: () => ({
+    getStarkProfile: mockGetStarkProfile,
+    getAddressFromStarkName: jest.fn()
+  })
```
wa0x6e, line 8: `getAddressFromStarkName` is only reached when `isStarkDomain(domainOrAddress)` is true, i.e. for a `*.stark` name. Every input used in this file (`ADDRESS`, `OVER_PRIME_ADDRESS`, `UNPREFIXED_ADDRESS`, `EVM_ADDRESS`, `UNPADDED_ADDRESS`) is a raw hex address, so that branch never runs and the stub is dead.

Adding a `.stark` domain test would be the other way to resolve it, but domain resolution is outside what this PR changes — dropping the unused member keeps the mock honest about what's covered.

```suggestion
  getProvider: () => ({ getStarkProfile: mockGetStarkProfile })
```

tony8713:
Replying to:

> **Review-panel suggestion: fold #610 into this PR and close it as superseded.**
>
> #610 (`fix(http): data: profile pictures are rewritten to the IPFS gateway and rejected`) touches `src/helpers/http.ts` and `test/unit/resolvers/image/starknet.test.ts` — both already in this PR's diff — so folding it in adds no new review surface (+18/−1, one commit).
>
> [...]
>
> If you fold it in: add `Closes #609` alongside `Closes #587`, and port #610's new test onto this PR's mock setup.

Reproduced independently before folding it in. `getUrl` (`src/helpers/http.ts`) has no `data:` guard on this branch, and `snapshot.js`'s `getUrl` treats anything not starting with `ipfs://`/`ipns://`/`http(s)://` as a bare CID, prepending the gateway path — so a Starknet profile picture that is a `data:` URI does get mangled into `.../ipfs/data:...`, matching what you measured.

Ported the fix and the test rather than a straight cherry-pick, since #610 predates this PR's stricter content-type gate in `fetchImageOrMetadata`/`readHttpImage`: the naive port of the test failed for the wrong reason (a mocked response with a default `text/plain` content type, which #598's own gate now rejects) rather than the `data:`-mangling bug. Gave the mocked image response an explicit `image/svg+xml` type, matching what native `fetch` actually derives from a `data:image/svg+xml` URI, and adapted the test to this file's existing `fetchSpy`/`beforeEach` convention rather than #610's separate `mockGlobalFetch()`.

Confirmed the ported test fails on the unfixed `getUrl` (both calls land on the gateway-mangled URL) and passes once the two-line guard is added. Full related set (28 files) green apart from one unrelated live-network flake in `lens.test.ts`, which doesn't touch `getUrl` at all and passed clean on its own re-run.

Added `Closes #609` alongside `Closes #587` in the PR body.

Pushed at `79d53b8`.

wa0x6e (approved):
utAck

PR #600, opened 2026-08-21
fix(lookupDomains): resolve every hashed label in multi-level ENS names

### LINKED ISSUE #577: fix(lookupDomains): resolve every hashed label in multi-level ENS names
## Today

`src/resolvers/lookupDomains/ens.ts`: `getLabelHash` extracts only the first bracketed segment of a name (`domain.name.match(/\[(.*?)\]/)?.[1]`), and the substitution replaces only that one `[hash]`.

For a name with two unknown labels, `[a].[b].eth`, the code looks up `0x<a>` — a subdomain labelhash, which is never a `Registration` id (registrations exist only for 2LDs) — so nothing comes back and `[b]` stays hashed even when the subgraph knows the 2LD's label.

Measured live on mainnet (`constants.ensSubgraph["1"]`):
- ≥888 `.eth` domains currently match this shape (`domains(first:1000, where:{name_contains:"].[", name_ends_with:".eth"})`, capped by page size).
- Example: `[27ac19d5…].[b50225e6…].eth` — `registrations(where:{id:"0xb50225e6…"})` → `labelName: "aragonid"`, but the code queries `0x27ac19d5…` → `registrations: []`, so the name is returned unchanged instead of `[27ac19d5…].aragonid.eth`.

Pre-existing; #571 batches the lookup but doesn't touch hash extraction.

## Change

Collect every bracketed hash per domain (`matchAll`), add all of them to the batch, and replace each `[hash]` that resolved. Hashes with no registration (subdomain labelhashes) simply stay as-is, same as today.

## Acceptance

- Unit test with mocked `graphQlCall`: `[a].[b].eth` with a registration for `0xb` → `[a].<label>.eth`; both `0xa` and `0xb` appear in the batched `ids`.
- Existing lookupDomains fixtures unchanged.

Found during review of #571.

tony8713:
`getLabelHash` read only the first bracketed segment of a subgraph name, and the substitution replaced only that one `[hash]`. Every other hashed label in the name was left alone, so a label the subgraph does know still came back hashed whenever it was not the first one.

Collect every bracketed hash per name and resolve them in one batched query against every domain the subgraph has a label for, not just labels that were independently registered as their own `.eth` 2LD name. A labelhash is a function of the label alone, so it decodes the same wherever in the name it sits, including a subdomain-only or DNS-imported label that never had its own registration. A hash the subgraph has never seen is left as it is, as before.

Fixes #577

### DIFF
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -17,11 +17,14 @@ type ResolvedLabel = {
   labelName: string;
 };
 
-const HASHED_LABEL = /\[(.*?)\]/g;
+const HASHED_LABEL = /\[([0-9a-f]{64})\]/g;
+// A labelhash isn't unique across domains, so the row count isn't bounded by
+// the number of hashes queried; 1000 is what every indexer in the subgraph's
+// pool accepts, not a value derived from the query itself.
 const DOMAINS_PAGE_SIZE = 1000;
 
 function getLabelHashes(domain: Domain) {
-  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash).filter(Boolean);
+  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash);
 }
 
 async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Handle[]> {
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -12,15 +12,16 @@ type Domain = {
   expiryDate?: number;
 };
 
-type Registration = {
-  id: string;
-  domain?: Pick<Domain, 'labelName'>;
+type ResolvedLabel = {
+  labelhash: string;
+  labelName: string;
 };
 
 const HASHED_LABEL = /\[(.*?)\]/g;
+const DOMAINS_PAGE_SIZE = 1000;
 
 function getLabelHashes(domain: Domain) {
-  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash);
+  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash).filter(Boolean);
 }
 
 async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Handle[]> {
@@ -28,23 +29,21 @@ async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Han
 
   if (!hashes.length) return domains.map(domain => domain.name);
 
-  const { data } = await graphQlCall<{ registrations: Registration[] }>(
+  const { data } = await graphQlCall<{ domains: ResolvedLabel[] }>(
     constants.ensSubgraph[chainId],
-    `query Registrations($ids: [String!]!, $first: Int!) {
-      registrations(first: $first, where: { id_in: $ids }) {
-        id
-        domain {
-          labelName
-        }
+    `query Labels($hashes: [Bytes!]!, $first: Int!) {
+      domains(first: $first, where: { labelhash_in: $hashes, labelName_not: null }) {
+        labelhash
+        labelName
       }
     }`,
     {
-      ids: hashes.map(hash => `0x${hash}`),
-      first: hashes.length
+      hashes: hashes.map(hash => `0x${hash}`),
+      first: DOMAINS_PAGE_SIZE
     }
   );
   const labelNames = new Map(
-    data.registrations.map(registration => [registration.id, registration.domain?.labelName])
+    data.domains.map(({ labelhash, labelName }) => [labelhash, labelName])
   );
 
   return domains.map(domain =>
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -17,14 +17,14 @@ type Registration = {
   domain?: Pick<Domain, 'labelName'>;
 };
 
-function getLabelHash(domain: Domain) {
-  return domain.name.match(/\[(.*?)\]/)?.[1];
+const HASHED_LABEL = /\[(.*?)\]/g;
+
+function getLabelHashes(domain: Domain) {
+  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash);
 }
 
 async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Handle[]> {
-  const hashes = [
-    ...new Set(domains.map(getLabelHash).filter((hash): hash is string => Boolean(hash)))
-  ];
+  const hashes = [...new Set(domains.flatMap(getLabelHashes))];
 
   if (!hashes.length) return domains.map(domain => domain.name);
 
@@ -47,12 +47,9 @@ async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Han
     data.registrations.map(registration => [registration.id, registration.domain?.labelName])
   );
 
-  return domains.map(domain => {
-    const hash = getLabelHash(domain);
-    const labelName = hash ? labelNames.get(`0x${hash}`) : undefined;
-
-    return labelName ? domain.name.replace(`[${hash}]`, () => labelName) : domain.name;
-  });
+  return domains.map(domain =>
+    domain.name.replace(HASHED_LABEL, (label, hash) => labelNames.get(`0x${hash}`) || label)
+  );
 }
 
 export default async function lookupDomains(
--- a/test/unit/helpers/graphqlEnvelope.test.ts
+++ b/test/unit/helpers/graphqlEnvelope.test.ts
@@ -21,6 +21,7 @@ const ADDRESS = '0xeF8305E140ac520225DAf050e2f71d5fBcC543e7';
 const IMAGE_URL = 'https://example.com/avatar.png';
 
 const ENS_SUBGRAPH = '[subgrapher.snapshot.org]';
+const ENS_HASH = 'e691d0079f79eca4304219d11ee5e5ff3df65a87c2595597cc15b4efd1dfda2b';
 
 function respondWith(body: any, status = 200) {
   mockedFetch.mockResolvedValue(jsonResponse(body, status));
@@ -66,7 +67,7 @@ describe('graphQlCall callers surface an envelope failure', () => {
   describe('src/resolvers/lookupDomains/ens.ts - registration', () => {
     const hashedDomain = {
       data: {
-        account: { domains: [{ name: '[7f3a].eth' }], wrappedDomains: [] }
+        account: { domains: [{ name: `[${ENS_HASH}].eth` }], wrappedDomains: [] }
       }
     };
 
@@ -78,7 +79,7 @@ describe('graphQlCall callers surface an envelope failure', () => {
 
     it('decodes the label when the subgraph answers', async () => {
       answerWith({
-        data: { domains: [{ labelhash: '0x7f3a', labelName: 'vitalik' }] }
+        data: { domains: [{ labelhash: `0x${ENS_HASH}`, labelName: 'vitalik' }] }
       });
 
       await expect(ensLookupDomains(ADDRESS)).resolves.toEqual(['vitalik.eth']);
@@ -87,7 +88,7 @@ describe('graphQlCall callers surface an envelope failure', () => {
     it('keeps the hashed name when the label is genuinely unknown', async () => {
       answerWith({ data: { domains: [] } });
 
-      await expect(ensLookupDomains(ADDRESS)).resolves.toEqual(['[7f3a].eth']);
+      await expect(ensLookupDomains(ADDRESS)).resolves.toEqual([`[${ENS_HASH}].eth`]);
     });
 
     it('rejects rather than returning the undecoded name when the subgraph fails', async () => {
--- a/test/unit/helpers/graphqlEnvelope.test.ts
+++ b/test/unit/helpers/graphqlEnvelope.test.ts
@@ -78,14 +78,14 @@ describe('graphQlCall callers surface an envelope failure', () => {
 
     it('decodes the label when the subgraph answers', async () => {
       answerWith({
-        data: { registrations: [{ id: '0x7f3a', domain: { labelName: 'vitalik' } }] }
+        data: { domains: [{ labelhash: '0x7f3a', labelName: 'vitalik' }] }
       });
 
       await expect(ensLookupDomains(ADDRESS)).resolves.toEqual(['vitalik.eth']);
     });
 
     it('keeps the hashed name when the label is genuinely unknown', async () => {
-      answerWith({ data: { registrations: [] } });
+      answerWith({ data: { domains: [] } });
 
       await expect(ensLookupDomains(ADDRESS)).resolves.toEqual(['[7f3a].eth']);
     });
--- a/test/unit/resolvers/lookupDomains/ens.test.ts
+++ b/test/unit/resolvers/lookupDomains/ens.test.ts
@@ -7,6 +7,9 @@ jest.mock('../../../../src/helpers/graphql', () => ({
 
 const mockedGraphQlCall = graphQlCall as jest.Mock;
 const ADDRESS = '0xeF8305E140ac520225DAf050e2f71d5fBcC543e7';
+const HASH_A = '9834876dcfb05cb167a5c24953eba58c4ac89b1adf57f28f2f9d09af107ee8f0';
+const HASH_B = '3e744b9dc39389baf0c5a0660589b8402f3dbb49b89b3e75f2c9355852a3c677';
+const HASH_C = '64daa44ad493ff28a96effab6e77f1732a3d97d83241581b37dbd70a7a4900fe';
 
 function graphQlResponse<T>(data: T) {
   return { data };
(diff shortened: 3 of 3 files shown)


src/resolvers/lookupDomains/ens.ts
```diff
@@ -17,14 +17,14 @@ type Registration = {
...
+
+function getLabelHashes(domain: Domain) {
+  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash);
 }
 
 async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Handle[]> {
-  const hashes = [
-    ...new Set(domains.map(getLabelHash).filter((hash): hash is string => Boolean(hash)))
-  ];
+  const hashes = [...new Set(domains.flatMap(getLabelHashes))];
```
wa0x6e, line 31: **The batched query resolves only `.eth` 2LD registrations, so most subdomain-position hashes still can't decode.**

`registrations.id` is populated only when a label is independently registered as a `.eth` second-level name (`BaseRegistrar` mints one `Registration` per 2LD). A hashed label at any other tree position — a subdomain segment, a DNS-imported 2LD, or a label nobody registered on its own — has no `Registration` row, so it decodes only by coincidence: when the same label string also happens to be registered as its own `.eth` 2LD elsewhere.

Collecting every bracket is the right half of the fix, but the preimage dictionary is the wrong entity for it. Measured against the live mainnet subgraph (`constants.ensSubgraph["1"]`), sampling 100 real `name_contains: "].["`, `name_ends_with: ".eth"` domains → 127 distinct hashes:

| query | hashes resolved |
|---|---|
| `registrations(where: { id_in: … })` | 19 / 127 |
| `domains(where: { labelhash_in: …, labelName_not: null })` | 24 / 127 (strict superset) |

A concrete live name this PR leaves fully hashed:

```
[7710d5ebf94bcebcf1996bb7a3f5e24a6d24435b314b3cec815da03640c2940c].[2e8eaa68c7e128861299162323c29c29672f5c094aceaf22d9c0935e4bbd3f85].[a64d2b5a93eda272d27734cc2fb8d1c468562e279f1e97e759eea1a5a410f8e3].[462a1d6391f7ea5916874504f3b5fc8cd43626f6bbabc8a22fe4312dc1585362].enspunks.eth
```

`registrations` resolves 0 of its 4 labels; `domains(labelhash_in:)` resolves all 4 → `🥛.🐄.🛸.👽.enspunks.eth`. Every bracketed label there is a pure subdomain segment, so no `Registration` for them will ever exist. For any collection minting subdomains under one non-hashed apex, this PR's benefit is currently 0%.

Two caveats the entity swap raises, both resolvable:

- **`Domain.labelhash` is not unique** — one hash can back many rows (the label `"8"` alone returned 122 in the sample). It is still a pure function of the label string: across 161 rows sharing 24 hashes there were 0 conflicting `labelName` values, so which row the `Map` keeps doesn't affect correctness. Worth a comment so the next reader doesn't have to re-derive it.
- **`first` can no longer be `hashes.length`**, since the row count isn't bounded by the number of hashes. Some fixed page size has to be picked instead. Probing the live gateway, `first` values up to 3500 succeed reproducibly and 4000+ fail, but the ceiling is a property of whichever indexer in the decentralized pool serves the request — at least one enforces exactly 1000, and the pool composition can shift. 1000 is the value every indexer accepts, so it's the safe choice rather than a measured hard limit.

Proposed patch (`tsc --noEmit` clean, eslint clean, 6/6 unit tests pass after the fixture updates):

```diff
diff --git a/src/resolvers/lookupDomains/ens.ts b/src/resolvers/lookupDomains/ens.ts
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -12,12 +12,18 @@ type Domain = {
   expiryDate?: number;
 };
 
-type Registration = {
-  id: string;
-  domain?: Pick<Domain, 'labelName'>;
+type ResolvedLabel = {
+  labelhash: string;
+  labelName: string;
 };
 
 const HASHED_LABEL = /\[(.*?)\]/g;
+// Page size every subgraph indexer accepts; a labelhash is shared by every domain that
+// ever used that label, so it is not unique and the row count per hash can't be
+// bounded by the number of hashes we're looking up (a common label like "8"
+// alone returns 100+ domains). Requesting the max page in one round trip is the
+// best coverage a single batched query can give.
+const DOMAINS_PAGE_SIZE = 1000;
 
 function getLabelHashes(domain: Domain) {
   return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash);
@@ -30,23 +36,24 @@ async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Han
 
   const {
     data: { data }
-  } = await graphQlCall<{ registrations: Registration[] }>(
+  } = await graphQlCall<{ domains: ResolvedLabel[] }>(
     constants.ensSubgraph[chainId],
-    `query Registrations($ids: [String!]!, $first: Int!) {
-      registrations(first: $first, where: { id_in: $ids }) {
-        id
-        domain {
-          labelName
-        }
+    `query Labels($hashes: [Bytes!]!, $first: Int!) {
+      domains(first: $first, where: { labelhash_in: $hashes, labelName_not: null }) {
+        labelhash
+        labelName
       }
     }`,
     {
-      ids: hashes.map(hash => `0x${hash}`),
-      first: hashes.length
+      hashes: hashes.map(hash => `0x${hash}`),
+      first: DOMAINS_PAGE_SIZE
     }
   );
+  // A labelhash is a pure function of the label, so every row sharing a
+  // labelhash carries the same labelName regardless of where in the ENS tree
+  // it sits - which row "wins" here doesn't matter.
   const labelNames = new Map(
-    data.registrations.map(registration => [registration.id, registration.domain?.labelName])
+    data.domains.map(({ labelhash, labelName }) => [labelhash, labelName])
   );
 
   return domains.map(domain =>
```

The unit-test fixtures need the matching rename (`registrations`/`id`/`domain.labelName` → `domains`/`labelhash`/`labelName`, and the `first` assertions → `1000`). Worth adding one test where a hashed subdomain label resolves *without* a same-label `.eth` registration — that's the case the current fixtures can't distinguish, since they fabricate registration rows for arbitrary hashes.

src/resolvers/lookupDomains/ens.ts
```diff
@@ -17,14 +17,14 @@ type Registration = {
   domain?: Pick<Domain, 'labelName'>;
 };
 
-function getLabelHash(domain: Domain) {
-  return domain.name.match(/\[(.*?)\]/)?.[1];
+const HASHED_LABEL = /\[(.*?)\]/g;
+
+function getLabelHashes(domain: Domain) {
+  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash);
```
wa0x6e, line 27: The old code filtered the hash list with `.filter((hash): hash is string => Boolean(hash))`, which dropped both `undefined` (no match) *and* `''`. `flatMap(getLabelHashes)` only drops the no-match case, and `/\[(.*?)\]/g` matches `[]` with an empty capture — so a name whose only bracket is `[]` now yields `hashes = ['']`, the `if (!hashes.length) return …` short-circuit below no longer fires, and the code issues a subgraph round trip that the old code never made.

The subgraph tolerates `id_in: ["0x"]` today (empty result, no error) so the returned name is unchanged — but the extra call is a new failure surface: on a subgraph error or a null list the unguarded `.map` throws, `lookupDomains` rejects, and `lookupDomains/index.ts` swallows it into an empty ENS list for that address. Before, that name came back with no network call at all.

Reachability is 0 today (`domains(where: { name_contains: "[]" })` returns no rows, and ENSIP-15 forbids `[`/`]` in labels), so this is defense-in-depth rather than a live bug.

```suggestion
  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash).filter(Boolean);
```

Worth a test alongside it — this one fails without the change with `TypeError: Cannot destructure property 'data' of '(intermediate value)' as it is undefined`:

```ts
it('leaves an empty bracket untouched instead of querying the subgraph for it', async () => {
  mockedGraphQlCall.mockResolvedValueOnce(accountResponse([{ name: '[].eth' }]));

  await expect(lookupDomains(ADDRESS)).resolves.toEqual(['[].eth']);
  expect(mockedGraphQlCall).toHaveBeenCalledTimes(1);
});
```

One alternative considered and rejected: tightening `HASHED_LABEL` to `/\[([0-9a-f]{64})\]/g` would also stop the malformed ids this line can produce (`'[[aaa]].eth'` → `0x[aaa`, `'[AAA].eth'` → `0xAAA`, which never matches the lowercase ids the subgraph stores). But it breaks 5 of the 7 tests in this file, since the fixtures all use short placeholders like `[aaa]` rather than real 64-char hex — a much larger change for a case with no measured reachability. The one-line filter is the surgical version.

wa0x6e (changes requested):

src/resolvers/lookupDomains/ens.ts
tony8713: Reproduced against the live subgraph with your exact `enspunks.eth` example: `registrations(where: id_in)` returns 0/4 for those hashes, `domains(where: labelhash_in, labelName_not: null)` returns all 4. Also reproduced the pagination point directly: querying that entity with `first: hashes.length` (4) returns only 2 of the 4 target hashes because rows are not 1:1 with hashes (one hash alone returned 12 duplicate rows in that 4-hash query), while `first: 1000` gets all 4.

Applied the entity swap (renamed the type to `ResolvedLabel`, `DOMAINS_PAGE_SIZE = 1000`), updated the query and the response mapping, and updated the fixtures in `ens.test.ts` plus the ones in `test/unit/helpers/graphqlEnvelope.test.ts` that were still shaped for `registrations`. Added a duplicate-labelhash-rows test and a fixed-page-size test.

Left out the two comments from the suggested diff since a test now guards each: reverting the query back to `registrations(id_in)` reddens 6 tests (4 in ens.test.ts, 2 in graphqlEnvelope.test.ts), and reverting `first` to a `hashes.length`-derived value reddens the fixed-page-size test.

Pushed in f450e3d.

src/resolvers/lookupDomains/ens.ts
tony8713: Reproduced with a small script: `[].eth` matched by `/\[(.*?)\]/g` yields `hashes = ['']`, `hashes.length` is 1 so the early-return never fires, and the subgraph is called with an empty id. Confirmed live that the subgraph tolerates it today (`registrations(where: { id_in: ["0x"] })` -> `[]`, no error), matching your "reachability is 0, defense-in-depth" read, and also confirmed no live domain matches `name_contains: "[]"`.

Added the `.filter(Boolean)` and the test you proposed (`leaves an empty bracket untouched instead of querying the subgraph for it`), verified it reproduces the failure you described when the filter is reverted (`TypeError: Cannot read properties of undefined (reading 'map')` from the `domains`-list read on a null/short-circuited response, same shape as the error you quoted).

Pushed in f450e3d.

src/resolvers/lookupDomains/ens.ts
```diff
@@ -12,47 +12,43 @@ type Domain = {
...
-  id: string;
-  domain?: Pick<Domain, 'labelName'>;
+type ResolvedLabel = {
+  labelhash: string;
+  labelName: string;
 };
 
-function getLabelHash(domain: Domain) {
-  return domain.name.match(/\[(.*?)\]/)?.[1];
+const HASHED_LABEL = /\[(.*?)\]/g;
```
wa0x6e: **A bracket in an ENS label now empties the whole address's name list, where `master` returned the names.**

`HASHED_LABEL` captures *any* bracketed content, not just a 64-hex labelhash, and line 41 feeds `0x${capture}` into `$hashes: [Bytes!]!`. `master`'s `id_in: [String!]` accepted anything; `Bytes` validates. Measured against the live mainnet gateway at `ddd80c3`:

| capture source | value sent | `Bytes` (this PR) | `String` (master) |
|---|---|---|---|
| `[]` | filtered by `.filter(Boolean)` | never sent | never sent |
| `[<64 uppercase hex>]` | `0xAAA…` | 200, **resolves** (improvement) | 200, 0 rows |
| `[aaa]` | `0xaaa` | **400** ``Failed to decode `Bytes` value: `Odd number of digits` `` | 200, 0 rows |
| `a[b]c` | `0xb` | **400** `Odd number of digits` | 200, 0 rows |
| `[zz]` | `0xzz` | **400** `Invalid character 'z' at position 0` | 200, 0 rows |
| `[[aaa]]` | `0x[aaa` | **400** `Invalid character '[' at position 0` | 200, 0 rows |

One bad capture kills every valid hash in the same request — `{ hashes: ['0x7710d5eb…' (real), '0x[aaa'] }` returns 400 with zero rows, not a partial result.

Failure chain: 400 → `helpers/graphql.ts:47` `!response.ok` throws `httpError(host, 400, …)` → `fetchDomainNames` rejects → `lookupDomains/index.ts:44` catches → `isSilencedError` is false for a 400 (`helpers/errors.ts` lists neither the message nor the code) → `capture()` to Sentry, then `return []`. The address loses **every** ENS name, not just the malformed one, and each lookup emits a Sentry event.

Concrete scenario: an address owns `vitalik.eth` and a NameWrapper subdomain `a[b]c.vitalik.eth`. Today Stamp returns both. After this PR it returns `[]` for that address on every request.

Reachability today is 0 by measurement — `labelName_contains: "["` and `"]"` return 0 rows on both the mainnet and sepolia subgraphs (positive control: `labelName_contains: "-"` → 5 rows on each, so the filter hits). But it is user-creatable at will: ENS labels carry no on-chain charset restriction, and NameWrapper `setSubnodeRecord` puts the label string into the event the subgraph indexes, so one cheap tx under a name you already own is enough.

Fix at this line rather than at the call site — it is the single boundary both `getLabelHashes` and the `replace` below go through, so constraining it covers the lookup and the substitution at once:

```suggestion
// The subgraph substitutes an unknown label with its lowercase 32-byte labelhash.
// Anything else in brackets is a literal label character: not a labelhash to look
// up, and `labelhash_in: [Bytes!]` rejects the whole batch if we send it one.
const HASHED_LABEL = /\[([0-9a-f]{64})\]/g;
```

This subsumes the `[]` case `.filter(Boolean)` handles on line 24 (harmless to keep) and the `[[aaa]]` / `[AAA]` cases raised in #discussion_r3840514198. Cost is fixture churn: `0xaaa`/`0xbbb`/`0xccc` and the 1500-name loop in `ens.test.ts`, plus `[7f3a]` in `graphqlEnvelope.test.ts`, need real 64-hex — about ten literals. Worth one test with a name like `a[b]c.eth` asserting no malformed value reaches the query.

**Why this is the same class as the last round.** #discussion_r3840514198 already enumerated this matrix — its own text lists `''`, `[[aaa]]` → `0x[aaa`, and `[AAA]` → `0xAAA` — and closed only the `''` cell, on the stated reasoning that the rest were harmless *because the subgraph took a `String`*. Swapping the entity changed that parameter to `Bytes`, which invalidates the reasoning, and no cell was re-checked. The table above is that matrix re-measured: one cell got better, four got worse.

src/resolvers/lookupDomains/ens.ts
```diff
@@ -12,47 +12,43 @@ type Domain = {
...
+const DOMAINS_PAGE_SIZE = 1000;
```
wa0x6e, line 24: Nit: nothing records *why* the page size is fixed at 1000 rather than `hashes.length`. `requests a fixed page size regardless of how many hashes are looked up` pins the *what*, and `resolves a label from duplicate rows sharing one labelhash` does document the non-uniqueness — but the step from "a labelhash is not unique" to "so 1000, and not more" is the part a reader has to re-derive from the subgraph's indexer limits. One line here would carry it.

src/resolvers/lookupDomains/ens.ts
```diff
@@ -12,47 +12,43 @@ type Domain = {
...
+
+function getLabelHashes(domain: Domain) {
+  return [...domain.name.matchAll(HASHED_LABEL)].map(([, hash]) => hash).filter(Boolean);
```
wa0x6e: Verified: dropping `.filter(Boolean)` reddens `leaves an empty bracket untouched instead of querying the subgraph for it`.

Small correction to the reply on #discussion_r3840514198, so it does not get carried forward — the without-fix failure is quoted there as `TypeError: Cannot read properties of undefined (reading 'map')`. The actual message, at both the pre-rebase head and `ddd80c3`, is `Cannot destructure property 'data' of '(intermediate value)' as it is undefined`. Substance is unaffected; the test does fail without the filter.

wa0x6e (changes requested):

src/resolvers/lookupDomains/ens.ts
tony8713: Reproduced independently against the live mainnet gateway. A mixed batch of one real hash plus `0xaaa` returns 400 with zero rows (not a partial result), and each malformed shape you listed (`0xaaa`, `0xb`, `0xzz`, `0x[aaa`) 400s while the uppercase-64 case actually resolves (200, `[]`). Confirmed your reachability read too: `labelName_contains: "["` / `"]"` return 0 rows on both mainnet and sepolia, `"-"` returns 5 on each as a positive control.

Applied your suggested regex (`/\[([0-9a-f]{64})\]/g`). Since the capture group can no longer be empty, `.filter(Boolean)` on the next line became dead code, so I dropped it rather than keep an unreachable branch. Updated the fixtures (`aaa`/`bbb`/`ccc`/`7f3a` placeholders → real 64-hex) and added a test asserting `a[b]c.eth`, `[AAA].eth` and `[[aaa]].eth` all pass through untouched with exactly one graphql call (the account query, no domains lookup).

Mutation-proved: reverting the regex to `/\[(.*?)\]/g` reddens exactly that new test plus the pre-existing empty-bracket test, nothing else. Pushed in 976b00c.

src/resolvers/lookupDomains/ens.ts
tony8713: Added a comment above `DOMAINS_PAGE_SIZE` covering the "why 1000" gap: it is the value every indexer in the subgraph pool accepts, not a value derived from the query itself, so raising it without knowing that is unsafe. Pushed in 976b00c.

src/resolvers/lookupDomains/ens.ts
tony8713: Noted for the record, thanks. No code change needed here, this was a correction to the error text quoted in my earlier reply, and the substance (the test does fail without the filter) is unaffected.

wa0x6e (approved):
tAck

PR #602, opened 2026-08-23
test: reshape the fixtures that still describe axios

tony8713:
## What

Four fixtures in two files still spell their errors the way `axios` did, in files that PR does not otherwise touch.

`axios` is gone from `package.json` and from `yarn.lock`, and nothing in the tree produces `Request failed with status code N` anymore. A GraphQL non-2xx now reads `[host] status code N: Reason`, a plain fetch non-2xx reads `[host] Reason`, and a transient ethers 504 arrives as `bad response (status=504, ...)`. Each message is swapped for the one its own path now produces. The old `[trustwallet] Not Found` was wrong for a second reason: trustwallet fetches through `fetchHttpImage`, so the host in that message is never the resolver name.

The assertions are unchanged. The two 404 cases in the image failure contract are renamed for the slot each one pins, `error.response.status` and `error.status`, since that is the only thing telling them apart. Each carries one slot rather than both, the same one-slot-per-case shape `test/integration/helpers/errors.test.ts` uses. `httpError` sets both, so neither of those two is a literal copy of an error it produces.

Same class as the fixture comment on #457, which reshaped `test/unit/api.test.ts` and `test/integration/helpers/errors.test.ts` and left these two out of that diff.

### DIFF
--- a/test/unit/resolvers/address.test.ts
+++ b/test/unit/resolvers/address.test.ts
@@ -37,6 +37,9 @@ const RESOLVERS = [
 ];
 
 const ADDRESS = '0xE6D0Dd18C6C3a9Af8C2FaB57d6e6A38E29d513cC';
+const ETHERS_504 =
+  'bad response (status=504, headers={}, body="error code: 504", code=SERVER_ERROR, version=web/5.7.1)';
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

PR #610, opened 2026-08-24
fix(http): data: profile pictures are rewritten to the IPFS gateway and rejected

### LINKED ISSUE #609: fix(http): data: profile pictures are rewritten to the IPFS gateway and rejected (STAMP-7E, STAMP-6P)

tony8713:
## Problem

`getUrl` (`src/helpers/http.ts`) treats any URL that isn't `ipfs://`, `ipns://`, `http://`, or `https://` as a bare CID and rewrites it onto the IPFS gateway path. A Starknet `getStarkProfile().profilePicture` for an on-chain-metadata NFT (Blobert, Realms, ...) is a `data:` URI, not a CID, so it gets concatenated onto the gateway path (`.../ipfs/data:application/json;base64,...`) and the gateway answers 400. The resolver reports the failure and falls back to a blockie; the real avatar never renders for these identities.

## Fix

Short-circuit `getUrl` for `data:` URIs: they carry their own bytes and need no gateway.

Closes #609.

### DIFF
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
@@ -35,6 +35,8 @@ export function fetchHttpImage(url: string): Promise<Buffer> {
 }
 
 export function getUrl(url) {
+  if (url.startsWith('data:')) return url;
+
   const gateway: string = process.env.IPFS_GATEWAY || 'cloudflare-ipfs.com';
   return snapshot.utils.getUrl(url, gateway);
 }
--- a/test/unit/resolvers/image/starknet.test.ts
+++ b/test/unit/resolvers/image/starknet.test.ts
@@ -5,7 +5,7 @@ jest.mock('../../../../src/helpers/provider', () => ({
 }));
 
 import starknet from '../../../../src/resolvers/image/starknet';
-import { incompleteJsonResponse, mockGlobalFetch } from '../../../helpers/fetch';
+import { incompleteJsonResponse, jsonResponse, mockGlobalFetch } from '../../../helpers/fetch';
 
 const mockedFetch = mockGlobalFetch();
 
@@ -57,5 +57,20 @@ describe('Starknet image resolver', () => {
     await expect(starknet(ADDRESS)).rejects.toMatchObject({
       name: 'AbortError'
     });
+  });
+
+  it('fetches a data: URI profile picture directly rather than through the IPFS gateway', async () => {
+    const metadataUri =
+      'data:application/json;base64,eyJpbWFnZSI6ImRhdGE6aW1hZ2Uvc3ZnK3htbDtiYXNlNjQsUEhOMlp5In0=';
+    const imageUri = 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0i';
+    mockGetStarkProfile.mockResolvedValue({ profilePicture: metadataUri });
+    mockedFetch
+      .mockResolvedValueOnce(jsonResponse({ image: imageUri }))
+      .mockResolvedValueOnce(new Response('svg-bytes'));
+
+    await expect(starknet(ADDRESS)).resolves.toBeInstanceOf(Buffer);
+
+    expect(mockedFetch).toHaveBeenNthCalledWith(1, metadataUri, expect.anything());
+    expect(mockedFetch).toHaveBeenNthCalledWith(2, imageUri, expect.anything());
   });
 });


src/helpers/http.ts
```diff
@@ -35,6 +35,8 @@ export function fetchHttpImage(url: string): Promise<Buffer> {
 }
 
 export function getUrl(url) {
+  if (url.startsWith('data:')) return url;
```
wa0x6e, line 38: URI schemes are case-insensitive ([RFC 3986 §3.1](https://www.rfc-editor.org/rfc/rfc3986#section-3.1)), so `startsWith('data:')` misses `DATA:` and `Data:`. Node's `fetch` accepts both (measured: `fetch('DATA:text/plain,ok')` and `fetch('Data:text/plain,ok')` each return 200 `ok`), but this guard doesn't, so a non-lowercase value still falls through to `snapshot.utils.getUrl` and comes back as `https://cloudflare-ipfs.com/ipfs/DATA:...` — the exact 400 this PR removes.

```suggestion
  if (/^data:/i.test(url)) return url;
```

On prevalence, so this is weighted honestly: I sampled all 13 addresses behind STAMP-7E/STAMP-6P through `getStarkProfile` on `rpc.snapshot.org` — 13/13 are lowercase `data:application/json;base64,…`, so this does not affect the incident being fixed here (Starknet PFPs are contract-template-generated). It matters for the other call sites: `getUrl` is shared, and `basename.ts:102` reads an ENS-style `avatar` text record, which is raw human-typed contract storage with no normalisation, as are `selfid.ts:14` and `snapshot.ts:141`. One character for a shared helper that several unvalidated inputs pass through.

No test asked for here — a casing case belongs in a direct `getUrl` unit test rather than a second copy of the resolver test.

test/unit/resolvers/image/starknet.test.ts
```diff
@@ -58,4 +58,19 @@ describe('Starknet image resolver', () => {
...
+    mockGetStarkProfile.mockResolvedValue({ profilePicture: metadataUri });
+    mockedFetch
+      .mockResolvedValueOnce(jsonResponse({ image: imageUri }))
+      .mockResolvedValueOnce(new Response('svg-bytes'));
+
+    await expect(starknet(ADDRESS)).resolves.toBeInstanceOf(Buffer);
+
+    expect(mockedFetch).toHaveBeenNthCalledWith(1, metadataUri, expect.anything());
+    expect(mockedFetch).toHaveBeenNthCalledWith(2, imageUri, expect.anything());
+  });
```
wa0x6e, line 75: This test can't fail for the reason the PR exists. Two things it doesn't reach:

**1. The real `data:` decode.** `global.fetch` is mocked for the whole file, so the load-bearing premise here — that undici decodes `data:` URLs *and* synthesises the `Content-Type` that `fetchImageOrMetadata` branches on at `src/resolvers/image/starknet.ts:28` — has no coverage at all. I checked that an un-mocked test is possible in this exact Jest environment (Node 22.23.2, `testEnvironment: 'node'`): a real `fetch()` against a `data:` URI works and never touches the network.

**2. The render claim.** `new Response('svg-bytes')` isn't image bytes and `resolves.toBeInstanceOf(Buffer)` passes for any bytes at all, so nothing pins that these avatars now render; the resolver is imported directly, bypassing `withResize`, so sharp is never in the loop.

The check that convinced me: with the patch below applied, reverting this PR's own `getUrl` change makes the new integration tests fail with the literal production error, `[snapshot.4everland.link] Bad Request` (STAMP-7E/STAMP-6P) — while the current mocked test stays green through the same revert.

`test/integration/helpers/http.test.ts` is already wired into CI via `yarn test:coverage`, and `data:` URIs never leave the process, so this adds no network dependency:

```diff
--- a/test/integration/helpers/http.test.ts
+++ b/test/integration/helpers/http.test.ts
@@ -1,7 +1,7 @@
-import { fetchHttpImage } from '../../../src/helpers/http';
+import { fetchHttpImage, fetchWithDeadline, getUrl } from '../../../src/helpers/http';
@@ -66,4 +66,21 @@ describe('fetchHttpImage', () => {
+  it('decodes a data: URI through the real fetch implementation, no network involved', async () => {
+    const svg = '<svg xmlns="http://www.w3.org/2000/svg"></svg>';
+    const dataUri = `data:image/svg+xml;base64,${Buffer.from(svg).toString('base64')}`;
+
+    await expect(fetchHttpImage(getUrl(dataUri))).resolves.toEqual(Buffer.from(svg));
+  });
+
+  it('gets the Content-Type undici synthesizes for a data: URI, the same branch fetchImageOrMetadata relies on', async () => {
+    const dataUri = 'data:application/json;base64,eyJmb28iOiJiYXIifQ==';
+    const read = (response: Response) =>
+      response.headers.get('content-type')?.includes('application/json')
+        ? response.json()
+        : Promise.reject(new Error('expected an application/json content-type'));
+
+    await expect(fetchWithDeadline(getUrl(dataUri), read)).resolves.toEqual({ foo: 'bar' });
+  });
 });
```

And the assertion in this file, so it fails when the bytes are wrong rather than merely present:

```diff
--- a/test/unit/resolvers/image/starknet.test.ts
+++ b/test/unit/resolvers/image/starknet.test.ts
@@ -60,15 +60,16 @@ describe('Starknet image resolver', () => {
   it('fetches a data: URI profile picture directly rather than through the IPFS gateway', async () => {
+    const svg = '<svg xmlns="http://www.w3.org/2000/svg"></svg>';
     const metadataUri =
       'data:application/json;base64,eyJpbWFnZSI6ImRhdGE6aW1hZ2Uvc3ZnK3htbDtiYXNlNjQsUEhOMlp5In0=';
     const imageUri = 'data:image/svg+xml;base64,PHN2ZyB4bWxucz0i';
     mockGetStarkProfile.mockResolvedValue({ profilePicture: metadataUri });
     mockedFetch
       .mockResolvedValueOnce(jsonResponse({ image: imageUri }))
-      .mockResolvedValueOnce(new Response('svg-bytes'));
+      .mockResolvedValueOnce(new Response(svg));
 
-    await expect(starknet(ADDRESS)).resolves.toBeInstanceOf(Buffer);
+    await expect(starknet(ADDRESS)).resolves.toEqual(Buffer.from(svg));
```

wa0x6e (changes requested):

wa0x6e:
Superseded by #598 — closing.

Both files here (`src/helpers/http.ts`, `test/unit/resolvers/image/starknet.test.ts`) are already in #598's diff, and the two changes are entangled: #598 rewrites the Starknet metadata path that consumes `getUrl`'s output, so a guard it may add there has to know whether `data:` URIs survive `getUrl`. A guard requiring an absolute `http(s)`/`ipfs` URL would reject the exact value the new test here asserts.

Traced using a `data:application/json` profile picture whose `image` is a `data:image/svg+xml` URI (the shape of the live STAMP-7E input): #598 alone answers 400 from the gateway and reports it; #598 plus this change returns the image. Keeping them separate also guarantees a conflict in `starknet.test.ts` for whichever merges second — verified by rebase.

The change itself is correct and shouldn't be lost: the `data:` short-circuit and a port of the new test belong in #598, which should then close #609 as well. Reopen this if #598 ends up not carrying it.

PR #614, opened 2026-08-24
feat(lookupDomains): include names registered on ENSv2

ChaituVR:
### Summary

Part of snapshot-labs/workflow#823. ENS moved Sepolia to ENSv2: names registered there live in per-name registries that the v1 subgraph does not index. `lookup_domains` only queried the v1 subgraph, so a wallet owning only ENSv2 names got an empty list — in the space-creation flow this shows as "No more domain names available for space creation found" even though the wallet owns a usable name.

- The ENS provider now also queries the ENSv2 graphql endpoint (`constants.ensV2Graphql`, Sepolia only for now — the same API the official ENS app uses) with `domains(where: { owner })`, and merges the result with the v1 subgraph names: deduplicated, expired names filtered, same 1000 cap.
- A v2 lookup failure is captured and only drops the v2 names, never the v1 result — relevant because the endpoint carries a `staging-` prefix and may move while ENSv2 is in beta.
- Mainnet and every other chain are untouched: without an `ensV2Graphql` entry no extra request is made.

### How to test

```
yarn jest test/unit/resolvers/lookupDomains/ens.test.ts
```

14 tests (9 existing untouched, 5 new: merge, dedupe, expiry filter, v2-failure resilience, no-v2-chain).

Live, `0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6` owns `boorger.eth` (v1) and `testspace.eth` (ENSv2) on Sepolia:

| | before | after |
|---|---|---|
| `lookup_domains` on `11155111` | `["boorger.eth"]` | `["boorger.eth", "testspace.eth"]` |
| create-space domain list on testnet.snapshot.box | v2 name missing | v2 name listed |

No UI change needed — the create-space flow reads this endpoint.

### DIFF
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -119,5 +119,5 @@ export default async function lookupDomains(
     return (!expiry || expiry > now) && !domain.name.endsWith('.addr.reverse');
   });
 
-  return [...new Set(await fetchDomainNames(domains, chainId))];
+  return fetchDomainNames(domains, chainId);
 }
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -1,6 +1,4 @@
-import { capture } from '@snapshot-labs/snapshot-sentry';
 import constants from '../../constants.json';
-import { isSilencedError } from '../../helpers/errors';
 import { graphQlCall } from '../../helpers/graphql';
 import { Address, Handle } from '../../helpers/types';
 
@@ -55,37 +53,6 @@ async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Han
   );
 }
 
-// names registered on ENSv2, whose ownership does not exist in the v1
-// subgraph. A failure only drops the v2 names, never the v1 result
-async function fetchV2Domains(address: Address, chainId: string): Promise<Domain[]> {
-  const endpoint = constants.ensV2Graphql[chainId];
-
-  if (!endpoint) return [];
-
-  try {
-    const { data } = await graphQlCall<{ domains: Domain[] }>(
-      endpoint,
-      `query Domains($where: DomainFilter!, $first: Int) {
-        domains(where: $where, first: $first) {
-          name
-          expiryDate
-        }
-      }`,
-      {
-        where: { owner: address.toLowerCase() },
-        first: DOMAINS_PAGE_SIZE
-      }
-    );
-
-    return data?.domains || [];
-  } catch (err) {
-    if (!isSilencedError(err)) {
-      capture(err, { contexts: { input: { address, chainId } } });
-    }
-    return [];
-  }
-}
-
 export default async function lookupDomains(
   address: Address,
   chainId = DEFAULT_CHAIN_ID
@@ -114,8 +81,7 @@ export default async function lookupDomains(
   const now = Date.now() / 1000;
   const domains: Domain[] = [
     ...(account?.domains || []),
-    ...(account?.wrappedDomains || []),
-    ...(await fetchV2Domains(address, chainId))
+    ...(account?.wrappedDomains || [])
   ].filter(domain => {
     const expiry = Number(domain.expiryDate ?? 0);
 
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -55,8 +55,6 @@ async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Han
   );
 }
 
-// names registered on ENSv2, whose ownership does not exist in the v1
-// subgraph. A failure only drops the v2 names, never the v1 result
 async function fetchV2Domains(address: Address, chainId: string): Promise<Domain[]> {
   const endpoint = constants.ensV2Graphql[chainId];
 
@@ -77,7 +75,7 @@ async function fetchV2Domains(address: Address, chainId: string): Promise<Domain
       }
     );
 
-    return data?.domains || [];
+    return data.domains || [];
   } catch (err) {
     if (!isSilencedError(err)) {
       capture(err, { contexts: { input: { address, chainId } } });
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -9,8 +9,7 @@ export const CHAIN_IDS = Object.keys(constants.ensSubgraph);
 
 type Domain = {
   name: string;
-  labelName?: string;
-  expiryDate?: number;
+  expiryDate?: string;
 };
 
 type ResolvedLabel = {
@@ -109,16 +108,16 @@ export default async function lookupDomains(
     { id: address.toLowerCase() }
   );
 
-  const now = (Date.now() / 1000).toFixed(0);
+  const now = Date.now() / 1000;
   const domains: Domain[] = [
     ...(account?.domains || []),
     ...(account?.wrappedDomains || []),
     ...(await fetchV2Domains(address, chainId))
-  ].filter(
-    domain =>
-      (!domain.expiryDate || domain.expiryDate === '0' || domain.expiryDate > now) &&
-      !domain.name.endsWith('.addr.reverse')
-  );
+  ].filter(domain => {
+    const expiry = Number(domain.expiryDate ?? 0);
+
+    return (!expiry || expiry > now) && !domain.name.endsWith('.addr.reverse');
+  });
 
   return [...new Set(await fetchDomainNames(domains, chainId))];
 }
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -1,3 +1,4 @@
+import { capture } from '@snapshot-labs/snapshot-sentry';
 import constants from '../../constants.json';
 import { graphQlCall } from '../../helpers/graphql';
 import { Address, Handle } from '../../helpers/types';
@@ -54,6 +55,35 @@ async function fetchDomainNames(domains: Domain[], chainId: string): Promise<Han
   );
 }
 
+// names registered on ENSv2, whose ownership does not exist in the v1
+// subgraph. A failure only drops the v2 names, never the v1 result
+async function fetchV2Domains(address: Address, chainId: string): Promise<Domain[]> {
+  const endpoint = constants.ensV2Graphql[chainId];
+
+  if (!endpoint) return [];
+
+  try {
+    const { data } = await graphQlCall<{ domains: Domain[] }>(
+      endpoint,
+      `query Domains($where: DomainFilter!, $first: Int) {
+        domains(where: $where, first: $first) {
+          name
+          expiryDate
+        }
+      }`,
+      {
+        where: { owner: address.toLowerCase() },
+        first: DOMAINS_PAGE_SIZE
+      }
+    );
+
+    return data?.domains || [];
+  } catch (err) {
+    capture(err, { contexts: { input: { address, chainId } } });
+    return [];
+  }
+}
+
 export default async function lookupDomains(
   address: Address,
   chainId = DEFAULT_CHAIN_ID
@@ -82,12 +112,13 @@ export default async function lookupDomains(
   const now = (Date.now() / 1000).toFixed(0);
   const domains: Domain[] = [
     ...(account?.domains || []),
-    ...(account?.wrappedDomains || [])
+    ...(account?.wrappedDomains || []),
+    ...(await fetchV2Domains(address, chainId))
   ].filter(
     domain =>
       (!domain.expiryDate || domain.expiryDate === '0' || domain.expiryDate > now) &&
       !domain.name.endsWith('.addr.reverse')
   );
 
-  return fetchDomainNames(domains, chainId);
+  return [...new Set(await fetchDomainNames(domains, chainId))];
 }
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -1,5 +1,6 @@
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import constants from '../../constants.json';
+import { isSilencedError } from '../../helpers/errors';
 import { graphQlCall } from '../../helpers/graphql';
 import { Address, Handle } from '../../helpers/types';
 
@@ -78,7 +79,9 @@ async function fetchV2Domains(address: Address, chainId: string): Promise<Domain
 
     return data?.domains || [];
   } catch (err) {
-    capture(err, { contexts: { input: { address, chainId } } });
+    if (!isSilencedError(err)) {
+      capture(err, { contexts: { input: { address, chainId } } });
+    }
     return [];
   }
 }
--- a/test/unit/resolvers/lookupDomains/ens.test.ts
+++ b/test/unit/resolvers/lookupDomains/ens.test.ts
@@ -15,13 +15,35 @@ function graphQlResponse<T>(data: T) {
   return { data };
 }
 
-function accountResponse(domains: { name: string }[]) {
+function accountResponse(domains: { name: string; expiryDate?: string }[]) {
   return graphQlResponse({ account: { domains, wrappedDomains: [] } });
 }
 
 describe('lookupDomains/ens', () => {
   beforeEach(() => {
     mockedGraphQlCall.mockReset();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('compares expiry dates numerically and preserves names without an expiry', async () => {
+    jest.spyOn(Date, 'now').mockReturnValue(1_787_572_496_000);
+    mockedGraphQlCall.mockResolvedValueOnce(
+      accountResponse([
+        { name: 'expired.eth', expiryDate: '1787572495' },
+        { name: 'permanent.eth', expiryDate: '0' },
+        { name: 'subdomain.eth' },
+        { name: 'far-future.eth', expiryDate: '13293458883' }
+      ])
+    );
+
+    await expect(lookupDomains(ADDRESS)).resolves.toEqual([
+      'permanent.eth',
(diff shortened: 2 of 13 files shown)


tony8713 (approved):
Approve. Core mechanism is sound: ENSv2 lookup is additive, chain-gated, and its failure path is isolated from the v1 result. Adversarial pass mutated the try/catch and confirmed a v2 failure never drops v1 results or throws. Two non-blocking notes, neither a defect in what's shipped: the cross-provider dedup in lookupDomains/index.ts has no dedicated test (mutating it to a plain concat left all related tests green, worth a follow-up test), and provider routing gates on ens.CHAIN_IDS rather than ensV2Graphql, so a hypothetical future v2-only chain would never invoke the v1 provider (not triggered by any chain today).

src/resolvers/lookupDomains/ens.ts
```diff
@@ -81,7 +112,8 @@ export default async function lookupDomains(
   const now = Date.now() / 1000;
   const domains: Domain[] = [
     ...(account?.domains || []),
-    ...(account?.wrappedDomains || [])
+    ...(account?.wrappedDomains || []),
+    ...(await fetchV2Domains(address, chainId))
```
wa0x6e: **Suggestion: make ENSv2 its own provider rather than a helper inside the ENS v1 provider.**

`fetchV2Domains` currently lives in `ens.ts` and its result is spread into the same array as the v1 subgraph names. Moving it to `src/resolvers/lookupDomains/ensV2.ts` and registering it in `PROVIDERS` — where every other domain source already lives — resolves several separate issues at once rather than patching each:

- **v1 and v2 stop failing together.** Today `fetchV2Domains` is awaited *after* the v1 `graphQlCall` resolves, so a v1 subgraph 5xx or the 5s `withDeadline` abort means the v2 request is never issued at all. A Sepolia wallet whose only name lives in ENSv2 then gets an empty list — the exact symptom this PR exists to fix — while the v2 endpoint is healthy and holding the name. As separate providers, `index.ts` already runs them concurrently under `Promise.all` with an independent `try/catch` each, so this is deletion rather than added error handling. Worst-case latency also becomes `max(v1, v2)` instead of `v1 + v2`.
- **`CHAIN_IDS` stops being keyed off the wrong config map.** It currently derives from `ensSubgraph`, so a chain added to `ensV2Graphql` alone would be silently dropped by the dispatcher's chain filter and never queried. The new provider derives its own from `ensV2Graphql`.
- **The metric stops lying.** The local `try/catch` returns `[]`, so the provider resolves normally and `lookup_domains_response_duration_seconds` records `status: 1` during a v2 outage — indistinguishable from "this chain has no v2 names". Removing the catch lets the dispatcher record `status: 0` and `capture` the error tagged `provider: 'Ens V2'` (the current capture also omits that tag, which every other lookupDomains failure carries).
- **The page size stops being shared.** `DOMAINS_PAGE_SIZE`'s comment justifies 1000 as what the v1 indexer pool accepts; that says nothing about the v2 API, which enforces its own cap. Measured: `first: 1000` succeeds, `first: 1001` returns `INVALID_ARGUMENT`. The shared literal sits at the v2 ceiling with zero headroom, so raising it for a v1 reason silently breaks v2.
- **The conflict with #597 disappears.** `gh pr view 597 --json files` returns exactly `src/resolvers/lookupDomains/ens.ts` and `test/unit/resolvers/lookupDomains/ens.test.ts` — the two files this restructure removes from this PR's diff. A real `git merge --no-commit` against #597's current head currently produces content conflicts in both. Afterwards the two PRs share no file. (#597 is also based on a commit predating 786c16c, so it still does the string expiry comparison — worth knowing whichever way you go.)

`DEFAULT_CHAIN_ID` is `'1'` deliberately, not `'11155111'`: `index.ts` builds its default chain set from every provider's `DEFAULT_CHAIN_ID`, so Sepolia there would make every caller that passes no `chains` query the v2 endpoint *and* drag the v1 Sepolia subgraph in with it. `'1'` keeps the default set at `['1','109','146']` and the provider starts working by itself the day an `ensV2Graphql` mainnet entry lands. This does mean `DEFAULT_CHAIN_ID` is not a member of `CHAIN_IDS`, which breaks a sibling-provider invariant — nothing enforces or depends on it, but it is worth a comment so it does not read as a mistake.

A few things dropped on measurement rather than carried over: no `.addr.reverse` filter (`name_ends_with: ".addr.reverse"` returns zero rows against the live index, positive control passes), no hashed-label resolution step (`name_contains: "["` with `includeUnnormalized`/`includeUnreachable` returns zero rows, control passes — this also removes the odd coupling where a v2 name was sent to the *v1* subgraph to resolve), and no `data.domains || []` (`Query.domains` is `[Domain!]!` and `graphQlCall` already throws on a bad envelope). `expiryDate` is typed `number` because the endpoint returns `Int`, not the `string` the v1 subgraph returns for its `BigInt`.

One non-obvious requirement: `test/unit/resolvers/lookupDomains.test.ts` mocks providers by module path, so a 4th provider without a mock means the **real** module runs inside the unit suite. Verified with a `fetch` spy — it hits `staging-graphql.ens.dev` from a unit test, and the `captures one event per failing chain` case asserts an exact call count, so an upstream outage would turn into a red unit test.

Verified with the patch applied: `tsc --noEmit` and eslint clean, `jest test/unit` 230/230, `jest test/integration/resolvers/lookupDomains.test.ts` 13/13 against live endpoints, and an end-to-end call through the dispatcher for `0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6` on `11155111` returns `["boorger.eth","testspace.eth"]` — v1 and v2 names both present.

`ens.ts` and `test/unit/resolvers/lookupDomains/ens.test.ts` simply revert to `master` (the `capture`/`isSilencedError` imports, `fetchV2Domains`, the spread, and the four ENSv2 tests all come out), so those hunks are omitted below. `src/constants.json` is unchanged — the `ensV2Graphql` entry stays exactly as this PR added it.

```diff
diff --git a/src/resolvers/lookupDomains/ensV2.ts b/src/resolvers/lookupDomains/ensV2.ts
new file mode 100644
--- /dev/null
+++ b/src/resolvers/lookupDomains/ensV2.ts
@@ -0,0 +1,47 @@
+import constants from '../../constants.json';
+import { graphQlCall } from '../../helpers/graphql';
+import { Address, Handle } from '../../helpers/types';
+
+export const NAME = 'Ens V2';
+// ENSv2 is L1-bound now that Namechain is cancelled. '1' keeps the dispatcher's
+// default chain set unchanged until an ensV2Graphql mainnet entry exists.
+export const DEFAULT_CHAIN_ID = '1';
+export const CHAIN_IDS = Object.keys(constants.ensV2Graphql);
+
+type Domain = {
+  name: string;
+  expiryDate?: number;
+};
+
+// Measured: first: 1000 succeeds, first: 1001 is rejected by the endpoint.
+const PAGE_SIZE = 1000;
+
+export default async function lookupDomains(
+  address: Address,
+  chainId = DEFAULT_CHAIN_ID
+): Promise<Handle[]> {
+  const endpoint = constants.ensV2Graphql[chainId];
+
+  if (!endpoint) return [];
+
+  const { data } = await graphQlCall<{ domains: Domain[] }>(
+    endpoint,
+    `query Domains($where: DomainFilter!, $first: Int!) {
+      domains(where: $where, first: $first) {
+        name
+        expiryDate
+      }
+    }`,
+    { where: { owner: address.toLowerCase() }, first: PAGE_SIZE }
+  );
+
+  const now = Date.now() / 1000;
+
+  return data.domains
+    .filter(domain => {
+      const expiry = Number(domain.expiryDate ?? 0);
+
+      return !expiry || expiry > now;
+    })
+    .map(domain => domain.name);
+}
diff --git a/src/resolvers/lookupDomains/index.ts b/src/resolvers/lookupDomains/index.ts
--- a/src/resolvers/lookupDomains/index.ts
+++ b/src/resolvers/lookupDomains/index.ts
@@ -1,6 +1,7 @@
 import { isAddress } from '@ethersproject/address';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import * as ens from './ens';
+import * as ensV2 from './ensV2';
 import * as shibarium from './shibarium';
 import * as unstoppableDomains from './unstoppableDomains';
 import { isSilencedError } from '../../helpers/errors';
@@ -15,7 +16,7 @@ type Provider = {
 };
 
 // Without the annotation a provider missing NAME still compiles.
-const PROVIDERS: Provider[] = [ens, shibarium, unstoppableDomains];
+const PROVIDERS: Provider[] = [ens, ensV2, shibarium, unstoppableDomains];
 
 const DEFAULT_CHAIN_IDS = PROVIDERS.map(provider => provider.DEFAULT_CHAIN_ID);
 
diff --git a/test/unit/resolvers/lookupDomains.test.ts b/test/unit/resolvers/lookupDomains.test.ts
--- a/test/unit/resolvers/lookupDomains.test.ts
+++ b/test/unit/resolvers/lookupDomains.test.ts
@@ -1,6 +1,7 @@
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import lookupDomains from '../../../src/resolvers/lookupDomains';
 import ens from '../../../src/resolvers/lookupDomains/ens';
+import ensV2 from '../../../src/resolvers/lookupDomains/ensV2';
 import shibarium from '../../../src/resolvers/lookupDomains/shibarium';
 import unstoppableDomains from '../../../src/resolvers/lookupDomains/unstoppableDomains';
 
@@ -15,6 +16,13 @@ jest.mock('../../../src/resolvers/lookupDomains/ens', () => ({
   CHAIN_IDS: ['1', '11155111'],
   default: jest.fn()
 }));
+jest.mock('../../../src/resolvers/lookupDomains/ensV2', () => ({
+  __esModule: true,
+  NAME: 'Ens V2',
+  DEFAULT_CHAIN_ID: '1',
+  CHAIN_IDS: ['11155111'],
+  default: jest.fn()
+}));
 jest.mock('../../../src/resolvers/lookupDomains/shibarium', () => ({
   __esModule: true,
   NAME: 'Shibarium',
@@ -37,6 +45,7 @@ const ENS_CHAINS = ['1', '11155111'];
 describe('lookupDomains - default chains', () => {
   it('calls every provider on its own default chain when no chain is given', async () => {
     (ens as jest.Mock).mockResolvedValue([]);
+    (ensV2 as jest.Mock).mockResolvedValue([]);
     (shibarium as jest.Mock).mockResolvedValue([]);
     (unstoppableDomains as jest.Mock).mockResolvedValue([]);
 
@@ -46,11 +55,24 @@ describe('lookupDomains - default chains', () => {
     expect(shibarium).toHaveBeenCalledWith(VALID_ADDRESS, '109');
     expect(unstoppableDomains).toHaveBeenCalledWith(VALID_ADDRESS, '146');
   });
+
+  it('does not widen the default chain set to a chain no provider defaults to', async () => {
+    (ens as jest.Mock).mockResolvedValue([]);
+    (ensV2 as jest.Mock).mockResolvedValue([]);
+    (shibarium as jest.Mock).mockResolvedValue([]);
+    (unstoppableDomains as jest.Mock).mockResolvedValue([]);
+
+    await lookupDomains(VALID_ADDRESS);
+
+    expect(ensV2).not.toHaveBeenCalled();
+    expect(ens).toHaveBeenCalledTimes(1);
+  });
 });
 
 describe('lookupDomains - chain routing', () => {
   beforeEach(() => {
     (ens as jest.Mock).mockResolvedValue([]);
+    (ensV2 as jest.Mock).mockResolvedValue([]);
     (shibarium as jest.Mock).mockResolvedValue([]);
     (unstoppableDomains as jest.Mock).mockResolvedValue([]);
   });
@@ -95,6 +117,7 @@ describe('lookupDomains - resolver failures', () => {
 
 describe('lookupDomains - error reporting', () => {
   beforeEach(() => {
+    (ensV2 as jest.Mock).mockResolvedValue([]);
     (shibarium as jest.Mock).mockResolvedValue([]);
     (unstoppableDomains as jest.Mock).mockResolvedValue([]);
   });
diff --git a/test/unit/resolvers/lookupDomains/ensV2.test.ts b/test/unit/resolvers/lookupDomains/ensV2.test.ts
new file mode 100644
--- /dev/null
+++ b/test/unit/resolvers/lookupDomains/ensV2.test.ts
@@ -0,0 +1,76 @@
+import { graphQlCall } from '../../../../src/helpers/graphql';
+import lookupDomains from '../../../../src/resolvers/lookupDomains/ensV2';
+
+jest.mock('../../../../src/helpers/graphql', () => ({
+  graphQlCall: jest.fn()
+}));
+
+const mockedGraphQlCall = graphQlCall as jest.Mock;
+const ADDRESS = '0xeF8305E140ac520225DAf050e2f71d5fBcC543e7';
+const CHAIN_ID = '11155111';
+const ENDPOINT = 'https://staging-graphql.ens.dev/';
+
+function domainsResponse(domains: { name: string; expiryDate?: number }[]) {
+  return { data: { domains } };
+}
+
+describe('lookupDomains/ensV2', () => {
+  beforeEach(() => {
+    mockedGraphQlCall.mockReset();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('returns the names owned by the address on the configured endpoint', async () => {
+    mockedGraphQlCall.mockResolvedValueOnce(
+      domainsResponse([{ name: 'testspace.eth' }, { name: 'boorger.eth' }])
+    );
+
+    await expect(lookupDomains(ADDRESS, CHAIN_ID)).resolves.toEqual([
+      'testspace.eth',
+      'boorger.eth'
+    ]);
+    expect(mockedGraphQlCall).toHaveBeenCalledTimes(1);
+    expect(mockedGraphQlCall).toHaveBeenCalledWith(
+      ENDPOINT,
+      expect.stringContaining('domains(where: $where, first: $first)'),
+      { where: { owner: ADDRESS.toLowerCase() }, first: 1000 }
+    );
+  });
+
+  it('filters expired names and keeps the ones without an expiry', async () => {
+    jest.spyOn(Date, 'now').mockReturnValue(1_787_572_496_000);
+    mockedGraphQlCall.mockResolvedValueOnce(
+      domainsResponse([
+        { name: 'expired.eth', expiryDate: 1_787_572_495 },
+        { name: 'permanent.eth', expiryDate: 0 },
+        { name: 'subdomain.eth' },
+        { name: 'far-future.eth', expiryDate: 13_293_458_883 }
+      ])
+    );
+
+    await expect(lookupDomains(ADDRESS, CHAIN_ID)).resolves.toEqual([
+      'permanent.eth',
+      'subdomain.eth',
+      'far-future.eth'
+    ]);
+  });
+
+  it('returns an empty array without a request on a chain without an endpoint', async () => {
+    await expect(lookupDomains(ADDRESS, '137')).resolves.toEqual([]);
+    expect(mockedGraphQlCall).not.toHaveBeenCalled();
+  });
+
+  it('does not query anything on the default chain', async () => {
+    await expect(lookupDomains(ADDRESS)).resolves.toEqual([]);
+    expect(mockedGraphQlCall).not.toHaveBeenCalled();
+  });
+
+  it('rejects instead of swallowing an upstream failure', async () => {
+    mockedGraphQlCall.mockRejectedValueOnce(new Error('endpoint moved'));
+
+    await expect(lookupDomains(ADDRESS, CHAIN_ID)).rejects.toThrow('endpoint moved');
+  });
+});
```

test/unit/resolvers/lookupDomains/ens.test.ts
```diff
@@ -175,4 +175,47 @@ describe('lookupDomains/ens', () => {
 
     await expect(lookupDomains(ADDRESS)).rejects.toThrow();
   });
+
+  describe('on a chain with ENSv2', () => {
```
wa0x6e: **Worth adding: one integration assertion for the ENSv2 leg.** (The test itself belongs in `test/integration/resolvers/lookupDomains.test.ts`, which has no hunk in this diff — commenting here since it is the nearest test file.)

The new unit tests cover the *isolation* of a v2 failure — that a v2 error doesn't take the v1 names down with it. What nothing covers is *detection*: if the endpoint moves or changes shape, the v2 names quietly stop appearing and every test still passes. That silent-empty-list outcome is the symptom this PR exists to fix, and the PR description names the endpoint moving as a live risk while it is in beta.

The repo already has the mechanism, and it genuinely runs — `.github/workflows/test.yml` runs `yarn test:coverage` → `jest --coverage` on every push, and `jest.config.ts` roots at `test/` with no `test/integration` exclusion and no env gate. Two things that look like existing coverage are not: the current sepolia assertion in `lookupDomains.test.ts` uses `0x24F15402C6Bb870554489b2fd2049A85d75B982f`, which owns zero ENSv2 names (the v2 endpoint returns `{"domains":[]}` for it), and the `shibarium`/`unstoppableDomains` files under `test/integration/` are fully fetch-mocked despite living in that directory.

```diff
diff --git a/test/integration/resolvers/lookupDomains.test.ts b/test/integration/resolvers/lookupDomains.test.ts
@@ -21,6 +21,12 @@ describe('lookupDomains', () => {
     expect(result).toContain('chaitu.eth');
   });
 
+  it('should include ENSv2 domains on sepolia', async () => {
+    const result = await lookupDomains('0x220bc93D88C0aF11f1159eA89a885d5ADd3A7Cf6', '11155111');
+
+    expect(result).toContain('testspace.eth');
+  });
+
   it('should return an empty array if the address is not provided', async () => {
     const result = await lookupDomains('');
```

Ran it against the live endpoint: passes in 796ms, 14/14 in that file. It drives the top-level `lookupDomains`, so it works unchanged whether or not the provider restructure lands.

Two costs, both worth stating rather than glossing: it adds one live call to a third-party beta service per push, and it hardcodes a fixture with a finite life (`testspace.eth` expires 2027-08-24). Both are the same tradeoff this file already makes for the v1 subgraph, D3 and Unstoppable Domains — `boorger.eth` / `boorger.shib` / `boorger.sonic` are hardcoded on that same owner wallet — so it is one more instance of an accepted convention, not a new dependency class. If you would rather not pin a specific name, asserting the result is non-empty for that address would still catch a total endpoint outage.

wa0x6e (changes requested):

wa0x6e (approved):
utAck

PR #619, opened 2026-08-24
fix(lookupDomains): drop a nameless ENS domain row instead of crashing

### LINKED ISSUE #618: fix(lookupDomains): a null domain name crashes the ENS expiry filter, dropping every name for that address
Found during review of #614. **Not from Sentry** — no occurrences observed; this is a reachable gap, not an active incident. Searched the `stamp` project over 90 days for `endsWith` and got nothing back. That negative is trustworthy: the same search for `Cannot read properties of null` returns STAMP-6W, so the lookup works rather than silently failing. The likely reason it has never fired is that the addresses which own a nameless row are not ones anybody looks up in practice.

## Today

`src/resolvers/lookupDomains/ens.ts:88` on `master` dereferences `domain.name` with no guard:

```ts
  ].filter(domain => {
    const expiry = Number(domain.expiryDate ?? 0);

    return (!expiry || expiry > now) && !domain.name.endsWith('.addr.reverse');
  });
```

The ENS subgraph can return a domain row whose `name` is `null`. Live against `constants.ensSubgraph["1"]` on 2026-08-24:

```
$ curl -s -X POST https://subgrapher.snapshot.org/subgraph/arbitrum/5XqPmWe6gjyrJtFn9cLy237i4cWw2j9HcUJEXsP5qGtH \
    -d '{"query":"query { account(id: \"0xab528d626ec275e3fad363ff1393a41f581c5897\") { domains { name expiryDate } wrappedDomains { name expiryDate } } }"}'

{"data":{"account":{"domains":[
  {"name":null,"expiryDate":null},
  {"name":"[eb9b3856c3238235793804382bb5c110a8fd10e7e0b6957146d50c81db8e5481].0x63616e.eth","expiryDate":null}
],"wrappedDomains":[]}}}
```

That address is the ENS root-node owner on mainnet, so this is reproducible on demand, not a hypothetical row shape.

`Number(null ?? 0)` evaluates to `0`, so `!expiry` is `true` and the `&&` short-circuit falls straight through to `null.endsWith('.addr.reverse')`. Replaying that exact payload through a jest test against the real resolver:

```
TypeError: Cannot read properties of null (reading 'endsWith')
```

Shapes that are safe, for contrast: `{name: "foo.eth", expiryDate: null}` and `{name: "foo.eth", expiryDate: "0"}` both pass the filter unharmed — it is specifically a null `name` that throws, and the null `expiryDate` only matters because it routes execution to the second operand.

## Root cause

1. `src/resolvers/lookupDomains/ens.ts:79-90` — the `account(id:)` query asks for `domains { name expiryDate }`, and the subgraph schema makes `name` nullable. Nothing between the response and the filter narrows that.
2. `ens.ts:88` — the filter treats `domain.name` as a `string`. The local `Domain` type declares `name: string`, but `graphQlCall<T>` is an unchecked cast, so TypeScript never sees the mismatch.
3. The throw escapes `lookupDomains` and is caught by the dispatcher at `src/resolvers/lookupDomains/index.ts`, which captures it to Sentry (tagged `provider: 'Ens'`) and substitutes `[]`.

Net effect: an address that owns a real, resolvable ENS name gets an empty list back, and every request for it emits a Sentry event. The failure is total for that address — one bad row discards the whole array, including the valid names alongside it.

Repo sweep for the same idiom: `domain.name` is dereferenced unguarded at exactly two sites, this filter and `getLabelHashes` (`ens.ts:26`). `src/resolvers/address/ens.ts` queries by `name_in`, so its rows always carry a name and it is not exposed.

## Impact

**Unmeasured.** No Sentry occurrences in 90 days (see header), so the reachable-but-unobserved framing is deliberate — please do not read a severity into it that the data does not support.

What would settle it: the count of distinct addresses reaching `lookup_domains` whose subgraph response contains a nameless row. There is no metric for that today; the cheapest proxy is watching for `TypeError` events tagged `provider: 'Ens'` on `lookup_domains_response_duration_seconds{status="0"}` once the guard is in place, or a one-off sweep of the subgraph for accounts owning a `name: null` domain.

## Provenance

Not a regression. The unguarded dereference dates to `9b38988` (2024-06-04), the commit that introduced `lookupDomains` (#231), and has been present continuously since. Today's `786c16c` ("compare ENS expiry numerically") rewrote the first operand of this same expression but does not change the exposure: the previous form was `(!domain.expiryDate || domain.expiryDate === '0' || domain.expiryDate > now)`, and `!null` was already `true`, so a null `name` reached `.endsWith` before that commit too.

## Fix

Guard inside the filter, so a nameless row is dropped rather than carried forward:

```diff
-    return (!expiry || expiry > now) && !domain.name.endsWith('.addr.reverse');
+    return domain.name && (!expiry || expiry > now) && !domain.name.endsWith('.addr.reverse');
```

**Optional chaining is not a correct fix here.** `domain.name?.endsWith('.addr.reverse')` returns `undefined`, which negates to `true`, so the nameless row *survives* the filter and is passed to `fetchDomainNames` — moving the same crash to `getLabelHashes` at `ens.ts:26`, where `domain.name.matchAll(...)` throws instead. The row has to be dropped, not tolerated.

Acceptance: one unit test replaying an `account` response containing `{name: null, expiryDate: null}` alongside a valid name, asserting the valid name is still returned and nothing throws.

## Sweep

- Checked open issues for the same root cause: #567 (each resolver validates the shape of *its input*) is the opposite direction — this is an unvalidated *upstream response* row; #566 is sibling-TLD exclusion in `address/ens.ts`, `address/unstoppableDomains.ts` and `image/ens.ts`; closed #539 covers the image family's no-data contract, and closed #272 covers `data.data` envelope destructuring, not row fields. None of them cover this.
- Checked open PRs touching `src/resolvers/lookupDomains/ens.ts`: #597 rewrites this function but keeps `!domain.name.endsWith('.addr.reverse')` unguarded, so it does not fix it. It already has a review, so this is filed as an issue rather than dropped into that round.

tony8713:
The ENS subgraph can return an account `domains` row with `name: null` (reproducible today against `constants.ensSubgraph["1"]` for the ENS root-node owner). The expiry filter in `lookupDomains/ens.ts` dereferences `domain.name` unconditionally, so that one row throws `TypeError: Cannot read properties of null (reading 'endsWith')`, which discards every valid name for the address alongside it, not just the bad row. The dispatcher in `lookupDomains/index.ts` catches the throw, reports it to Sentry, and returns `[]`.

Guard the filter so a nameless row is dropped instead of crashing the whole batch. Optional chaining would be wrong here: `domain.name?.endsWith(...)` evaluates to `undefined`, which negates to `true`, so the nameless row would survive the filter and reach `fetchDomainNames`, moving the same crash to `getLabelHashes` instead of fixing it.

Closes #618

### DIFF
--- a/src/resolvers/lookupDomains/ens.ts
+++ b/src/resolvers/lookupDomains/ens.ts
@@ -85,7 +85,7 @@ export default async function lookupDomains(
   ].filter(domain => {
     const expiry = Number(domain.expiryDate ?? 0);
 
-    return (!expiry || expiry > now) && !domain.name.endsWith('.addr.reverse');
+    return domain.name && (!expiry || expiry > now) && !domain.name.endsWith('.addr.reverse');
   });
 
   return fetchDomainNames(domains, chainId);
--- a/test/unit/resolvers/lookupDomains/ens.test.ts
+++ b/test/unit/resolvers/lookupDomains/ens.test.ts
@@ -15,7 +15,7 @@ function graphQlResponse<T>(data: T) {
   return { data };
 }
 
-function accountResponse(domains: { name: string; expiryDate?: string }[]) {
+function accountResponse(domains: { name: string | null; expiryDate?: string | null }[]) {
   return graphQlResponse({ account: { domains, wrappedDomains: [] } });
 }
 
@@ -44,6 +44,14 @@ describe('lookupDomains/ens', () => {
       'subdomain.eth',
       'far-future.eth'
     ]);
+  });
+
+  it('drops a nameless row instead of throwing', async () => {
+    mockedGraphQlCall.mockResolvedValueOnce(
+      accountResponse([{ name: null, expiryDate: null }, { name: 'alice.eth' }])
+    );
+
+    await expect(lookupDomains(ADDRESS)).resolves.toEqual(['alice.eth']);
   });
 
   it('loads all hashed labels in one domains query', async () => {


wa0x6e (approved):
utAck

PR #624, opened 2026-08-25
refactor(api): clean up image route id parsing

### LINKED ISSUE #601: refactor(api): clean up image route id parsing
`parseQuery` (`src/helpers/api.ts:12`) is the single entry point for every image route id. Four small cleanups, all in that function or caused by it.

### 1. Trim the id once, at the entry point

`parseQuery` normalizes with `address.toLowerCase()` but never trims, so each resolver is on its own. Only `resolvers/image/snapshot.ts` does anything about it, twice (`:106`, `:116`); `ens`, `starknet`, `lens` etc. get the raw string.

```ts
address = address.trim().toLowerCase();
```

Both trims in `snapshot.ts` then become dead. Note this is a behavior change for every resolver, and cache keys for `%200x1` and `0x1` collapse into one entry (`getCacheKey` takes `address`) — which is the point, but worth stating.

### 2. The `did:` branch is unreachable

`src/helpers/api.ts:25` handles `id.startsWith('did:')`, but the colon-count branches above it run first:

- `did:ethr:0xabc` → 3 chunks → `network = 'ethr'`, `networkId = 'eth'`; address is right, network is garbage.
- `did:0xabc` → 2 chunks → `networkId = 'did'`, `network = '1'` via the fallback.
- Only 4+ chunks reach the branch, and `id.slice(4)` then leaves the remaining colons in the address.

One occurrence in the repo, no test, no caller we know of. Either delete it, or hoist it above the split if DIDs are still expected — needs a call on which.

### 3. `parseQuery` is `async` with nothing to await

Both call sites `await` it (`src/api.ts:61`, `:82`), and the try/catch at `src/api.ts:81-89` exists only to wrap that call (`500 failed to load content`). No path in the body throws. Make it sync and drop the wrapper.

### 4. No unit test for `parseQuery`

Only e2e covers it today. If the above lands, one table test over the id shapes (plain, `shortName:addr`, `chainId:network:addr`, `did:`, whitespace) is the regression check.

### Out of scope

- Unifying `parseQuery` with the POST route's `normalizeAddresses` / `normalizeHandles` — different semantics on purpose (checksum vs lowercase-for-cache-key).
- `spaceIds` using strict `isStarknetAddress` while the image resolvers use `isStarknetFelt` — deliberate, unpadded felts are user-typed input.

tony8713:
Fixes #601

`parseQuery` (`src/helpers/api.ts`) is the single entry point for every image route id. It normalized case but never trimmed whitespace, so every resolver reached via the route (`ens`, `starknet`, `lens`, ...) except `snapshot.ts` got the raw string, and a padded vs unpadded id landed as two separate cache entries since `getCacheKey` hashes the address.

It also carried a `did:` branch that never produced a usable id (`id.slice(4)` still leaves the remaining colons in the address, and the only other thing it set was never returned), and it was `async` with nothing inside it to `await`, so both call sites in `src/api.ts` awaited a value that never needed it, and the route handler wrapped the call in a try/catch whose only job was to catch a throw the function cannot produce.

- Trim and lowercase the id once, up front, per colon-separated chunk, so a padded or differently-cased shortName/network segment resolves the same network as its clean equivalent, not just the address.
- Delete the `did:` branch: it only ran for 4+ colon-segment ids and never resolved anything, so removing it changes the fallback identicon and cache key for that shape, not behaviour anyone relied on.
- Make `parseQuery` synchronous, drop the dead try/catch in `src/api.ts` and both `await`s.
- Add a unit test table over `parseQuery`'s id shapes (plain, shortName:address, chainId:network:address, multi-segment did, whitespace in various positions, mixed case), its first unit coverage; previously only e2e touched it.
- Drop `DID` from the identifier list in `README.md`, since no code path has resolved one since the resolver's own DID handling was removed years ago.

`snapshot.ts`'s own two trims stay: `normalizeUserId`/`normalizeSpaceId` are called directly by `test/unit/resolvers/image/noData.test.ts` with padded input, bypassing `parseQuery` entirely, so they're load-bearing for a caller that skips the route layer, not dead code. Trimming twice on the route path is a no-op.

### DIFF
--- a/src/helpers/api.ts
+++ b/src/helpers/api.ts
@@ -10,10 +10,11 @@ export function sha256(str) {
 }
 
 export function parseQuery(id: string, type: ResolverType, query) {
+  id = id.trim().toLowerCase();
   let address = id;
   let network = '1';
   let networkId: string | undefined = undefined;
-  const chunks = id.split(':');
+  const chunks = id.split(':').map(chunk => chunk.trim());
   if (chunks.length === 2) {
     address = chunks[1];
     networkId = chunks[0];
@@ -24,7 +25,7 @@ export function parseQuery(id: string, type: ResolverType, query) {
     networkId = chainIdToShortName(network) || 'eth';
   }
 
-  address = address.trim().toLowerCase();
+  address = address.toLowerCase();
   const size = 64;
   const maxSize = type.includes('-cover') ? constants.maxCover : constants.max;
   let s = query.s ? parseInt(query.s) : size;
--- a/src/helpers/api.ts
+++ b/src/helpers/api.ts
@@ -9,7 +9,7 @@ export function sha256(str) {
   return createHash('sha256').update(str).digest('hex');
 }
 
-export async function parseQuery(id: string, type: ResolverType, query) {
+export function parseQuery(id: string, type: ResolverType, query) {
   let address = id;
   let network = '1';
   let networkId: string | undefined = undefined;
@@ -22,11 +22,9 @@ export async function parseQuery(id: string, type: ResolverType, query) {
     address = chunks[2];
     network = chunks[1];
     networkId = chainIdToShortName(network) || 'eth';
-  } else if (id.startsWith('did:')) {
-    address = id.slice(4);
   }
 
-  address = address.toLowerCase();
+  address = address.trim().toLowerCase();
   const size = 64;
   const maxSize = type.includes('-cover') ? constants.maxCover : constants.max;
   let s = query.s ? parseInt(query.s) : size;
--- a/test/unit/helpers/api.test.ts
+++ b/test/unit/helpers/api.test.ts
@@ -25,14 +25,34 @@ describe('parseQuery()', () => {
       { address: '0xabc', network: '999999999999', networkId: 'eth' }
     ],
     [
-      'did: is treated as a shortName, not stripped',
-      'did:0xABC',
-      { address: '0xabc', network: '1', networkId: 'did' }
+      'multi-segment did is treated as an opaque id, not stripped',
+      'did:pkh:eip155:1:0xABC',
+      { address: 'did:pkh:eip155:1:0xabc', network: '1', networkId: undefined }
     ],
     [
       'whitespace around the id is trimmed',
       '  0xABC  ',
       { address: '0xabc', network: '1', networkId: undefined }
+    ],
+    [
+      'whitespace-padded shortName:address still resolves the network',
+      ' matic:0xABC',
+      { address: '0xabc', network: '137', networkId: 'matic' }
+    ],
+    [
+      'whitespace inside a shortName:address id is trimmed per chunk',
+      'matic :0xABC',
+      { address: '0xabc', network: '137', networkId: 'matic' }
+    ],
+    [
+      'whitespace inside a chainId:network:address id is trimmed per chunk',
+      '1: 137 :0xABC',
+      { address: '0xabc', network: '137', networkId: 'matic' }
+    ],
+    [
+      'uppercase shortName:address is normalized before the network lookup',
+      'MATIC:0xABC',
+      { address: '0xabc', network: '137', networkId: 'matic' }
     ]
   ])('%s', (_name, id, expected) => {
     const result = parseQuery(id, 'avatar', {});
--- a/test/unit/helpers/api.test.ts
+++ b/test/unit/helpers/api.test.ts
@@ -1,0 +1,43 @@
+import { parseQuery } from '../../../src/helpers/api';
+
+describe('parseQuery()', () => {
+  it('is synchronous', () => {
+    const result = parseQuery('0xabc', 'avatar', {});
+    expect(result).not.toBeInstanceOf(Promise);
+  });
+
+  it.each([
+    ['plain address', '0xABC', { address: '0xabc', network: '1', networkId: undefined }],
+    ['shortName:address', 'matic:0xABC', { address: '0xabc', network: '137', networkId: 'matic' }],
+    [
+      'unknown shortName:address falls back to mainnet',
+      'bogus:0xABC',
+      { address: '0xabc', network: '1', networkId: 'bogus' }
+    ],
+    [
+      'chainId:network:address',
+      '1:137:0xABC',
+      { address: '0xabc', network: '137', networkId: 'matic' }
+    ],
+    [
+      'unknown chainId:network:address falls back to eth',
+      '1:999999999999:0xABC',
+      { address: '0xabc', network: '999999999999', networkId: 'eth' }
+    ],
+    [
+      'did: is treated as a shortName, not stripped',
+      'did:0xABC',
+      { address: '0xabc', network: '1', networkId: 'did' }
+    ],
+    [
+      'whitespace around the id is trimmed',
+      '  0xABC  ',
+      { address: '0xabc', network: '1', networkId: undefined }
+    ]
+  ])('%s', (_name, id, expected) => {
+    const result = parseQuery(id, 'avatar', {});
+    expect(result.address).toBe(expected.address);
+    expect(result.network).toBe(expected.network);
+    expect(result.networkId).toBe(expected.networkId);
+  });
+});
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ cdn.stamp.fyi/**token**/0x0bc529c00C6401aEF6D220BE8C6Ea1667F6Ad93e
 
 ### Identifier
 
-The identifier can be an address (case insensitive), ENS name, Basename, CAIP-10, EIP-3770, DID or Starknet domain.
+The identifier can be an address (case insensitive), ENS name, Basename, CAIP-10, EIP-3770 or Starknet domain.
 
 #### Examples
 
--- a/src/api.ts
+++ b/src/api.ts
@@ -58,7 +58,7 @@ router.get(`/clear/:type(${TYPE_CONSTRAINTS})/:id`, async (req, res) => {
     if (type === 'address' || type === 'name') {
       result = await clearCache(id, type);
     } else {
-      const { address, network, w, h, fallback, cb, fit } = await parseQuery(id, type, {
+      const { address, network, w, h, fallback, cb, fit } = parseQuery(id, type, {
         s: constants.max,
         fb: req.query.fb,
         cb: req.query.cb,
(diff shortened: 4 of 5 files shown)


src/helpers/api.ts
```diff
@@ -9,7 +9,7 @@ export function sha256(str) {
   return createHash('sha256').update(str).digest('hex');
 }
 
-export async function parseQuery(id: string, type: ResolverType, query) {
+export function parseQuery(id: string, type: ResolverType, query) {
```
wa0x6e, line 12: **[major] The trim runs after `id.split(':')`, so padded prefixed ids still resolve on the wrong chain — and now collide in cache with an unrelated request.**

Flagged independently by all four panel reviewers; confirmed by two verification passes, one of which measured it against the live CDN.

`address = id` is read, and `chunks = id.split(':')` computed, before any trimming — the `.trim()` was only added to the final `address` assignment. A space on a colon-prefixed id survives into `networkId`, `shortNameToChainId(' matic')` misses (no such key in `src/chains.json`), and `network` silently falls back to `'1'`.

```
parseQuery(' matic:0xABC', 'avatar', {}) → { address: '0xabc', network: '1', networkId: ' matic' }   // should be '137'
parseQuery('0xABC',        'avatar', {}) → { address: '0xabc', network: '1', networkId: undefined }
getCacheKey(...) for both → identical hash 7a521602fa544f6f87077ecfedd6172875142d3522afe1be0ba591b6a5203318
```

Against `cdn.stamp.fyi` today: `GET /token/%20%20matic:0x0bc529…` returns the same 1654 bytes / sha `82969fa4…` as the plain address, while canonical `matic:0x0bc529…` returns 284 bytes / `1c395df3…`. Express decodes `%20` in path params, so this is reachable from unauthenticated traffic, and `README.md:31,38-39` documents `shortName:address` / `chainId:network:address` as first-class public identifiers.

Two notes on the fix shape, both found while verifying:

1. Trimming only at the split (`id.trim().split(':')`) **breaks this PR's own `'  0xABC  '` test** — `let address = id` is never overwritten when the id has no colon, so the trim has to land before that initializer.
2. An outer trim alone still leaves per-component padding broken: `'matic :0xABC'` → `networkId 'matic '`, and `'1: 137 :0xABC'` → `network ' 137 '` → `networkId 'eth'`.

```diff
 export function parseQuery(id: string, type: ResolverType, query) {
+  id = id.trim();
   let address = id;
   let network = '1';
   let networkId: string | undefined = undefined;
-  const chunks = id.split(':');
+  const chunks = id.split(':').map(chunk => chunk.trim());
@@
-  address = address.trim().toLowerCase();
+  address = address.toLowerCase();
```

```diff
   [
     'whitespace around the id is trimmed',
     '  0xABC  ',
     { address: '0xabc', network: '1', networkId: undefined }
-  ]
+  ],
+  [
+    'whitespace-padded shortName:address still resolves the network',
+    ' matic:0xABC',
+    { address: '0xabc', network: '137', networkId: 'matic' }
+  ],
+  [
+    'whitespace inside a shortName:address id is trimmed per chunk',
+    'matic :0xABC',
+    { address: '0xabc', network: '137', networkId: 'matic' }
+  ],
+  [
+    'whitespace inside a chainId:network:address id is trimmed per chunk',
+    '1: 137 :0xABC',
+    { address: '0xabc', network: '137', networkId: 'matic' }
+  ]
```

⚠️ **The three source edits are interdependent — please don't apply them piecemeal.** Taking the `address.toLowerCase()` line without the `id = id.trim()` line regresses the padded plain-address case this PR already tests. That is also why this is one comment rather than three suggestion blocks (and the `.map()` line falls outside both diff hunks, so GitHub can't anchor a suggestion on it).

Verified: `npx jest -i test/unit/helpers/api.test.ts` → 11/11 pass, `npx eslint` clean on both files.

test/unit/helpers/api.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    [
+      'unknown chainId:network:address falls back to eth',
+      '1:999999999999:0xABC',
+      { address: '0xabc', network: '999999999999', networkId: 'eth' }
+    ],
+    [
+      'did: is treated as a shortName, not stripped',
+      'did:0xABC',
+      { address: '0xabc', network: '1', networkId: 'did' }
+    ],
```
wa0x6e, line 31: **[minor] The deleted `did:` branch *was* reachable — and this test row doesn't cover the change it's meant to cover.**

Flagged by three reviewers and confirmed by two verification passes plus a codex cross-check.

The branch was skipped only for 2- and 3-chunk ids; it ran for every 4+ chunk DID. Measured against a reconstructed pre-diff `parseQuery`:

```
did:0xABC              → PRE {address:'0xabc', networkId:'did'}   POST identical    ← this test row
did:ethr:0xABC         → PRE {address:'0xabc'}                    POST identical    ← 3 chunks, never affected
did:pkh:eip155:1:0xABC → PRE {address:'pkh:eip155:1:0xabc'}       POST {address:'did:pkh:eip155:1:0xabc'}
```

So the `'did:0xABC'` row passes against the *old* code too — it exercises no changed code. The one class whose output actually moved (4+ chunks) is untested.

**The deletion is still right, and for a stronger reason than the PR body gives.** `ccf6be36` (2022-02-20, "Separate resolvers") introduced this branch already positioned after the colon-count checks, and its only other output — a `format` variable — was never returned, just `console.log`'d. The README in that same commit listed `TO-DO: Ceramic IDX`. So the branch has been inert since the day it landed: `id.slice(4)` turns `did:pkh:eip155:1:0xABC` into `pkh:eip155:1:0xabc`, which is still colon-laden and unusable by every resolver. Nothing has ever resolved a DID identifier.

Suggested: replace the row with one that pins the real delta.

```suggestion
    [
      // Multi-segment DIDs (did:pkh:..., did:ethr:..., did:key:...) have 4+ colon
      // chunks, so they skip both branches above and fall through untouched: the
      // `did:` prefix stays part of the opaque id. Nothing resolves a DID — the
      // branch removed here was inert from the day it landed (ccf6be36, 2022):
      // its only other output, `format`, was never returned, and `id.slice(4)`
      // still left an unusable colon-laden string. This row pins the fallback
      // identicon's seed / cache key, nothing more.
      'multi-segment did is treated as an opaque id, not stripped',
      'did:pkh:eip155:1:0xABC',
      { address: 'did:pkh:eip155:1:0xabc', network: '1', networkId: undefined }
    ],
```

One wording fix in the PR description, so the record is accurate for whoever reads it next: "a `did:` branch that could never run" → it *did* run, for 4+ chunk ids; what it could never do is produce anything usable. Removing it changes those ids' fallback identicon and S3 cache key, which is worth stating as deliberate.

Verified: `npx jest -i test/unit/helpers/api.test.ts` passes with this row.

---

**And one more DID leftover, in `README.md:31`:**

> The identifier can be an address (case insensitive), ENS name, Basename, CAIP-10, EIP-3770, **DID** or Starknet domain.

Since this PR removes the last DID-specific code, that line should drop `DID` to match:

```diff
-The identifier can be an address (case insensitive), ENS name, Basename, CAIP-10, EIP-3770, DID or Starknet domain.
+The identifier can be an address (case insensitive), ENS name, Basename, CAIP-10, EIP-3770 or Starknet domain.
```

Not a regression from this change — it never had a working code path. (`src/resolvers/image/selfid.ts` isn't a counterexample: it takes an *address* and derives a DID internally to query Ceramic, so it's an avatar source rather than DID input support, and it's been unreachable since `303e1072` dropped it from the `constants.json` resolver lists.)

`README.md` has no hunk in this diff, so this can't be its own inline comment — folding it in here since it's the same concern.

src/helpers/api.ts
```diff
@@ -22,11 +22,9 @@ export async function parseQuery(id: string, type: ResolverType, query) {
     address = chunks[2];
     network = chunks[1];
     networkId = chainIdToShortName(network) || 'eth';
-  } else if (id.startsWith('did:')) {
-    address = id.slice(4);
   }
 
-  address = address.toLowerCase();
+  address = address.trim().toLowerCase();
```
wa0x6e: **FYI, no change requested here** — filed as #635.

While reviewing this file I found that `getCacheKey` (`src/helpers/api.ts:50-74`, no hunk in this diff) hashes `network` but not `networkId`, while the resolvers branch on `networkId`. Since several shortNames share a chainId while pointing at different upstreams (`s` and `eth` are both chainId `1` → `hub.snapshot.org` vs `api.snapshot.box`), and unrecognized shortNames fall back to `network '1'`, colon-prefixed ids can be served the plain address's cached image. Reproduced live: `/avatar/bogus:0x…543e7` resolves to `66f74a1c…` on its own but serves `619b9374…` once the plain address has warmed the cache.

**This PR does not cause it** — the collision reproduces identically before and after this diff, with no whitespace involved. I checked specifically because the panel initially flagged it as widened by the trim; it isn't. Noting it here only so it isn't re-discovered as a regression from this change. Details, evidence and a ready patch are in #635.

src/helpers/api.ts
wa0x6e, line 12: **[minor] The shortName chunk is never lowercased, so `MATIC:0xABC` silently resolves on mainnet.** Same root cause as the trim comment above — normalization runs after the lookup, and only on the address chunk.

```
parseQuery('MATIC:0xABC', 'avatar', {}) → { network: '1',   networkId: 'MATIC' }
parseQuery('matic:0xABC', 'avatar', {}) → { network: '137', networkId: 'matic' }
```

`shortNameToChainId` is a plain object key lookup (`src/helpers/chains.ts:4`), so it's case-sensitive: `shortNameToChainId('MATIC')` is `null` and `network` falls back to `'1'`. `README.md:31` promises the identifier is "case insensitive", which today holds only for the address portion.

This wants the *same line* as the trim fix, so please merge rather than stacking two statements:

```diff
-  id = id.trim();
+  id = id.trim().toLowerCase();
```

Safe to move the lowercasing earlier: `address` was already being lowercased further down, so the only behaviour that changes is that the shortName and network chunks now get normalized too — which is the point. Deliberately not posting this as a second suggestion block on line 12, since GitHub would offer two competing "insert here" suggestions and applying both would leave redundant statements.

wa0x6e (changes requested):

src/helpers/api.ts
tony8713: Confirmed independently, exact numbers match (` matic:0xABC` -> `networkId: ' matic'`, and both cache keys hashed to `7a521602fa...` on this tree). Applied your diff essentially as-is: trim+lowercase moved before the split, chunks mapped through `.trim()`, and combined with the lowercase-early fix from your comment below on the same line. Added the three whitespace test rows plus one for the uppercase case. Pushed in 5d9397a; mutation-tested by reverting the fix locally, which reddened exactly those 4 new rows and nothing else.

tony8713: Confirmed (`MATIC:0xABC` -> network `'1'` instead of `'137'`). Merged into the same `id = id.trim().toLowerCase()` line as the fix above, so no redundant statement. Added a test row for it. Pushed in 5d9397a.

test/unit/helpers/api.test.ts
tony8713: Reproduced the PRE/POST outputs exactly, including `did:pkh:eip155:1:0xABC` -> `pkh:eip155:1:0xabc` before this PR vs `did:pkh:eip155:1:0xabc` after. You're right the old `'did:0xABC'` row (2 chunks) never reached the removed branch and didn't cover the actual delta. Replaced it with a 4-chunk row that does, left out the inline comment though (kept the test title as the only explanation; a comment that long describing history/rationale isn't something I want to leave for the next reader to maintain). Corrected the "could never run" wording in the PR body and dropped DID from README.md as you suggested. Pushed in 5d9397a.

src/helpers/api.ts
tony8713: Thanks for chasing that down and separating it out. Verified independently that the collision reproduces identically on this PR's base and head, no whitespace involved, so agreed it's not something to fix here. No change made in this PR; tracked in #635.

wa0x6e (approved):
utAck

PR #625, opened 2026-08-25
fix(http): bound image response reads by size, not just deadline

### LINKED ISSUE #607: fix(http): image responses are read with no size bound, only a 5s deadline
Found during review of #598. Not from Sentry — no occurrences observed; this is a reachable gap, not an active incident.

## Today

`readHttpImage` in `src/helpers/http.ts` gates a response on `response.ok` and on `Content-Type`, then buffers the whole body:

```ts
return Buffer.from(await response.arrayBuffer());
```

Nothing bounds how many bytes that is. The only limit on the read is the 5s `IMAGE_FETCH_BUDGET` deadline, which is a *time* bound, not a *size* bound — an upstream on a fast link can deliver a very large body well inside it. `Content-Length` is never inspected.

Swept `src/` for any size bound: the only hits are `express.json({ limit: '4mb' })` / `express.urlencoded({ limit: '4mb' })` in `src/index.ts:17-18`, which bound **inbound request bodies only**, and `LOCAL_NAME_MAX_BYTES` in `src/resolvers/image/lens.ts:8`, which bounds a name string. There is no cap on any outbound fetch.

This is not a regression from #598 — `axios@0.25` defaulted `maxContentLength: -1`, so the pre-`fetch` code had no cap either. It is pre-existing and unchanged, raised now because #598 moved both reads.

## Blast radius

Every image resolver funnels through this one function, so a single guard covers all ten call sites:

`ens.ts:32`, `lens.ts:57`, `coingecko.ts:41`, `trustwallet.ts:17`, `selfid.ts:15`, `farcaster.ts:58`, `space-sx.ts:49`, `snapshot.ts:146`, `basename.ts:8`, `starknet.ts:36,49`.

Several of those URLs are user-supplied and served by hosts we do not control — the ENS avatar text record, the Lens `picture` field, the self.id image src, the Farcaster hub avatar, the Starknet `profilePicture` and its metadata `image`. So the input side is attacker-chosen by design; that is the product, not the bug. The bug is that the response side is unbounded.

The buffer is then handed to sharp (`src/helpers/image.ts:51`). sharp's own `limitInputPixels` default (~268MP) bounds decoded pixel area but not the input buffer we allocate before it.

## Impact

Unmeasured. The mechanism is a large allocation per concurrent request, and `src/api.ts:137` fans out over the resolver chain with `Promise.all`, so one request can hold several image buffers at once. Whether that is reachable in practice depends on request concurrency and on whether any upstream actually serves bodies that large — neither of which I measured.

What would settle it: `process_resident_memory_bytes` for the stamp instances against request rate, and a distribution of `Content-Length` over real avatar responses. Worth checking before choosing a limit, so the number isn't arbitrary.

## Change

A declared-length check next to the existing media-type check in `readHttpImage`, throwing the same `httpError(host, 404, …)` so an oversized body lands as a routine miss like every other "not an image" case:

```ts
const length = Number(response.headers.get('content-length'));
if (length > MAX_IMAGE_BYTES) {
  throw httpError(new URL(url).host, 404, `image too large: ${length} bytes`);
}
```

Note what this does *not* do: `Content-Length` is advisory. An upstream that omits it, lies about it, or uses chunked transfer-encoding walks straight past this check. Bounding those needs a streaming read that counts bytes and aborts past the limit, rather than `arrayBuffer()`. Worth deciding up front which of the two is wanted — the declared-length guard is a few lines and covers the honest-but-huge case; the streaming one covers the hostile case and is a real change to how the body is read.

Related but distinct: #536 bounds the *time* axis of the same calls. This is the byte axis.

tony8713:
## Problem

Every image resolver reads its response through `fetchHttpImage`, or the sibling read in the Starknet resolver's own `fetchImageOrMetadata`, both of which buffered `Buffer.from(await response.arrayBuffer())` with no size limit, only the 5s fetch deadline. A fast upstream can deliver an arbitrarily large body well inside that budget, and several of the URLs involved are attacker-chosen by design (ENS avatar record, Lens picture, Starknet profile picture, and so on).

## Fix

Both read sites now go through `readBoundedImage` in `src/helpers/http.ts`. It rejects a declared `Content-Length` over `MAX_IMAGE_BYTES` before transferring anything, and otherwise counts bytes as the body streams, aborting once the running total crosses the cap, so an upstream that omits or lies about `Content-Length` (or uses chunked transfer encoding) is still bounded. Either path raises the same `httpError(host, 404, ...)` used elsewhere in this file, landing as a routine miss like a non-2xx or an empty body already do, and releases the upstream connection rather than leaving it open.

`MAX_IMAGE_BYTES` is 10MB, comfortably above any legitimate avatar or cover image (`constants.json` caps rendered output at 500x500 / 1500x1500).

Closes #607.

## Scope

The JSON metadata branch of the Starknet resolver's fetch (NFT `token_uri` lookups) is a separate, much smaller surface and is left unbounded here to keep this change scoped to the image-byte read the issue describes.

This is independent of the in-flight #598, which also rewrites `fetchHttpImage`'s read step, for a different reason (content-type gating). Whichever of the two lands second will need a small rebase.

### DIFF
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
@@ -30,8 +30,35 @@ export function fetchWithDeadline<T>(
   }, 5e3);
 }
 
+export const MAX_IMAGE_BYTES = 10 * 1024 * 1024;
+
+export async function readBoundedImage(url: string, response: Response): Promise<Buffer> {
+  const host = new URL(url).host;
+  const declared = Number(response.headers.get('content-length'));
+  if (declared > MAX_IMAGE_BYTES) {
+    await response.body?.cancel();
+    throw httpError(host, 404, `image too large: ${declared} bytes`);
+  }
+
+  if (!response.body) return Buffer.from(await response.arrayBuffer());
+
+  const chunks: Uint8Array[] = [];
+  let total = 0;
+
+  for await (const chunk of response.body) {
+    total += chunk.length;
+    if (total > MAX_IMAGE_BYTES) {
+      throw httpError(host, 404, `image too large: over ${MAX_IMAGE_BYTES} bytes`);
+    }
+
+    chunks.push(chunk);
+  }
+
+  return Buffer.concat(chunks);
+}
+
 export function fetchHttpImage(url: string): Promise<Buffer> {
-  return fetchWithDeadline(url, async response => Buffer.from(await response.arrayBuffer()));
+  return fetchWithDeadline(url, response => readBoundedImage(url, response));
 }
 
 export function getUrl(url: string): string | null {
--- a/src/resolvers/image/starknet.ts
+++ b/src/resolvers/image/starknet.ts
@@ -1,7 +1,7 @@
 import { byteArray, CallData, constants, shortString, starknetId } from 'starknet';
 import { isStarkDomain, isStarknetFelt } from '../../helpers/address';
 import { untilAborted, withDeadline } from '../../helpers/deadline';
-import { fetchHttpImage, fetchWithDeadline, getUrl } from '../../helpers/http';
+import { fetchHttpImage, fetchWithDeadline, getUrl, readBoundedImage } from '../../helpers/http';
 import { getProvider } from '../../helpers/provider';
 
 const DEFAULT_IMG_URL = 'https://starknet.id/api/identicons/0';
@@ -119,7 +119,7 @@ function fetchImageOrMetadata(url: string): Promise<Buffer | { image?: string }>
   return fetchWithDeadline(url, async response =>
     response.headers.get('content-type')?.includes('application/json')
       ? await response.json()
-      : Buffer.from(await response.arrayBuffer())
+      : readBoundedImage(url, response)
   );
 }
 
--- a/test/integration/helpers/http.test.ts
+++ b/test/integration/helpers/http.test.ts
@@ -1,21 +1,58 @@
 import http from 'http';
 import { AddressInfo, Socket } from 'net';
 import { isSilencedError } from '../../../src/helpers/errors';
-import { fetchHttpImage } from '../../../src/helpers/http';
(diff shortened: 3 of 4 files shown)


tony8713:
The red Test job here is not this branch. Both failing cases (\`test/e2e/api.test.ts\` "returns different space cover for different network" and \`test/integration/resolvers/image/snapshot.test.ts\` "matches the image snapshot for snapshot-space-cover") depend on live data for the \`test.wa0x6e.eth\` space: its mainnet cover is currently unset and the testnet space does not exist, so both requests fall through to the same blockie fallback.

Confirmed by pushing untouched master (3d49fb6) to a throwaway branch just now, run 32831607669: same two suites fail, same counts, no code involved.

    curl -s https://hub.snapshot.org/graphql -d '{"query":"{ space(id:\"test.wa0x6e.eth\") { cover } }"}'
    -> {"data":{"space":{"cover":null}}}

tony8713:
Follow-up on the latest push (dedup only, per review): the Test job on this head also shows \`test/integration/resolvers/image/lens.test.ts\` and \`test/integration/resolvers/address/lens.test.ts\` failing. Same shape as above, not this branch: the address suite never touches image code at all, so a change here cannot be the cause. Confirmed \`api.lens.xyz\` is timing out on the exact query the address resolver uses, and reproduced on untouched master pushed to a throwaway branch just now, run 32869696891.

    curl -m 15 -X POST https://api.lens.xyz/graphql -d '{"query":"query AccountBulk($request: AccountsBulkRequest!) { accountsBulk(request: $request) { username { localName ownedBy } } }","variables":{"request":{"addresses":["0x218F68106128E637fc942C2b1Ed1e3c326125344"]}}}'
    # times out

wa0x6e (approved):
utAck

PR #626, opened 2026-08-25
fix(errors): share the image path's routine-miss classifier with address/lookupDomains resolvers

### LINKED ISSUE #617: fix(errors): share the image path's routine-miss classifier with address/lookupDomains resolvers
PR #595 (closing #592) added `isRoutineMiss` to `src/resolvers/image/index.ts` to stop reporting third-party avatar-URL failures — dead DNS, refused connections, 4xx, 520-524, TLS cert errors — as Sentry events, since a failed avatar fetch is no-data, not a defect (STAMP-6P, STAMP-75).

The same class of error reaches the equivalent `capture` sites in the other resolver families, which still classify with `isSilencedError` alone and have no equivalent for `ENOTFOUND` / `EAI_AGAIN` / `ECONNREFUSED` / TLS-cert codes / plain 4xx:

- `src/resolvers/lookupDomains/index.ts:45`
- `src/resolvers/address/index.ts:72`
- `src/resolvers/address/ens.ts:99` and `:120`
- `src/resolvers/address/unstoppableDomains.ts:60`

## Reproduction

`src/resolvers/lookupDomains/shibarium.ts` fetches its REST API with native `fetch()`. A DNS failure on that host throws `TypeError: fetch failed` with `cause.code: 'ENOTFOUND'` — the same shape as STAMP-75, which #595 just fixed for the image path. It reaches `lookupDomains/index.ts`'s catch and is captured, unsilenced, because `isSilencedError`'s code list never included `ENOTFOUND`.

`test/integration/resolvers/lookupDomains/shibarium.test.ts:70-81` constructs this exact error but only exercises the raw provider module, never the `lookupDomains/index.ts` wrapper where `capture` actually fires — so its `expect(capture).not.toHaveBeenCalled()` assertion passes vacuously and the gap is untested as well as unfixed.

## Not covered by the open issues in this family

- #590 — ethers `CALL_EXCEPTION` wrapping an upstream 5xx, matched via `error.error?.status`.
- #606 — viem `TimeoutError` unrecognised, in `src/resolvers/image/ens.ts`.

Both are narrower instances of the same classifier-blindness family (#606's own text notes it is "the fourth time `isSilencedError` has missed a wrapped error"), but neither adds the `ENOTFOUND` / `EAI_AGAIN` / `ECONNREFUSED` / TLS-cert / plain-4xx classes anywhere.

## Suggested fix

Move `isRoutineMiss` and `ROUTINE_NETWORK_ERROR_CODES` from `src/resolvers/image/index.ts` into `src/helpers/errors.ts`, beside `isSilencedError`, and call it from the four sites above the same way `isSilencedError` is called today.

Add one regression test per site mirroring the rows #595 added to `test/unit/resolvers/image/failureContract.test.ts` — `ENOTFOUND`, a plain 4xx, and a TLS cert code, each expecting no `capture`. The shibarium integration test above should be driven through the wrapper, not the bare provider, so the assertion stops being vacuous.

Note the classification decisions themselves are still being settled on #595 (whether the 4xx band should be an explicit no-data allowlist instead, and whether the 520-524 band should exist at all given `hub.snapshot.org` and `api.snapshot.box` are both behind Cloudflare). Whatever shape lands there is what should move to `helpers/errors.ts` — this issue is about sharing it, not about redesigning it.

---

Found by a review panel on #595.

tony8713:
Closes #617.

PR #595 added `isRoutineMiss` to the image resolver wrapper to stop reporting third-party avatar-URL failures (dead DNS, refused connections, plain 4xx, TLS certificate errors) as Sentry events, since those are no-data results rather than resolver defects. The other resolver families reach the same error shapes but still classify with `isSilencedError` alone, which has no equivalent for `ENOTFOUND`/`EAI_AGAIN`/`ECONNREFUSED`, TLS cert codes, or plain 4xx, so a routine DNS failure or host 4xx on `lookupDomains`, `address`, `address/ens`, or `address/unstoppableDomains` still gets reported as a defect.

Moves `isRoutineMiss` (and its network error code list) out of the image resolver into `src/helpers/errors.ts`, beside `isSilencedError`, and calls it alongside `isSilencedError` at the four other capture sites named in the issue. Also drives the existing shibarium `ENOTFOUND` regression test through the `lookupDomains` wrapper instead of the bare provider module, since the assertion never reached `capture` from there.

`address/ens.ts`'s subgraph-fetch catch keeps its own `getAddress` parsing on a separate catch from the network fetch, so a malformed address returned by the subgraph itself is still reported rather than swallowed by the same routine-miss check meant for the fetch's network errors.

### DIFF
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
@@ -56,3 +56,29 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
     )
   );
 }
+
+export const ROUTINE_NETWORK_ERROR_CODES = [
+  'ENOTFOUND',
+  'EAI_AGAIN',
+  'ECONNREFUSED',
+  'ERR_TLS_CERT_ALTNAME_INVALID',
+  'CERT_HAS_EXPIRED',
+  'UNABLE_TO_VERIFY_LEAF_SIGNATURE',
+  'DEPTH_ZERO_SELF_SIGNED_CERT'
+];
+
+// 401/402/403 stay out of the routine band: they are how a dead or rotated
+// credential shows up on a resolver's own authenticated call, and silencing
+// them would hide that outage as a routine miss.
+const AUTH_STATUS_CODES = [401, 402, 403];
+
+export function isRoutineMiss(error: any): boolean {
+  const status = Number(error?.status ?? error?.response?.status);
+  const code = error?.cause?.code ?? error?.code;
+
+  return (
+    (status >= 400 && status < 500 && !AUTH_STATUS_CODES.includes(status)) ||
+    ROUTINE_NETWORK_ERROR_CODES.includes(code) ||
+    error?.code === 'INVALID_ARGUMENT'
+  );
+}
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
@@ -57,7 +57,7 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
   );
 }
 
-export const ROUTINE_NETWORK_ERROR_CODES = [
+const ROUTINE_NETWORK_ERROR_CODES = [
   'ENOTFOUND',
   'EAI_AGAIN',
   'ECONNREFUSED',
--- a/src/resolvers/address/ens.ts
+++ b/src/resolvers/address/ens.ts
@@ -93,7 +93,13 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
     );
 
     for (const item of items) {
-      results[item.name] = item.resolvedAddress ? getAddress(item.resolvedAddress.id) : '';
+      try {
+        results[item.name] = item.resolvedAddress ? getAddress(item.resolvedAddress.id) : '';
+      } catch (err) {
+        if (!isSilencedError(err)) {
+          capture(err, { input: { handles: normalizedHandles } });
+        }
+      }
     }
   } catch (err) {
     if (!isSilencedError(err) && !isRoutineMiss(err)) {
--- a/src/resolvers/address/ens.ts
+++ b/src/resolvers/address/ens.ts
@@ -4,7 +4,7 @@ import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
 import constants from '../../constants.json';
 import { isEvmAddress } from '../../helpers/address';
-import { isSilencedError } from '../../helpers/errors';
+import { isRoutineMiss, isSilencedError } from '../../helpers/errors';
 import { graphQlCall } from '../../helpers/graphql';
 import { getProvider } from '../../helpers/provider';
 import { Address, Handle } from '../../helpers/types';
@@ -96,7 +96,7 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
       results[item.name] = item.resolvedAddress ? getAddress(item.resolvedAddress.id) : '';
     }
   } catch (err) {
-    if (!isSilencedError(err)) {
+    if (!isSilencedError(err) && !isRoutineMiss(err)) {
       capture(err, { input: { handles: normalizedHandles } });
     }
   }
@@ -117,7 +117,7 @@ export async function resolveNames(handles: Handle[]): Promise<Record<Handle, Ad
       }
     });
   } catch (err) {
-    if (!isSilencedError(err)) {
+    if (!isSilencedError(err) && !isRoutineMiss(err)) {
       capture(err, { input: { handles: normalizedHandles } });
     }
   }
--- a/src/resolvers/address/index.ts
+++ b/src/resolvers/address/index.ts
@@ -15,7 +15,7 @@ import {
   normalizeHandles,
   withoutEmptyAddress
 } from '../../helpers/address';
-import { isSilencedError } from '../../helpers/errors';
+import { isRoutineMiss, isSilencedError } from '../../helpers/errors';
 import { timeAddressResolverResponse as timeResponse } from '../../helpers/metrics';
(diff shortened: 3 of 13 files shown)


wa0x6e (changes requested):
`isRoutineMiss` bundles four error classes, and only two of them mean the same thing on the image path and on the address/lookupDomains paths. Sharing the whole predicate silences real defects on the new call sites.

| Class | Share? |
|---|---|
| `ENOTFOUND` / `EAI_AGAIN` / `ECONNREFUSED` | yes |
| TLS cert codes | yes, with a caveat below |
| 4xx band | no |
| `INVALID_ARGUMENT` | no |

### The 4xx band means different things

Image resolvers fetch a **per-input, often user-controlled** URL — a trustwallet `<checksum>/logo.png`, an ENS avatar record that can point anywhere. A 404 there *is* the miss; that is what #595 was fixing.

Address and lookupDomains resolvers all hit a **fixed** endpoint: [`constants.ensSubgraph[NETWORK]`](https://github.com/snapshot-labs/stamp/blob/74445bc7f3a3b3e3ac1bf2fe4079d6b19918142c/src/resolvers/address/ens.ts#L83) via `graphQlCall`, `api.unstoppabledomains.com`, `constants.d3[chainId].apiUrl`, or an RPC. A miss on those comes back as HTTP 200 with an empty array — they cannot 404 for "this address has no name". So the only 4xx they can emit is ours: a decommissioned subgraph URL, a malformed query, a wrong path. After this change that is silent, and ENS name resolution can go fully dead with zero Sentry events.

shibarium makes the point concrete: it [already handles its own 404](https://github.com/snapshot-labs/stamp/blob/74445bc7f3a3b3e3ac1bf2fe4079d6b19918142c/src/resolvers/lookupDomains/shibarium.ts#L38) with a `break`, so the shared 4xx band mutes nothing real there and only costs signal on the failure case.

### `INVALID_ARGUMENT` has a different origin on each path

Its only source in this repo is ethers' `getAddress`.

On the image path it parses **caller input**: [`parseQuery`](https://github.com/snapshot-labs/stamp/blob/74445bc7f3a3b3e3ac1bf2fe4079d6b19918142c/src/helpers/api.ts#L12-L14) lowercases the path segment and validates nothing, so `GET /token/hello` reaches an unguarded [`getAddress(address)`](https://github.com/snapshot-labs/stamp/blob/74445bc7f3a3b3e3ac1bf2fe4079d6b19918142c/src/resolvers/image/trustwallet.ts#L12) (same in [coingecko](https://github.com/snapshot-labs/stamp/blob/74445bc7f3a3b3e3ac1bf2fe4079d6b19918142c/src/resolvers/image/coingecko.ts#L24)) and throws `INVALID_ARGUMENT`. Garbage in the URL, no image — routine, correctly. The other three image call sites (lens, farcaster, snapshot) already guard it, so the clause is load-bearing only for the `token` family.

On the address path the caller input is filtered upstream by `normalizeHandles` / `isEvmAddress`, so `getAddress` only ever parses an **upstream response** — the subgraph's `resolvedAddress.id`, `provider.resolveName`'s return, a contract return in `basename`/`gwei`. `INVALID_ARGUMENT` there means an upstream returned corrupt data. That is a defect, not a miss.

74445bc found half of this and gave [the subgraph parse](https://github.com/snapshot-labs/stamp/blob/74445bc7f3a3b3e3ac1bf2fe4079d6b19918142c/src/resolvers/address/ens.ts#L96-L102) its own catch. The [provider fallback](https://github.com/snapshot-labs/stamp/blob/74445bc7f3a3b3e3ac1bf2fe4079d6b19918142c/src/resolvers/address/ens.ts#L122) still swallows it — the test named *"classifies a malformed provider-fallback result as routine, defensively"* asserts no `capture` for exactly that corruption. Needing one carve-out already is the signal that the predicate does not transfer.

### DNS/TLS caveat

Worth sharing, but not for the stated reason. On the image path `ENOTFOUND` means one user's avatar host is dead. On a fixed endpoint it means the dependency is down for every request — the thing you would want paged for. Silencing it is defensible on volume grounds (that was the STAMP-75 repro), not as a "routine miss". Worth naming honestly if it stays.

### Precedent

12a1e524, the commit that introduced both contested clauses, ruled on this explicitly:

> an upstream 404, which in this family is the image not being there; an ethers argument error, which is the request carrying something that is not an address, since `parseQuery` validates nothing; … **The first two are policy for this family rather than for every caller, so they sit beside the failure contract and not in the shared classifier.**

This PR moves exactly those two into the shared classifier. #617's closing note anticipated the tension — *"the classification decisions themselves are still being settled on #595"*.

### Suggested shape

Split the predicate. A shared `isTransportFailure(err)` in `helpers/errors.ts` holding just the seven network/TLS codes, called at all five sites; the 4xx band and `INVALID_ARGUMENT` stay in `resolvers/image/index.ts` as family policy, where 12a1e524 put them. That also fixes the `ens.ts` provider-fallback hole for free, and drops the need for the defensive carve-out.

wa0x6e:
Split the `INVALID_ARGUMENT` half of my review out into #638 — it is a token-route problem that exists on `master` independently of this PR, and it does not belong in your scope here.

Short version: `parseQuery` validates nothing and `image/trustwallet.ts` / `image/coingecko.ts` parse the raw id unguarded, so `GET /token/hello` throws into the failure contract. Every other image resolver already returns no-data instead. Guarding those two removes the only caller-input source of `INVALID_ARGUMENT`, and the clause can then be deleted rather than shared.

Either merge order works — both touch the same line. If #638 lands first, this PR just carries one less class.

That leaves the 4xx band as the open question here, and the `isTransportFailure` split as the ask.

tony8713:
Superseded by #639, which implements the isTransportFailure split exactly as requested in review, everything this PR was doing is already merged. A rebase confirmed this branch now diffs to nothing against master.

Issue #627, opened 2026-08-25
fix(http): fetchHttpImage blocks the event loop decoding a large data: URL, and the request deadline cannot preempt it


tony8713:
`fetchHttpImage` (`src/helpers/http.ts:33`) hands whatever URL it is given straight to native `fetch()`. For a `data:` URL, Node/undici decodes the base64 payload synchronously inside the `fetch()` call before returning, and that decode is not preemptible: the same event loop that would run `withDeadline`'s `setTimeout` abort (`src/helpers/deadline.ts:11`, 5s budget for image fetches) cannot tick while the decode is running, so the deadline provides no protection against a large payload.

## Reproduction

Verified 2026-08-25 on Node v22.23.1, replicating `fetchHttpImage`/`withDeadline` exactly (5000ms budget):

```
80MB data: URI (raw bytes; ~112MB base64) -> fetchHttpImage total: 7392ms
5ms-interval timer ticks during that window: 0
```

A `setInterval(fn, 5)` armed before the call and cleared after did not fire even once in 7.4 seconds — the process was not idling, the whole request+decode is one synchronous stretch that starves the event loop, including the timer meant to bound it. The call still resolved successfully at 7392ms, 2.4s past its 5000ms budget, confirming the abort never had a chance to fire.

## Who is exposed today

Three resolvers hand a URL to `fetchHttpImage` directly, with no length check and no gate against `data:`:

- `src/resolvers/image/lens.ts:62` — `img_url` from a Lens profile's `metadata.picture`, attacker/subject-controlled.
- `src/resolvers/image/farcaster.ts:54` — `pfp_url` from Neynar, subject-controlled.
- `src/resolvers/image/coingecko.ts:41` — `data.image.large` from CoinGecko.

Resolvers that route through `getUrl` (`src/helpers/http.ts:37`, wrapping `snapshot.utils.getUrl`) are currently NOT exposed: on the installed `@snapshot-labs/snapshot.js`, `getUrl` doesn't recognize `data:` and rewrites it onto the IPFS gateway path instead (`${gateway}/ipfs/data:...`), which just fails as a malformed HTTP request rather than reaching `fetch()`'s native `data:` handling. That changes once a `data:`-aware `getUrl` lands (open PR #598 adds exactly that, for the Starknet NFT-metadata path) — at that point `starknet.ts`, `snapshot.ts`, `space-sx.ts`, `selfid.ts`, and `basename.ts` all become exposed too.

[...]

PR #628, opened 2026-08-25
fix(ens): treat a data: avatar record as a fetchable URL

### LINKED ISSUE #616: fix(ens): data: avatar records are discarded and refetched from the ENS metadata service
Found while reviewing #598, after #610 was folded into it. Not from Sentry — no occurrences observed and nothing on this path throws.

## Today

#610 (folded into #598, which closes #609) taught `getUrl` about `data:`:

```ts
// src/helpers/http.ts
export function getUrl(url) {
  ...
}
```

That covers every resolver that routes its reference through `getUrl` — `starknet.ts:48,52`, `snapshot.ts:132`, `space-sx.ts:49`, `selfid.ts:14`, and `basename.ts` via `src/resolvers/address/basename.ts:102`. `lens.ts:53`, `farcaster.ts:58` and `coingecko.ts:41` hand their URL straight to `fetchHttpImage`, and native `fetch` reads `data:` itself, so those work too.

`src/resolvers/image/ens.ts:27-32` is the one resolver that does not:

```ts
let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
url = url?.startsWith('http')
  ? url
  : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;

return await fetchHttpImage(url);
```

The gate is a string prefix test against `http`. A `data:image/*` avatar record does not match, so the record is discarded and replaced by a metadata.ens.domains lookup — and **that service answers 404 for exactly these records**, so the avatar never renders. The user gets a blockie while the image sits in the record we already read.

### Fixtures — the two `data:` avatar records on mainnet

Both verified 2026-08-25 against `rpc.snapshot.org` and the repo's `sharp`:

| name | record | direct fetch | `sharp` resize | metadata.ens.domains |
|---|---|---|---|---|
| `abefdfabbb.mainnet-ethereum.eth` | `data:image/png;base64,iVBORw0KGgo…` (1222 B) | 200 `image/png` 898 B | 3026 B webp | **404** in 2633 ms |
| `fcdfcfddfd.mainnet-ethereum.eth` | `data:image/gif;base64,R0lGODdhIAAg…` (726 B) | 200 `image/gif` 528 B | 4126 B webp | **404** in 2106 ms |

The 404 is caused by the record, not by the name shape. Controlled against four siblings under the same parent, same resolver, same label format:

```
eedecedb.mainnet-ethereum.eth    record: https://www.ungravel.com/ung.p…   metadata service: 200
ffbffedbbe.mainnet-ethereum.eth  record: https://www.ungravel.com/ung.p…   metadata service: 200
fedaaabbbd.mainnet-ethereum.eth  record: https://www.ungravel.com/ung.p…   metadata service: 200
cddfdecfde.mainnet-ethereum.eth  record: https://www.ungravel.com/ung.p…   metadata service: 200
abefdfabbb.mainnet-ethereum.eth  record: data:image/png;base64            metadata service: 404
fcdfcfddfd.mainnet-ethereum.eth  record: data:image/gif;base64            metadata service: 404
```

So the fallback does not rescue a `data:` record. (`@ensdomains/ens-avatar` has `isSVGDataUri` / `isImageDataUri` branches in `src/utils/getImageURI.ts`, but whatever metadata.ens.domains runs today does not reach them for these records.)

Nothing downstream in stamp needs `data:` support added — `fetch`, `readHttpImage`'s `image/*` gate (#598) and `sharp` all take the record as-is, as the table shows. Only the gate in `ens.ts` rejects it.

## Impact

**Measured, and small.** Scanned every mainnet ENS resolver with an `avatar` text key set, reading the value on-chain via multicall (2026-08-25, 30,994 records):

```
https://   24445
eip155:     6027
empty        405
ipfs://       95
other         19
http://        1
data:          2      <- 0.006%
```

Both `data:` records are the two names above — machine-generated subdomains under `mainnet-ethereum.eth`, not organic user avatars. No organic instance exists in the current mainnet set.

Per affected name the avatar is a permanent miss, not a transient failure: `fetchHttpImage` gets a 404 from the metadata service, `isRoutineMiss` forgives it, `withFailureContract` returns `false`, and `api.ts` falls through to a blockie. Nothing is reported to Sentry, which is why this never surfaced as an error.

Sizing this honestly: two names today. It is filed as a correctness gap and a cheap one to close, not as an incident. The value is mostly in not having the one resolver that reads a raw on-chain record be the one that cannot read the on-chain format.

## Why file it separately rather than fold it into #605

#605 rewrites this exact line, and its suggested fix would cement the drop:

```ts
const httpUrl = z.string().url().refine(u => /^https?:$/.test(new URL(u).protocol));
url = httpUrl.safeParse(record).success ? record : `https://metadata.ens.domains/…`;
```

Both halves of that check are `http(s)`-only, so a `data:` record still routes to the fallback after it lands. The `data:` branch belongs in the same change; filed separately so it is not lost, since the two have different acceptance criteria.

Also worth correcting the record: #615's *What this is not* section states the `data:` cell is already handled. That is true of `getUrl`, not of `ens.ts`.

## Fix shape

Two options, both one line:

1. Add `data:` to the gate alongside `http`:

   ```ts
   url = url?.startsWith('http') || url?.startsWith('data:')
     ? url
     : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;
   ```

2. Route the record through `getUrl` like every other resolver, keeping the "what is a fetchable reference" knowledge in one place — the same argument #615 makes. That also picks up the 95 `ipfs://` records without the metadata-service round-trip, which is a behaviour change worth deciding on purpose rather than as a side effect.

One thing to decide with it: for the names it does resolve, the metadata service runs on-chain SVG through DOMPurify before returning it; reading the record directly skips that. Not a new exposure — SVG already reaches `sharp` unsanitized from every `getUrl` resolver, and `data:image/svg+xml` from the Starknet path since #598 — but the ENS path specifically loses a sanitisation step it has today, so it should be a conscious trade. Size is bounded by what the record cost to store on-chain, so #607 (no size bound on image reads) is not made worse here.

## Acceptance criteria

- `abefdfabbb.mainnet-ethereum.eth` and `fcdfcfddfd.mainnet-ethereum.eth` render their real avatar from the record, with no request to metadata.ens.domains.
- Records that are neither `data:` nor `http(s)` (`ipfs://`, CAIP `eip155:1/erc721:…`, empty, malformed) keep their current destination.
- One regression test per shape: `data:image/png;base64,…`, `data:image/svg+xml;base64,…`, `https://…`, `ipfs://…`, `eip155:1/erc721:…/1`, and no record at all.

## Sweep

- `coverage-sweep.sh -R snapshot-labs/stamp --topic "data:" --topic "data URI" --topic "avatar record" --file src/resolvers/image/ens.ts`, positive control `getUrl` returns 6 issues.
- Related but distinct: **#609** (`getUrl` mangled `data:` onto the IPFS gateway — fixed by #598; Starknet fixtures listed there), **#615** (`getUrl` builds unfetchable URLs for other non-CID references), **#605** (`ens.ts` string-matches `http`, malformed records reach `fetch`), **#608** (no scheme/host allow-list on user-supplied avatar URLs). None covers a valid `data:` record being discarded by `ens.ts`.
- Open PRs touching `src/resolvers/image/ens.ts`: **#580** only, and it changes `castToEnsName` (sibling-TLD gate), not the avatar-record line. It is already under review, so this is filed here rather than raised in that thread.

tony8713:
Closes #616.

`src/resolvers/image/ens.ts` decides whether an ENS `avatar` text record is a fetchable URL with `url?.startsWith('http')`. A `data:` URI avatar record fails that check even though it's already a usable image reference, so it was discarded and the resolver fell through to a `metadata.ens.domains` lookup, which cannot resolve a `data:` record either — the avatar was lost even though it was already in hand.

Accept a `data:` value in the same gate, requiring the RFC 2397 comma so a truncated or garbled value (e.g. bare `data:`) still falls through to the metadata-service fallback instead of reaching `fetch()` as if it were real. Records that are neither `http(s)` nor a well-formed `data:` URI (`ipfs://`, CAIP references, empty, malformed) keep going to the same fallback as before.

Filed #627 separately for a related but distinct finding from review: a large `data:` payload blocks Node's event loop synchronously inside `fetch()`, and the existing request deadline can't preempt that. It's pre-existing (lens, farcaster and coingecko already pass arbitrary `data:` URLs to the same `fetchHttpImage` unconditionally), not introduced by this change, and needs a fix that applies to all of those call sites rather than just this one.

### DIFF
--- a/src/resolvers/image/ens.ts
+++ b/src/resolvers/image/ens.ts
@@ -26,7 +26,7 @@ export default async function resolve(nameOrAddress: string) {
 
   let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
   url =
-    url?.startsWith('http') || url?.startsWith('data:')
+    url?.startsWith('http') || (url && /^data:[^,]*,/.test(url))
       ? url
       : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;
 
--- a/src/resolvers/image/ens.ts
+++ b/src/resolvers/image/ens.ts
@@ -25,9 +25,10 @@ export default async function resolve(nameOrAddress: string) {
   if (!ensName) return false;
 
   let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
-  url = url?.startsWith('http')
-    ? url
-    : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;
+  url =
+    url?.startsWith('http') || url?.startsWith('data:')
+      ? url
+      : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;
 
   return await fetchHttpImage(url);
 }
--- a/test/unit/resolvers/image/ens.test.ts
+++ b/test/unit/resolvers/image/ens.test.ts
@@ -109,7 +109,7 @@ describe('resolvers/image/ens', () => {
   it.each([
     'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII=',
     'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4='
-  ])('uses a data: avatar record directly', async record => {
+  ])('uses a data: avatar record directly %s', async record => {
     const image = Buffer.from('avatar');
     jest.spyOn(snapshot.utils, 'getEnsTextRecord').mockResolvedValue(record);
     mockedFetchHttpImage.mockResolvedValue(image);
@@ -121,7 +121,9 @@ describe('resolvers/image/ens', () => {
 
   it.each([
     'ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi',
-    'eip155:1/erc721:0xac5c7493036de60e63eb81c5e9a440b42f47ebf5/5413'
+    'eip155:1/erc721:0xac5c7493036de60e63eb81c5e9a440b42f47ebf5/5413',
+    'data:',
+    'data:image/png;base64'
   ])('falls back to the metadata service for a non-fetchable record %s', async record => {
     const image = Buffer.from('avatar');
     jest.spyOn(snapshot.utils, 'getEnsTextRecord').mockResolvedValue(record);
--- a/test/unit/resolvers/image/ens.test.ts
+++ b/test/unit/resolvers/image/ens.test.ts
@@ -106,6 +106,34 @@ describe('resolvers/image/ens', () => {
     expect(mockedFetchHttpImage).toHaveBeenCalledWith(url);
   });
 
+  it.each([
+    'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII=',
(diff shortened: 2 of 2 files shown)


wa0x6e (approved):
utAck

PR #629, opened 2026-08-25
fix(http): reject a reference getUrl cannot turn into a fetchable URL

### LINKED ISSUE #615: fix(http): getUrl builds an unfetchable gateway URL for a non-CID reference, so garbage input is reported as an upstream failure
Found during the review of #598. Related to #608, which describes the same mechanism from a security angle (no scheme/host allow-list); this issue is about a different symptom of it — reported failures — and has different acceptance criteria. Fixing either well would likely fix both.

## Today

`getUrl` (`src/helpers/http.ts`) delegates to `snapshot.utils.getUrl`, which rewrites `ipfs://` / `ipns://` / a bare CID onto the configured gateway and passes anything else through unchanged. It never signals that a value cannot produce a fetchable URL — it just builds one anyway. Probed directly against production's configured gateway (`IPFS_GATEWAY=snapshot.4everland.link`):

```
"https://example.com/a.png"      -> "https://example.com/a.png"                    (passthrough, fine)
"ipfs://QmYwAPJzv5…"             -> "https://snapshot.4everland.link/ipfs/QmYwAPJzv5…"   (fine)
"QmYwAPJzv5…"       (bare CID)   -> "https://snapshot.4everland.link/ipfs/QmYwAPJzv5…"   (fine)
"bafkreifnrjhkl7…"  (bare CIDv1) -> "https://snapshot.4everland.link/ipfs/bafkreifnrjhkl7…" (fine)
"logo.png"                       -> "https://snapshot.4everland.link/ipfs/logo.png"
"not a url at all"               -> "https://snapshot.4everland.link/ipfs/not a url at all"
```

The last two are syntactically valid URLs, so `fetch` proceeds and the gateway answers:

```
$ curl -sS -I "https://snapshot.4everland.link/ipfs/logo.png"
HTTP/2 400
content-type: text/plain; charset=utf-8
```

## Why that ends up in Sentry

`readHttpImage` (`src/helpers/http.ts`) turns the response into `httpError(host, 400, statusText)`. A 400 matches neither classifier:

- `isRoutineMiss` (`src/resolvers/image/index.ts`) matches only `404` / `INVALID_ARGUMENT`;
- `isSilencedError` (`src/helpers/errors.ts`) matches only TIMEOUT / ECONNABORTED / ETIMEDOUT / ECONNRESET / 504 / 429 and a fixed message list.

So `withFailureContract` reports it. The input is user-supplied garbage — a routine miss — but it is escalated as an upstream failure, which is the same class of noise #587 exists to remove.

Note the `statusText` is blank in the report, because the gateway answers over HTTP/2, which carries no reason phrase. That is why these show up as `Error: [snapshot.4everland.link] ` with an empty message and are hard to search for.

## Scope

Not Starknet-specific. Every caller that passes a user-controlled string through `getUrl` has the same shape — `starknet.ts` (profile picture, and the `image` field of its metadata), `ens.ts`, `lens.ts`, `selfid.ts`, `snapshot.ts`, `space-sx.ts`.

## What this is *not*

The `data:` cell of this problem is already handled: `getUrl` rewrote `data:` URIs onto the gateway path too, which is what drove STAMP-7E (51 events in 18h, `provider:starknet`). #610 fixed that with a `data:` short-circuit and was folded into #598, which now closes #609 as well. Measured on that input shape:

```
before      -> httpError(snapshot.4everland.link, 400) => Sentry capture
after #610  -> image returned (image/svg+xml)
```

What remains is every *other* unusable shape — relative paths, empty strings, arbitrary prose. Those have **no measured occurrences** in Sentry over 90 days, so this is a reachable gap rather than an active incident, and should be sized accordingly.

## Fix shape

The knowledge of what constitutes a fetchable reference (`http(s)`, `ipfs`, `ipns`, `data:`, bare CID) currently lives in `snapshot.utils.getUrl` and, after #610, partly in our own `getUrl`. It should stay in one place. Rejecting per-caller was considered during the #598 review and turned down for exactly that reason — a guard in `src/resolvers/image/starknet.ts` would have to re-encode the same scheme/CID list and would drift from `getUrl`.

Two candidate approaches:

1. Have `getUrl` return `null` for a value it cannot turn into a fetchable URL, and let callers treat `null` as no-data. Explicit, but changes a widely-used signature.
2. Classify the gateway's 400 as a routine miss. Cheaper and catches every caller at once, but it is a broader silence — a genuine gateway 400 on a *valid* CID would be swallowed too.

Worth deciding alongside #608, since a single validation point serves both.

## Acceptance criteria

- An unusable `image` / avatar reference resolves to no-data (fallback avatar) instead of a reported failure.
- A valid absolute URL, `ipfs://` URI, bare CIDv0/CIDv1 and `data:` URI all still resolve.
- A regression test covering each of those five shapes.

tony8713:
Closes #615.

`getUrl` (`src/helpers/http.ts`) delegates to `snapshot.utils.getUrl`, which rewrites `ipfs://`/`ipns://`/bare CIDs onto the configured gateway but never checks that the result is a URL `fetch()` can actually use. A bare scheme like `http://` or a URL with an unparseable port survives unchanged and reaches `fetch()`, which throws before any response exists. That throw carries no `.status`, so it isn't covered by `isRoutineMiss` or `isSilencedError`, and it gets captured/reported as an upstream failure instead of treated as the no-data case it is.

`getUrl` now validates the candidate URL and returns `null` when it can't produce a fetchable one. The four resolvers that call it directly (starknet, selfid, snapshot, space-sx) treat `null` the same way `basename.ts`'s avatar resolver already does: no-data rather than a failure.

A syntactically valid but meaningless gateway URL (e.g. a plain filename) still round-trips to the gateway and is already silenced once it 400s, by the existing routine-miss classification (`isRoutineMiss` now treats any 4xx except 401/402/403 as routine). This closes the other half, where the URL itself was never fetchable and `fetch()` never got a response to classify.

`data:` URI handling is out of scope here; it's a separate defect (#609) already being fixed in #598. This PR doesn't depend on #598 and doesn't touch the same lines as #625 (which bounds response body size, not URL validity), so either can land in any order relative to this one.

### DIFF
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
@@ -34,7 +34,14 @@ export function fetchHttpImage(url: string): Promise<Buffer> {
   return fetchWithDeadline(url, async response => Buffer.from(await response.arrayBuffer()));
 }
 
-export function getUrl(url) {
+export function getUrl(url: string): string | null {
   const gateway: string = process.env.IPFS_GATEWAY || 'cloudflare-ipfs.com';
-  return snapshot.utils.getUrl(url, gateway);
+  const candidate = snapshot.utils.getUrl(url, gateway);
+  if (!candidate) return null;
+
+  try {
+    return ['http:', 'https:'].includes(new URL(candidate).protocol) ? candidate : null;
+  } catch {
+    return null;
+  }
 }
--- a/src/resolvers/image/selfid.ts
+++ b/src/resolvers/image/selfid.ts
@@ -12,5 +12,7 @@ export default async function resolve(address: string) {
   if (!src) return false;
 
   const url = getUrl(src);
+  if (!url) return false;
+
   return await fetchHttpImage(url);
 }
--- a/src/resolvers/image/snapshot.ts
+++ b/src/resolvers/image/snapshot.ts
@@ -139,6 +139,8 @@ function createPropertyResolver(entity: Entity, property: Property) {
     if (!value) return false;
 
     const url = getUrl(value);
+    if (!url) return false;
+
     return await fetchHttpImage(url);
   };
 }
--- a/src/resolvers/image/space-sx.ts
+++ b/src/resolvers/image/space-sx.ts
@@ -44,7 +44,10 @@ function createPropertyResolver(property: 'avatar' | 'cover') {
     const value = answers.find(Boolean);
     if (!value) return false;
 
-    return await fetchHttpImage(getUrl(value));
+    const url = getUrl(value);
+    if (!url) return false;
+
+    return await fetchHttpImage(url);
   };
 }
 
--- a/src/resolvers/image/starknet.ts
+++ b/src/resolvers/image/starknet.ts
@@ -123,16 +123,24 @@ function fetchImageOrMetadata(url: string): Promise<Buffer | { image?: string }>
   );
 }
 
+async function fetchMetadataImage(image: string): Promise<Buffer | null> {
+  const url = getUrl(image);
+  return url ? fetchHttpImage(url) : null;
+}
+
 export default async function resolve(domainOrAddress: string) {
   const img_url = await getImage(domainOrAddress);
 
   if (!img_url || img_url === DEFAULT_IMG_URL) return false;
 
-  const fetched = await fetchImageOrMetadata(getUrl(img_url));
+  const url = getUrl(img_url);
+  if (!url) return false;
+
+  const fetched = await fetchImageOrMetadata(url);
   const buffer = Buffer.isBuffer(fetched)
     ? fetched
     : fetched.image
-      ? await fetchHttpImage(getUrl(fetched.image))
+      ? await fetchMetadataImage(fetched.image)
       : null;
 
   return buffer ?? false;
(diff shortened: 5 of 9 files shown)


tony8713:
The two Test failures (\`test/e2e/api.test.ts\` "returns different space cover for different network" and \`test/integration/resolvers/image/snapshot.test.ts\` "matches the image snapshot for snapshot-space-cover") are pre-existing and unrelated to this change: the fixture space \`test.wa0x6e.eth\` currently has no cover set on either network.

    curl -s https://hub.snapshot.org/graphql -d '{"query":"{ space(id:\"test.wa0x6e.eth\") { cover } }"}'
    -> {"data":{"space":{"cover":null}}}
    curl -s https://testnet.hub.snapshot.org/graphql -d '{"query":"{ space(id:\"test.wa0x6e.eth\") { cover } }"}'
    -> {"data":{"space":null}}

Same two failures reproduce identically on unmodified master (verified locally against \`HEAD~1\`, and separately via a throwaway-branch CI run on untouched master, 32826159238). Lint is green.

wa0x6e (approved):
utAck

Issue #606, opened 2026-08-23
fix(errors): viem timeouts escape isSilencedError, so our own 5s deadline reports as an error (STAMP-76)


wa0x6e:
Sentry: [STAMP-76](https://snapshot-labs.sentry.io/issues/STAMP-76) — 5 occurrences, first seen 2026-08-20 15:46:54 UTC, last 2026-08-22 02:14:16 UTC, `environment:production`, `handled:yes`, `provider:ens`, 0 users.

All the Sentry data needed is copied below; no Sentry access is required to fix this.

## What is reported

A three-level viem error chain, all frames third-party:

```
ContractFunctionExecutionError: The request took too long to respond.
  ← CallExecutionError: The request took too long to respond.
    ← TimeoutError: The request took too long to respond.

URL: https://rpc.brovider.xyz/1
Contract Call:
  address:   0xeEeEEEeE14D718C2B47D9923Deab1335E144EeEe
  function:  resolveWithGateways(bytes name, bytes data, string[] gateways)
Version: viem@2.55.13

viem/_cjs/actions/ens/getEnsText.js:50   (getEnsText)
viem/_cjs/actions/public/readContract.js:17
viem/_cjs/utils/rpc/http.js:50           new TimeoutError({ body, url })
```

Sample event: `GET https://cdn.stamp.fyi/avatar/0x7eb413211a9DE1cd2FE8b8Bb6055636c43F7d206`, resolver input `args: ["0x7eb413211a9de1cd2fe8b8bb6055636c43f7d206", "1", null]`.

The `name` argument decodes to **`n0b0dy.eth`** (`0x066e30623064790365746800` — DNS wire format: length 6 `n0b0dy`, length 3 `eth`, terminator), and the data argument is `text(node, "avatar")`. The contract is the ENS Universal Resolver.

## It is our own deadline, and the upstream is fine

`src/helpers/provider.ts:5` sets `DEFAULT_PROVIDER_TIMEOUT = 5e3`, and `resolvers/image/ens.ts:28` calls `getProviderOptions()` with no argument, so the viem client is given a 5 second budget. `TimeoutError` is viem reporting that budget expiring — not an upstream error response.

Measured 2026-08-23, replaying the exact `eth_call` from the event against `https://rpc.brovider.xyz/1`:

```
attempt 1: 0.586s http=200
attempt 2: 0.532s http=200
attempt 3: 0.538s http=200
```

[...]

PR #630, opened 2026-08-25
fix(errors): silence viem RPC timeouts and CCIP-Read aborts nested under ContractFunctionExecutionError

tony8713:
## Problem

Stamp imposes its own 5s deadline on viem RPC calls (`DEFAULT_PROVIDER_TIMEOUT`). When that budget is hit on a contract read (e.g. `getEnsTextRecord`, used by the ENS avatar resolver), viem wraps the resulting error three levels deep: `ContractFunctionExecutionError` → `CallExecutionError` → `TimeoutError`. `isSilencedError` only ever looked one level down (`error.cause?.code`/`.message`), so it never recognized this shape and reported our own deadline as a resolver defect (Sentry STAMP-76, issue #606).

A second, related path goes through the same caller: when the CCIP-Read gateway deadline fires instead of the RPC transport one, viem wraps a raw `DOMException` that has no `shortMessage`, so `ContractFunctionExecutionError`'s own message falls back to a generic one rather than inheriting the abort's message directly.

## Fix

viem's `BaseError` hierarchy always folds a wrapped cause's message into its own composed message (via `shortMessage` when present, or a `Details:` line otherwise), so the deepest failure's text reaches the top of the chain regardless of nesting depth or which of those two paths fired. Both fixed strings are added to `isSilencedError`'s existing message list, the same mechanism this function already uses for other known-safe-to-silence shapes (ethers `status=504`, DNS SERVFAIL, etc).

### DIFF
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
@@ -25,6 +25,7 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
     // visible as they may signal a real problem (e.g. FORMERR 1 = malformed query);
     // NXDOMAIN (3) never reaches here (dns-connect returns it as an empty result).
     'Received error status from DNS server: 2.',
+    'The request took too long to respond.',
     ...(additionalMessages || [])
   ];
   const codes = [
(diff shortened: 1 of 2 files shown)


wa0x6e (approved):
utAck

Issue #608, opened 2026-08-23
fix(image): user-supplied avatar URLs reach fetch with no scheme or host restriction


wa0x6e:
Found during review of #598. Not from Sentry — no occurrences observed; this is a reachable gap, not an active incident.

## Today

Image resolvers fetch a URL taken from user-controlled data, with no restriction on scheme or host. The clearest pair is in `src/resolvers/image/starknet.ts`:

- line 29 — `fetch(url)` on `getUrl(profilePicture)`, where `profilePicture` comes from the resolved Starknet profile, i.e. set by the address owner;
- line 49 — `fetchHttpImage(getUrl(fetched.image))`, where `image` comes from JSON that the *first* URL's host served.

`getUrl` (`src/helpers/http.ts:44`) delegates to `snapshot.utils.getUrl`, which only rewrites `ipfs://` / `ipns://` / a bare CID to the configured gateway and passes anything else through unchanged. Probed directly:

```
"https://example.com/nft.png"   => "https://example.com/nft.png"          (passthrough)
"ipfs://QmYwAP…"                => "https://<gateway>/ipfs/QmYwAP…"
"not a url at all"              => "https://<gateway>/ipfs/not a url at all"
```

No scheme allow-list, no host allow-list, no private/loopback/link-local check anywhere on the path.

## Blast radius

Not Starknet-specific. Every resolver that calls `fetchHttpImage` on a user-supplied URL has the same shape: `ens.ts:32` (avatar text record), `lens.ts:57` (`picture`), `selfid.ts:15` (image src), `farcaster.ts:58` (hub avatar), `snapshot.ts:146`, `space-sx.ts:49`, `coingecko.ts:41`, `trustwallet.ts:17`, `basename.ts:8`.

Fetching user-supplied URLs is the product, not the bug — that is what an avatar resolver does. The gap is that nothing distinguishes "an arbitrary host on the internet" from "something on our own network".

## Impact

Unmeasured, and the honest read is that it depends entirely on what is reachable from the stamp containers. The generic SSRF concerns are: requests originating from our network to internal addresses; cloud instance-metadata endpoints; and the response being *returned to the caller* as image bytes, which turns it into a read primitive rather than a blind request.

Mitigating factors worth weighing before sizing this: the response must pass the `image/*` media-type gate (added in #598) before its bytes are returned, so a JSON or HTML internal endpoint is rejected rather than echoed; and the 5s deadline bounds each attempt.

[...]

PR #631, opened 2026-08-25
fix(image): parse ENS and Farcaster avatar records as URLs before fetching

### LINKED ISSUE #605: fix(ens): avatar records are string-matched, not parsed, so malformed URLs reach fetch (STAMP-7C)
Sentry: [STAMP-7C](https://snapshot-labs.sentry.io/issues/STAMP-7C) — 1 occurrence, first and last seen 2026-08-23 03:47:04 UTC, `environment:production`, `handled:yes`, `provider:ens`, 0 users.

## What is reported

```
TypeError: fetch failed
  caused by: Error: bad port
    at fetch (node:internal/deps/undici/undici:13728:20)
    at makeNetworkError (node:internal/deps/undici/undici:12443:35)
```

Ours are the first two frames:

```
src/resolvers/image/ens.ts:33   return await fetchHttpImage(url)
src/helpers/http.ts:35          const response = await fetch(url, { signal })
```

Sample event: `GET https://cdn.stamp.fyi/avatar/eth:0x60c63bbb27f1f49f63eacadc773c79bf0c17ac7a`, resolver input `args: ["0x60c63bbb27f1f49f63eacadc773c79bf0c17ac7a", "1", "eth"]`, client `ImageBot 1.0`.

## Root cause

That address is `vishni.eth`. Its `avatar` text record, read from the public resolver `0x4976fb03C32e5B8cfe2b6cCB31c09Ba78EBaBa41` on mainnet (verified 2026-08-23):

```
http://eip155:1/erc721:https://etherscan.io/address/0xac5c7493036de60e63eb81c5e9a440b42f47ebf5/5413
```

Someone pasted an Etherscan link into what should be a CAIP identifier (`eip155:1/erc721:<contract>/<tokenId>`) and prefixed it with `http://`. The record is malformed and unresolvable by anyone.

`src/resolvers/image/ens.ts:29-31` decides what to do with it:

```ts
url = url?.startsWith('http')
  ? url
  : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;
```

The gate is a **string prefix test, not a URL check**. This record starts with `http`, so it is taken as a fetchable URL and passed straight to `fetchHttpImage`. Parsed, it is not one:

```
new URL(record) → protocol=http:  hostname=eip155  port=1
```

Port 1 (tcpmux) is on the WHATWG fetch bad-ports blocklist, so undici rejects the request before opening a socket and throws `TypeError: fetch failed` with `cause.message = 'bad port'` and no `cause.code`. Reproduced directly with `fetch()` on that record, same error.

The same prefix test also accepts anything else beginning with those four characters — `httpfoo://…`, `http-x`.

## Why nothing silences it

`isRoutineMiss` (`src/resolvers/image/index.ts:33`) keys on `status` / `response.status` / `code`; this error has none of them. `isSilencedError` (`src/helpers/errors.ts:10`) matches messages against a fixed list and codes against `['TIMEOUT','ECONNABORTED','ETIMEDOUT','ECONNRESET',504,429]`; `'fetch failed'` / `'bad port'` are in neither, and every code slot is `undefined`.

PR #595 does not change this: it keys on `Number(status)` (`NaN` here) and `cause?.code ?? code` (`undefined` here), so this case still reports after that merges.

**But the classifier is the wrong place to fix it.** A record that cannot be parsed into a fetchable URL is not an upstream failure to be forgiven — it is input we should never have fetched.

## Suggested fix

Validate the record before choosing the path, rather than string-matching it. `zod` is already a dependency (`^3.25.76`) and `src/helpers/validation.ts` already exists from #468:

```ts
const httpUrl = z.string().url().refine(u => /^https?:$/.test(new URL(u).protocol));
url = httpUrl.safeParse(record).success
  ? record
  : `https://metadata.ens.domains/mainnet/avatar/${encodeURIComponent(ensName)}`;
```

Worth noting what the fallback then does here: `https://metadata.ens.domains/mainnet/avatar/vishni.eth` returns **HTTP 404** (checked 2026-08-23) — the ENS metadata service cannot resolve this record either. That 404 is already a routine miss, so the corrected path ends silently through existing code, with no classifier change and no new mute. The user still gets a blockie, which is the correct outcome for a record no one can resolve.

A stricter version could parse the CAIP form properly and hand it to the metadata service, but that is a separate feature; this issue is only about not fetching a URL we have not checked.

## Impact

None per request. `withFailureContract` returns `false` and `api.ts` falls through to the next resolver, then to a blockie. The avatar endpoint returns 200.

Volume is one event from one address in three days, against 63,880 stamp requests in the last 24h (Prometheus, `http_request_duration_seconds_count{instance=~"stamp.fyi:.*"}`), so this is noise-level today. It is filed as a class rather than an incident: any ENS avatar record that starts with `http` but is not a valid URL takes this path, and each new one arrives as its own Sentry id because the host differs.

## Provenance

Not a regression. The `startsWith('http')` guard predates the current file layout — it was carried through `95c4fe7` (`refactor: group resolver types`) and again through `e7a827a` (`refactor(ens): read avatar through universal resolver`, #569) without changing shape. Any malformed-but-http-prefixed record has always taken this path.

Related: #567 establishes that each resolver validates the shape of what it can serve — that issue covers addresses and handles; this is the same principle applied to a URL read from an external record. #595 widens the error classifier for genuinely unreachable third-party hosts, which is the neighbouring case, not this one.

Sweep: checked the 11 unresolved stamp issues; STAMP-7C is the only one with this cause. STAMP-77 and STAMP-78 look similar (`provider:space`, third-party URL) but are reachable hosts failing at the network layer, and are raised on #595.

tony8713:
Closes #605.

`ens.ts` accepted any avatar text record starting with the string `http` as a fetchable URL. A malformed record (a CAIP identifier wrongly prefixed with `http://`) reached `fetch()` unparsed: the resulting throw was already caught by the resolver's existing failure-contract wrapper, but wasn't recognized as a routine miss, so it was reported to Sentry as noise on every occurrence, and the resolver never got to try its real ENS-metadata fallback for that record. `farcaster.ts` had no check at all: a relative `pfp_url` from Neynar hit `fetch()` the same way.

Both now go through a shared `isHttpUrl` helper in `helpers/http.ts`. It parses the value as a URL, requires an http/https protocol, and rejects fetch's own blocked-port list, since a record can be syntactically a valid URL and still be something fetch will never attempt (the reported crash was exactly this: a well-formed `http://` URL on a blocked port, which a plain URL-parse-and-protocol check alone would not have caught). Anything that fails validation now falls through to each resolver's existing no-avatar path instead of reaching fetch.


src/helpers/http.ts
```diff
@@ -61,6 +61,31 @@ export function fetchHttpImage(url: string): Promise<Buffer> {
...
+    return false;
+  }
+
+  return (
+    (url.protocol === 'http:' || url.protocol === 'https:') &&
+    (url.port === '' || !FORBIDDEN_PORTS.has(Number(url.port)))
+  );
+}
+
 export function getUrl(url: string): string | null {
```
wa0x6e, line 75: `isHttpUrl` (added just above) checks protocol **and** `FORBIDDEN_PORTS`. `getUrl`, five lines below, checks only protocol via its own inline `['http:', 'https:'].includes(...)` — it never consults the port list. So the crash class this PR closes for ens/farcaster stays open at every `getUrl` consumer: `snapshot.ts:141-144`, `space-sx.ts:47-50`, `starknet.ts:127-128` (and `:136`, which reaches `fetch` via `fetchImageOrMetadata`/`fetchWithDeadline`), and `address/basename.ts:102` → `image/basename.ts:8`.

Measured on Node v26.7.0: `snapshot.utils.getUrl('http://example.com:25/avatar.png', 'cloudflare-ipfs.com')` returns the input unchanged, `getUrl` accepts it, and `fetch()` then rejects with `TypeError: fetch failed`, `cause.message = 'bad port'`, **no** `cause.code`. `isRoutineMiss` (`resolvers/image/index.ts:50-59`) reads `cause.code ?? code` → `undefined` and `Number(status)` → `NaN`, so neither branch fires and `withFailureContract` captures it to Sentry — the exact noise this PR exists to remove.

Second effect of the same shape: the file now holds two definitions of "is this an http URL", and only one of them knows about ports. Folding `getUrl` into the new helper removes the divergence and covers the four call sites in one line:

```diff
 export function getUrl(url: string): string | null {
   const gateway: string = process.env.IPFS_GATEWAY || 'cloudflare-ipfs.com';
   const candidate = snapshot.utils.getUrl(url, gateway);
   if (!candidate) return null;

-  try {
-    return ['http:', 'https:'].includes(new URL(candidate).protocol) ? candidate : null;
-  } catch {
-    return null;
-  }
+  return isHttpUrl(candidate) ? candidate : null;
 }
```

`test/unit/helpers/http.test.ts` stays green with this applied (16/16, including all 7 existing `getUrl` cases). Worth a `getUrl` regression case on a blocked port too — the current suite only exercises blocked ports through `isHttpUrl`, which is why the divergence isn't caught.

No conflict with #632: that PR guards inside `fetchWithDeadline` and never touches `getUrl`'s body, and it carries no port check of its own, so this is still needed after it merges.

src/resolvers/image/ens.ts
```diff
@@ -26,7 +26,7 @@ export default async function resolve(nameOrAddress: string) {
 
   let url = await snapshot.utils.getEnsTextRecord(ensName, 'avatar', '1', getProviderOptions());
   url =
-    url?.startsWith('http') || (url && /^data:[^,]*,/.test(url))
+    url && (isHttpUrl(url) || /^data:[^,]*,/.test(url))
```
wa0x6e, line 29: This gate diverts the `http://eip155:<chainId>/erc721:...` record class — the class the PR exists to fix — only when the chain id happens to collide with a WHATWG blocked port. Chain id `1` (the reported record) is caught for that reason alone. Base (`:8453`), Optimism (`:10`) and Arbitrum (`:42161`) all pass `isHttpUrl`, reach `fetch()`, and throw `TypeError: fetch failed` with `cause.code = 'ENOTFOUND'` on hostname `eip155`.

**See the comment on `helpers/http.ts` for the suggested fix** — validating the hostname shape inside `isHttpUrl` covers this whole class for every resolver at once, so this line stays exactly as you wrote it. Leaving the evidence here since it's specific to this record class:

Impact is smaller than it first looks, and worth recording so nobody re-raises it as a user-facing bug: `metadata.ens.domains` 404s for this whole record shape regardless of chain id. It runs `@ensdomains/ens-avatar`'s `parseNFT`, which does `uri.split('/')` — the `//` after the scheme always inserts an empty element, so the destructured `reference` is always `"http:"` rather than `"eip155:<chainId>"`, and it throws *"Only EIP-155 is supported"*. Checked live: `GET https://metadata.ens.domains/mainnet/avatar/vishni.eth` → 404 `{"message":"Error fetching avatar: Provided url or NFT source is broken."}`. `ENOTFOUND` is also already in `ROUTINE_NETWORK_ERROR_CODES` (`resolvers/image/index.ts:34`), so the unlucky path is silent in Sentry as well. Both branches end in a blockie, quietly.

So this is a robustness fix — correctness shouldn't depend on an unrelated port table — rather than a lost-avatar bug. Worth a regression case on a non-blocked chain id either way:

```ts
[
  'a CAIP identifier prefixed with http, on a chain id that is not a blocked port',
  'http://eip155:8453/erc721:0xac5c7493036de60e63eb81c5e9a440b42f47ebf5/5413'
]
```

src/resolvers/image/farcaster.ts
```diff
@@ -49,7 +49,7 @@ export default async function resolve(address: string): Promise<Buffer | false>
   if (!normalizedAddress) return false;
 
   const url = await fetchAddressImageUrl(normalizedAddress);
-  if (!url) return false;
+  if (!url || !isHttpUrl(url)) return false;
 
   return await fetchHttpImage(withCache(url));
```
wa0x6e, line 54: Nice fix on this line — worth sweeping it to the two siblings with the same shape, since they're the only other resolvers that hand a raw third-party string to `fetchHttpImage`:

- `src/resolvers/image/lens.ts:59-62` — `normalizeImageUrl` only rewrites the `gw.ipfs-lens.dev` prefix and passes everything else through unchanged, so `account.metadata.picture` reaches `fetchHttpImage` unparsed. Lens's published Account Metadata schema types `picture` as an unconstrained `URI`, not http(s)-only.
- `src/resolvers/image/coingecko.ts:39-41` — `data.image.large` gets only a truthiness check, and nothing in the repo constrains it.

A relative or `ipfs://`/`lens://` value throws `TypeError: Failed to parse URL` (`cause.code = 'ERR_INVALID_URL'`) or `TypeError: fetch failed` (`cause: 'unknown scheme'`, no `code`). Neither is in `ROUTINE_NETWORK_ERROR_CODES`, so both land in Sentry unclassified — the same thing this PR removes for farcaster.

```diff
--- a/src/resolvers/image/lens.ts
+++ b/src/resolvers/image/lens.ts
-import { fetchHttpImage } from '../../helpers/http';
+import { fetchHttpImage, isHttpUrl } from '../../helpers/http';
@@
   const img_url = normalizeImageUrl(account?.metadata?.picture);
-  if (!img_url) return false;
+  if (!img_url || !isHttpUrl(img_url)) return false;

--- a/src/resolvers/image/coingecko.ts
+++ b/src/resolvers/image/coingecko.ts
-import { fetchHttpImage } from '../../helpers/http';
+import { fetchHttpImage, isHttpUrl } from '../../helpers/http';
@@
-  if (!data?.image?.large) return false;
+  if (!data?.image?.large || !isHttpUrl(data.image.large)) return false;
```

Calibration, so this doesn't read as more urgent than it is: sampling ~500 live CoinGecko coins and ~75 live Lens accounts turned up **zero** non-http values, which matches a 90-day Sentry sweep finding no `provider:lens` or `provider:coingecko` events of this class. The `missing_large.png` relative placeholder that used to make CoinGecko the likely trigger appears deprecated — it returns `null` now, which the existing check already handles. So this is closing a contract gap cheaply while the pattern is fresh, not fixing an active bug; a follow-up issue would be a fair call instead.

The guard belongs at these two call sites rather than inside `fetchHttpImage` — `trustwallet.ts` passes internally-constructed URLs and should keep throwing on those.

Note `#608` names both files, but it's the destination/SSRF axis (which hosts we're allowed to reach), not this parsing axis, so it doesn't cover this.

src/helpers/http.ts
```diff
@@ -61,6 +61,31 @@ export function fetchHttpImage(url: string): Promise<Buffer> {
...
+  } catch {
+    return false;
+  }
+
+  return (
+    (url.protocol === 'http:' || url.protocol === 'https:') &&
+    (url.port === '' || !FORBIDDEN_PORTS.has(Number(url.port)))
+  );
+}
+
```
wa0x6e: Suggest dropping this table and splitting the concern along a different axis: **validate the hostname before fetching, and let a blocked port be handled at the error level.**

The reasoning is that the reported record's defect was never its port. `http://eip155:1/erc721:...` is broken because `eip155` was never a hostname — the port check catches it only because chain id `1` happens to sit on the WHATWG list. The same record on Base, Optimism or Arbitrum (`:8453`, `:10`, `:42161`) is not a blocked port and sails straight through to `fetch()`. A hostname check catches the whole shape regardless of which chain the record names.

Conversely a blocked port is fetch's own internal rule, not a defect in the URL we were handed. It moves when undici moves, and there is no way to know it except hand-copying a spec list — which is a maintenance burden this table has already demonstrated: `9225879` had to add the missing `4190` and `6679`, found only by brute-forcing the port range against a live `fetch()`. `isSilencedError` already inspects `error.cause?.message`, and `'bad port'` is exactly where undici puts it, so one entry there covers the case permanently and stays correct as the spec evolves.

```diff
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
-// fetch's own port blocklist (https://fetch.spec.whatwg.org/#port-blocking): ...
-const FORBIDDEN_PORTS = new Set([ /* 82 entries */ ]);
-
 export function isHttpUrl(value: string): boolean {
   ...
-  return (
-    (url.protocol === 'http:' || url.protocol === 'https:') &&
-    (url.port === '' || !FORBIDDEN_PORTS.has(Number(url.port)))
-  );
+  return (url.protocol === 'http:' || url.protocol === 'https:') && url.hostname.includes('.');
 }
```

```diff
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
     'This operation was aborted',
+    // undici refuses to open a socket on a port the fetch spec blocks
+    // (https://fetch.spec.whatwg.org/#port-blocking) and reports it only here,
+    // as `cause.message`, with no status or code. isHttpUrl validates protocol
+    // and hostname shape, not port — a blocked port is fetch's own rule, not a
+    // URL-shape defect.
+    'bad port',
```

Test churn that goes with it: the `rejects every port fetch() itself refuses` case can go (it asserts the constant against a copy of itself, so it could never catch drift anyway), as can the `'http://example.com:25/avatar.png'` rejection case. Worth adding in their place: the CAIP shape on chain ids `8453`/`10`/`42161`, a few real avatar hosts (`ik.imagekit.io`, `gw.ipfs-lens.dev`, `wrpcd.net`, `metadata.ens.domains`, `assets.coingecko.com`) as acceptance cases, and an `isSilencedError` case built from the **real** rejection `fetch('http://example.com:25/x')` produces rather than a hand-built stand-in.

Checked locally: with this applied, `test/unit/helpers/http.test.ts`, `test/unit/helpers/errors.test.ts`, `test/unit/resolvers/image/ens.test.ts` and `test/unit/resolvers/image/noData.test.ts` are 88/88 green, and the existing "falls back to the metadata service rather than fetching it" case for the reported record still passes — so the motivating record is still never fetched.

Two things to be aware of rather than fix here:
- `url.hostname.includes('.')` accepts `192.168.1.1` (dotted) but rejects `[::1]` (no dot), so it is not consistent SSRF fencing. That is fine — #632 is adding address-level checks at the fetch seam, which is the right layer for it. This check is only about URL shape. It does incidentally reject single-label hosts like `localhost` and `internal-svc`, which is a welcome side effect.
- The suggested `isSilencedError` test calls a real `fetch()`. It is offline-safe (fetch refuses a blocked port before opening a socket, in ~0ms) but it is a live call in a unit test, so flagging it in case you'd rather it live in `test/integration/helpers/`.

If you take this, the separate comment I left on `resolvers/image/ens.ts:29` becomes unnecessary — the host check moves into the helper, so that line stays exactly as you wrote it.

wa0x6e (changes requested):

src/helpers/http.ts
tony8713: Folded `getUrl` into `isHttpUrl` in d53b9f7, so it delegates instead of carrying its own inline protocol check. All four consumers (`snapshot.ts`, `space-sx.ts`, `starknet.ts`, `address/basename.ts`) now get the same hostname validation as ens/farcaster. Added a `getUrl` regression case for the CAIP record, mutation-tested: reverting the fold turns it red and nothing else.

src/resolvers/image/ens.ts
tony8713: Took the hostname-check suggestion from the helpers/http.ts thread, so this line stays exactly as written, as you noted. The CAIP class is now rejected regardless of chain id (added regression cases for Base/Optimism/Arbitrum in both http.test.ts and ens.test.ts, mutation-confirmed against reverting the hostname check).

src/resolvers/image/farcaster.ts
tony8713: Added the same isHttpUrl guard to lens.ts and coingecko.ts in d53b9f7, plus a noData.test.ts case for each (mocked picture/image.large set to a non-http value, asserting fetchHttpImage is never called). Mutation-confirmed: removing either guard turns exactly that one new test red.

src/helpers/http.ts
tony8713: Took this in d53b9f7: dropped FORBIDDEN_PORTS, isHttpUrl now checks protocol + `hostname.includes('.')`, and added 'bad port' to isSilencedError's message list.

Test churn matches what you suggested: removed the self-referential port-table test and the now-wrong 'blocked port' rejection case, added acceptance cases for the five real avatar hosts, rejection cases for the CAIP shape on chain ids 1/8453/10/42161 (isHttpUrl and ens.test.ts both), and an isSilencedError case in test/integration/helpers/errors.test.ts built from a real fetch('http://example.com:25/x') rejection, since it's a genuine fetch() call even though it never opens a socket (cause.message is 'bad port' on this box's Node 22 too, no code, matching what you measured on v26.7.0).

Mutation-confirmed each piece separately: reverting the hostname check kills 7 tests across http.test.ts and ens.test.ts (including the Base/Optimism/Arbitrum cases you called out), reverting the isSilencedError entry kills exactly the new integration test, and reverting the getUrl fold kills exactly its own regression case. Also ran the full findRelatedTests set for helpers/{http,errors}.ts plus the four resolvers (46 suites): only the 3 pre-existing key-dependent lookupDomains failures (missing D3_API_KEY_MAINNET/UNSTOPPABLE_DOMAINS_API_KEY on this box), nothing new.

wa0x6e (approved):
utAck

PR #632, opened 2026-08-25
fix(image): restrict avatar fetches to public http(s) hosts

### LINKED ISSUE #608: fix(image): user-supplied avatar URLs reach fetch with no scheme or host restriction

tony8713:
Every image resolver eventually calls `fetchHttpImage`/`fetchWithDeadline` in `helpers/http.ts` on a URL taken from user-controlled data (an ENS avatar record, a Farcaster pfp, a Starknet profile, ...), with no restriction on scheme or destination. `getUrl`'s IPFS/IPNS rewrite passes anything else through unchanged, so a URL pointing at a loopback address, a cloud metadata endpoint, or an internal service reaches `fetch` exactly like a public avatar host would, and the response comes back to the caller as image bytes.

`fetchWithDeadline` now rejects a URL before it reaches the network: only `http:`/`https:` are allowed. The destination address is checked in the dispatcher's own `connect` function rather than against the URL string, so it applies uniformly to the original request and to every redirect hop `undici` follows internally, a literal IP is checked directly and a hostname is checked against the address it actually resolves to.

The check lives once in `helpers/http.ts`, so it covers every resolver that calls `fetchHttpImage`, not just Starknet. Switching the underlying client to `undici`'s own `fetch` does not change the error shapes the resolvers' classifiers (`isRoutineMiss`, `isSilencedError`) depend on.

Closes #608


tony8713:
@wa0x6e on the new deps: `ipaddr.js` isn't actually new, it's already pinned in the lockfile at `1.9.1` via `proxy-addr` (express's own dependency), so declaring it direct here adds nothing to `node_modules`. `undici` is the real addition, but it's zero-dependency (~2MB) and it's literally what Node bundles internally to implement its own global `fetch` (confirmed `require('node:undici')` still isn't exposed as a builtin, even on Node 22.23 here). It's needed because there's no way to hook DNS resolution on a `fetch()` call without a custom `Agent`, and you can't construct one without importing `undici` directly.

On https-only being strict enough: it isn't, and today's code doesn't lean on scheme either way (both `http:`/`https:` stay allowed). TLS checks the hostname against the cert, not the destination IP, so an attacker with a normal cert for a domain they control just points DNS at `127.0.0.1` or `169.254.169.254` (which itself talks plain HTTP anyway) and https buys nothing. Proved it by disabling only the address checks here (left the scheme allow-list untouched) and rerunning: `rejects a literal loopback address` and `rejects a hostname that resolves to loopback` both went red, the request actually reached the local test server and got the image back, while the scheme-rejection test and the `isPublicAddress` unit tests stayed green. Scheme filtering blocks a different class of thing (`file://`, `gopher://`); it doesn't touch this one.

Also, re: whether #629 already covers this, different layer. #629 makes `getUrl` return `null` for a string that can never become a fetchable URL (bare scheme, unparseable port); it never looks at the destination. Checked directly on #629's branch: `getUrl('http://127.0.0.1/x')` returns the string unchanged, no block. #632 sits one layer down in `fetchWithDeadline`/`fetchHttpImage`, which is what actually catches that case, and it covers every caller of `fetchHttpImage`, including the ones that don't go through `getUrl` at all. Merged both branches locally to check for conflicts: `http.ts` and `starknet.ts` auto-merge clean since they touch different functions, the only conflict is both adding a new test file at the same path.

wa0x6e:
Found while reviewing #631 (the two PRs touch `helpers/http.ts` from different directions) — flagging here rather than filing, since you haven't requested review yet.

**A redirect to a bare IP literal bypasses the guard entirely.**

`safeLookup` validates addresses at DNS-resolution time, which is the right place and correctly defeats DNS-rebinding TOCTOU. But Node skips DNS resolution altogether when the host is already an IP literal, so the custom `lookup` never fires for those. `assertAllowedUrl` covers that case for the *original* URL — it just never sees a redirect target, because undici follows redirects internally and only the first URL reaches the assertion.

Reproduced end to end against a local server: `302` → `http://127.0.0.1:<port>/` → `200` with the internal response body returned to the caller, and **0** calls to the `lookup` hook.

So the current shape covers:
- IP literal in the original URL → `assertAllowedUrl` ✅
- hostname resolving to private space, original URL → `safeLookup` ✅
- redirect to a *hostname* resolving to private space → `safeLookup` ✅ (fires per-connection)
- redirect to a bare **IP literal** → nothing fires ❌

The fix likely belongs in the same per-connection place the rest of the guard already lives, so redirect hops inherit it rather than needing `assertAllowedUrl` to be re-run per hop — but you know this code better than I do. A regression test driving a 302 to `127.0.0.1` would pin whichever shape you pick.

Two smaller notes from the same read, both non-blocking:

- The `mergeable: CONFLICTING` status on this PR is a `yarn.lock` collision with the already-merged #637 (node engine bump), not anything to do with #631. I checked `git merge-tree` against #631 directly: `src/helpers/http.ts` auto-merges cleanly — the two PRs write to disjoint regions. Only `test/unit/helpers/http.test.ts` conflicts, trivially (same import line, adjacent `describe`), for whichever merges second.
- I verified undici's `fetch` produces byte-identical error shapes to global `fetch` for the cases the resolvers' classifiers depend on — `cause.message === 'bad port'` with no `cause.code`, and `cause.code === 'ENOTFOUND'`. So switching the module to undici doesn't silently change `isRoutineMiss`/`isSilencedError` behaviour. Worth a line in the PR description, since it's the kind of thing a reviewer would reasonably worry about.

PR #633, opened 2026-08-25
chore(provider): default brovider URL to rpc.brovider.xyz

### LINKED ISSUE #583: chore(provider): default brovider URL to rpc.brovider.xyz
Align with the sx-monorepo convention (sx-monorepo#2106) and snapshot.js#1239: the org's standard brovider host is `https://rpc.brovider.xyz`; `rpc.snapshot.org` remains only as legacy traffic.

## Change

Three occurrences:

- `src/helpers/provider.ts:5` — fallback: `process.env.BROVIDER_URL || 'https://rpc.brovider.xyz'`
- `.env.example:7` — `BROVIDER_URL=https://rpc.brovider.xyz`
- `test/.env.test:2` — same

Check the deploy config's `BROVIDER_URL` too: if it explicitly pins `rpc.snapshot.org`, update it in the same change, otherwise the code fallback never applies.

## Impact

Hostname swap only; provider behaviour unchanged. After stamp #581 (clientName) both land in brovider metrics on one vhost.

Related: snapshot.js#1239 (SDK default), #581 (attribution).

tony8713:
## Problem

The org's standard brovider host is `rpc.brovider.xyz` (sx-monorepo#2106, snapshot.js#1239). Stamp's own fallback still defaults to the legacy `rpc.snapshot.org`, so when `BROVIDER_URL` is unset, stamp's RPC traffic doesn't land on the current vhost.

## Fix

Swap the hardcoded fallback from `rpc.snapshot.org` to `rpc.brovider.xyz` in the three places it lives: the code default in `src/helpers/provider.ts`, and the two env templates (`.env.example`, `test/.env.test`). Anyone with `BROVIDER_URL` already set explicitly is unaffected.

Note: if a deploy config outside this repo pins `BROVIDER_URL` to `rpc.snapshot.org` explicitly, that needs updating separately, otherwise this fallback never applies there. Not something I can check from this repo.

Closes #583


wa0x6e (approved):
utAck

wa0x6e (approved):
utAck

PR #636, opened 2026-08-28
chore(resolvers): remove the dead Self.ID resolver

### LINKED ISSUE #634: chore(resolvers): remove the dead Self.ID resolver
Found during a code audit of the image resolvers. **Not from Sentry** — no occurrences observed, and none are possible: the resolver is unreachable from the HTTP surface. Dead code, not an active incident.

## Today

`selfid` is registered as an image resolver but cannot be selected:

- `src/constants.json` `resolvers.avatar` is `["snapshot", "ens", "basename", "lens", "farcaster", "starknet"]`. `selfid` was dropped in 303e107 (2023-01-30, "hotfix: remove selfid resolver").
- `src/api.ts:122` picks the chain from that list, and `src/api.ts:131-137` rejects an explicit `?resolver=` that is not in it (`500 invalid resolvers`).
- The `fallback` param is the only other way into the resolver map, and `src/helpers/api.ts:45` clamps it to `jazzicon` / `blockie`.

So the entry at `src/resolvers/image/index.ts:109` is its only reference, and no request reaches `src/resolvers/image/selfid.ts`.

Were it reachable, every call would fail. `src/resolvers/image/selfid.ts:5` hardcodes `https://gateway.ceramic.network`, whose host no longer resolves (checked 2026-08-28):

```
$ dig +short gateway.ceramic.network @1.1.1.1
ceramic-prod-ex-gateway-500821070.us-east-2.elb.amazonaws.com.

$ dig +short A ceramic-prod-ex-gateway-500821070.us-east-2.elb.amazonaws.com @1.1.1.1
$ dig +short A ceramic-prod-ex-gateway-500821070.us-east-2.elb.amazonaws.com @8.8.8.8
                                            (empty on both resolvers)

$ curl https://gateway.ceramic.network/api/v0/node/healthcheck
curl: (6) Could not resolve host: gateway.ceramic.network
```

The CNAME survives; the load balancer it points at has no A record on either resolver — it was deleted. Positive control: `hub.snapshot.org` resolves from the same queries. `ceramic-clay.3boxlabs.com`, the old Clay node, is gone as well.

`ENOTFOUND` is listed in `ROUTINE_NETWORK_ERROR_CODES` (`src/resolvers/image/index.ts:34`), so even a reachable-but-failing call would be silenced and return `false`. That is why nothing has ever flagged this.

The client is abandoned too: `@self.id/core` last published 0.4.1 on 2022-05-26. We pin `^0.3.0` and have 0.3.0 installed.

## Impact

None for users. The resolver has not been in the avatar chain since 2023-01-30 and cannot be reached today. The cost sits elsewhere:

- `@self.id/core` drags the `@ceramicnetwork/*` tree in — 1.4M installed (`@ceramicnetwork` 1.3M, `@self.id` 76K) — into the image and the dependency/audit surface, for code that cannot run.
- The registry entry, unit test, skipped integration test, fixtures and README section are carried through every resolver-wide refactor. #511's registry work touches every entry in `RESOLVERS`, this one included.

## Provenance

Not a regression. Added in #18 (372ec22, 2022-06-12), removed from the avatar chain in 303e107 (2023-01-30), unit test neutered in e88a519 (2023-06-17), integration test filed as `skip: true` with the todo "resolves and matches the reference avatar (Ceramic gateway deprecated)" in #456 (49afb2b, 2026-06-10).

## What to remove

- `src/resolvers/image/selfid.ts`
- its import and `RESOLVERS` entry in `src/resolvers/image/index.ts`
- `test/unit/resolvers/image/selfid.test.ts` and `test/integration/resolvers/image/selfid.test.ts`
- `selfidMissingDid` / `selfidNoAvatar` in `test/fixtures/image-snapshot-addresses.ts`, and `'selfid'` in `test/unit/resolvers/image.test.ts`
- the Self.ID section in `README.md`
- `@self.id/core` from `package.json` and `yarn.lock`

`src/constants.json` needs no change — `selfid` is already absent.

## Sweep

Checked open and closed stamp issues for `selfid`, `ceramic` and `self.id`, and open PRs touching `src/resolvers/image/selfid.ts`: no issue proposes removing it, no PR touches it. #511 (provider registries) is adjacent — it restructures `RESOLVERS` but removes no provider.

wa0x6e:
Closes #634

`selfid` was registered as an image resolver but could not be selected. It has not been in any chain in `constants.json` since 2023-01-30, and an explicit `?resolver=` is rejected unless it is in the chain, so no request reached it. Its backend is gone too — `gateway.ceramic.network`'s CNAME target has no A record on either public resolver, so a call would fail with `ENOTFOUND`, which the registry classifies as a routine miss and silences. That is why nothing ever flagged it.

No user-visible change. Removing it drops `@self.id/core` and the `@ceramicnetwork` tree it pulled in from the image and the dependency surface, and takes the entry out of the registry that resolver-wide refactors keep having to carry.

Left alone: `parseQuery` still strips a `did:` prefix from the requested id. That path predates this branch and is not tied to the resolver map, so it stays for a separate look.


tony8713 (approved):
Approve, clean self-contained dead-code removal. Checked constants.json history: selfid was already dropped from resolvers.avatar in a 2023-01-30 hotfix, and current master has no resolver list referencing selfid anywhere. src/api.ts also rejects any explicit ?resolver= value that isn't in the applicable list, so this path could never be hit that way either. Grepped the repo and nothing else references it. Pure subtraction of an already fully-unreachable path, no adversarial pass needed.

PR #637, opened 2026-08-28
chore: bump node engine requirement to >=24

tony8713:
## What

Bump stamp's Node floor from 22 to 24: `engines.node` (`>=22.6.0` -> `>=24`), `@types/node` (`^14.14.21` -> `^24`), the resulting `yarn.lock` update, and the CI runner's `node-version`.

## Why

Node 22 left Active LTS in October 2025 and is now in Maintenance (EOL April 2027). Node 24 is Active LTS until October 2026, then Maintenance until April 2028. Same kind of sweep the org already did for Node 22 across several sibling services (see e.g. laser#76 and its table of prior PRs).

## Testing

Verified locally against a real Node 24.20.0 runtime (not just the version bump): `yarn install --frozen-lockfile` succeeds with the updated lockfile, `yarn typecheck` and `yarn lint` are clean against the new `@types/node`, and a full `yarn test:coverage` run's only red tests are the two live-API integration files that are already red here under Node 22 too (missing `D3_API_KEY_MAINNET` / `UNSTOPPABLE_DOMAINS_API_KEY` locally; both are injected in CI). So the runtime bump itself introduces nothing new.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,5 +4,5 @@ jobs:
   lint:
     uses: snapshot-labs/actions/.github/workflows/lint.yml@main
     with:
-      target: '22'
+      target: '24'
     secrets: inherit
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '22'
+          node-version: '24'
           cache: 'yarn'
 
       - name: Setup Redis
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "license": "MIT",
   "engines": {
-    "node": ">=22.6.0"
+    "node": ">=24"
   },
   "scripts": {
     "generate-chains": "node ./dist/src/scripts/generate-chains.js",
@@ -57,7 +57,7 @@
     "@types/express": "^4.17.11",
     "@types/jest": "^28.1.0",
     "@types/jest-image-snapshot": "^6.4.1",
-    "@types/node": "^14.14.21",
+    "@types/node": "^24",
     "@types/supertest": "^6.0.3",
     "eslint": "^9.0.0",
     "jest": "^28.1.0",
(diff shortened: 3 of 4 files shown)


tony8713:
Confirming the expected red, so nobody has to dig: both [`Test`](https://github.com/snapshot-labs/stamp/actions/runs/33174748338/job/98860340638) and [`Lint`](https://github.com/snapshot-labs/stamp/actions/runs/33174748622/job/98860341771) fail at the same step, `yarn install --frozen-lockfile`, on the engines check:

```
error stamp@0.1.0: The engine "node" is incompatible with this module. Expected version ">=24". Got "22.23.2"
```

CI is still pinned to Node 22 because the `node-version: '22'` -> `'24'` edit in `.github/workflows/test.yml` needs a token with the `workflow` scope, which mine lacks (push rejected: `refusing to allow an OAuth App to create or update workflow .github/workflows/test.yml without workflow scope`). Nothing else in the change is unhappy. The diff is in the PR body. Please do not merge before it lands, and there's no `lint.yml`/CI-adjacent file besides `test.yml` pinning a version here, so that one line is the whole remaining piece.

tony8713:
Node 24 confirmed applied in `test.yml` (d45a2aa). `Test` is green: https://github.com/snapshot-labs/stamp/actions/runs/33241921147/job/99072612039

`Lint` is still red though, and it's a second, separate manual-apply spot I missed in my earlier comment. `.github/workflows/lint.yml` calls the shared `snapshot-labs/actions` lint workflow with `target: '22'`, which that reusable workflow feeds straight into `setup-node`. My claim that "there's no lint.yml/CI-adjacent file besides test.yml pinning a version here" was wrong, sorry, this one does too: https://github.com/snapshot-labs/stamp/actions/runs/33241921360/job/99072612449

```diff
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,5 +4,5 @@ jobs:
   lint:
     uses: snapshot-labs/actions/.github/workflows/lint.yml@main
     with:
-      target: '22'
+      target: '24'
     secrets: inherit
```

Same story as before: my token has no `workflow` scope, push rejected with `refusing to allow an OAuth App to create or update workflow .github/workflows/lint.yml without workflow scope` (confirmed by actually trying it). Needs the same manual push you just did for test.yml.

wa0x6e (approved):
utAck

PR #639, opened 2026-08-29
fix(errors): split isRoutineMiss into a shared transport-failure check

### LINKED ISSUE #617: fix(errors): share the image path's routine-miss classifier with address/lookupDomains resolvers

tony8713:
## Summary

wa0x6e's review on #626 pointed out that sharing the full `isRoutineMiss` predicate (DNS/TLS codes, the 4xx band, and `INVALID_ARGUMENT`) across the image and address/lookupDomains resolver families doesn't transfer cleanly: image resolvers fetch a per-input, often user-controlled URL where a 404 is an expected miss, while the address/lookupDomains resolvers all hit a fixed endpoint (ENS subgraph, api.unstoppabledomains.com, D3, an RPC provider) that answers a miss with an empty result rather than an HTTP error, so any 4xx those emit means our own endpoint is broken.

This splits the classifier along that line:
- `isTransportFailure` (`src/helpers/errors.ts`) carries only the DNS/TLS error codes and is shared at all five non-image capture sites (`address/index.ts`, `address/ens.ts` x2, `address/unstoppableDomains.ts`, `lookupDomains/index.ts`).
- The 4xx band and `INVALID_ARGUMENT` stay local to `resolvers/image/index.ts`'s own `isRoutineMiss`, where they were already family-specific policy.

As a side effect, `address/ens.ts`'s provider-fallback catch no longer mutes a malformed address returned by `provider.resolveName` under the old shared predicate's `INVALID_ARGUMENT` clause — that corruption now reports instead of being silently swallowed.

Addresses the review on #626. Closes #617.

### DIFF
--- a/src/helpers/errors.ts
+++ b/src/helpers/errors.ts
@@ -57,3 +57,17 @@ export function isSilencedError(error: any, additionalMessages?: string[]): bool
     )
   );
 }
+
+const TRANSPORT_FAILURE_CODES = [
+  'ENOTFOUND',
+  'EAI_AGAIN',
+  'ECONNREFUSED',
+  'ERR_TLS_CERT_ALTNAME_INVALID',
+  'CERT_HAS_EXPIRED',
+  'UNABLE_TO_VERIFY_LEAF_SIGNATURE',
+  'DEPTH_ZERO_SELF_SIGNED_CERT'
+];
+
+export function isTransportFailure(error: any): boolean {
+  return TRANSPORT_FAILURE_CODES.includes(error?.cause?.code ?? error?.code);
+}
--- a/src/resolvers/address/ens.ts
+++ b/src/resolvers/address/ens.ts
@@ -4,7 +4,7 @@ import { capture } from '@snapshot-labs/snapshot-sentry';
 import snapshot from '@snapshot-labs/snapshot.js';
 import constants from '../../constants.json';
 import { isEvmAddress } from '../../helpers/address';
-import { isSilencedError } from '../../helpers/errors';
+import { isSilencedError, isTransportFailure } from '../../helpers/errors';
 import { graphQlCall } from '../../helpers/graphql';
 import { getProvider } from '../../helpers/provider';
 import { Address, Handle } from '../../helpers/types';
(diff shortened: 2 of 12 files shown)


wa0x6e (approved):
utAck

Issue #638, opened 2026-08-29
fix(image): guard the token resolvers' address parsing and drop the INVALID_ARGUMENT mute


wa0x6e:
`isRoutineMiss` mutes `INVALID_ARGUMENT` ([`src/resolvers/image/index.ts:57`](https://github.com/snapshot-labs/stamp/blob/a4b6bea099110dc3b080daa674e173eff06da406/src/resolvers/image/index.ts#L57)) so that a request carrying something that is not an address does not get reported as a resolver defect. That is the right outcome, but it is bought at the wrong layer: two resolvers throw on caller input and the failure contract catches it, rather than the resolvers answering "no data" the way the rest of the family already does.

The cost is that a genuine `INVALID_ARGUMENT` — ethers rejecting an address that a *contract or upstream API returned* — is muted by the same clause.

## Reproduction

[`parseQuery`](https://github.com/snapshot-labs/stamp/blob/a4b6bea099110dc3b080daa674e173eff06da406/src/helpers/api.ts#L12-L14) trims and lowercases the path segment and validates nothing, and the route regex constrains only `:type`. `constants.resolvers.token` is `[trustwallet, coingecko]`, both `failureContract: true`, both parsing the raw input:

- [`src/resolvers/image/trustwallet.ts:12`](https://github.com/snapshot-labs/stamp/blob/a4b6bea099110dc3b080daa674e173eff06da406/src/resolvers/image/trustwallet.ts#L12)
- [`src/resolvers/image/coingecko.ts:24`](https://github.com/snapshot-labs/stamp/blob/a4b6bea099110dc3b080daa674e173eff06da406/src/resolvers/image/coingecko.ts#L24)

So `GET /token/hello` throws `INVALID_ARGUMENT` ("invalid address") out of both, into `withFailureContract`, where the clause swallows it.

## The rest of the family already guards

`token` is the only family that reaches the clause. Every other image resolver returns no-data instead of throwing:

| resolver | guard |
|---|---|
| `image/lens.ts:34` | `isAddress()` |
| `image/farcaster.ts:23` | try/catch → `null` |
| `image/snapshot.ts:102,110` | try/catch |
| `image/basename.ts` | delegates to `address/basename.getAvatar`, gated on `isEvmAddress` / `normalizeBasename` |
| `image/ens.ts` | `isAddress()` |
| `image/starknet.ts` | `isStarkDomain` / `isStarknetFelt` |
[...]

Issue #517, opened 2026-08-14
Add DefiLlama as a token icon source


tony8713:
With zapper removed in #514 the token icon chain is trustwallet then coingecko. Coingecko needs `COINGECKO_API_KEY`, which is missing from `.env.example` and the test workflow (#515), so CI effectively exercises trustwallet alone.

That leaves thin coverage. Over the 1,820 distinct (chainId, address) pairs used by the strategies of the top 5,000 spaces by followers, trustwallet covers 7.1% as we call it today and 10.8% with the chain name fix in #516. Zapper covered 18.0% before the bucket went dark on 2026-08-03, and 198 of those tokens (10.9% of the set) were served by zapper and not by trustwallet.

DefiLlama publishes token icons at `token-icons.llamao.fi/icons/tokens/{chainId}/{address}`, out of the actively maintained `DefiLlama/icons`. On the same population it covers 25.8%, or 26.2% alongside a fixed trustwallet, and on the coingecko overlap sample it was a strict superset. It returns webp bytes on a hit and a clean 404 on an unknown token.

Shape of the fix: add a defillama resolver to the token chain. It is keyed by chainId, so unlike trustwallet and coingecko it needs no name map and no per-chain allowlist, and it does not silently lose a chain the day we start serving one.

PR #641, opened 2026-09-05
feat: replace CoinGecko with DefiLlama token icons and fix BNB Chain logos

bonustrack:
CoinGecko Pro needed a paid key and its platform map never included BNB Chain. This swaps the token image resolver for DefiLlama's keyless icon service, which is keyed by numeric chain id, so the platform map and `COINGECKO_API_KEY` are gone. Same change as snapshot-labs/sx-monorepo#2353 on the UI side, applied to images instead of prices.

While looking at BNB Chain: the native asset had no entry in `getBaseAssetIconUrl` and fell through to the Ethereum logo, and trustwallet was pointed at the `binance` directory (Beacon Chain) instead of `smartchain`, so every BEP-20 on chain 56 rendered a blockie. Both fixed. Fallback blockies are never written to S3, only served with a one hour cache header, so BEP-20 misses need no cache clearing. The two BNB native-asset sentinels are different: they previously resolved to the Ethereum logo, which was stored in the base cache with no expiry, so clear them once after deploy:

```
GET /clear/token/bnb:0x0000000000000000000000000000000000000000
GET /clear/token/bnb:0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE
```

DefiLlama serves webp rather than PNG, so the resolver snapshot baseline was regenerated. Tokens neither source carries still get a blockie: ANDY on BNB Chain, and everything on ApeChain, where DefiLlama has no 33139 icons and TrustWallet has no `apechain` directory. CoinGecko listed some ApeChain tokens, but production already serves blockies for them.

## Test Plan
1. Run `yarn dev` with no `AWS_REGION` and no `REDIS_URL` set so every request resolves live
2. Open `http://localhost:3008/token/bnb:0x0000000000000000000000000000000000000000?s=128` and confirm the yellow BNB logo, not the Ethereum diamond
3. Open `http://localhost:3008/token/bnb:0x55d398326f99059fF775485246999027B3197955?s=128` and confirm the Tether logo instead of a blockie
4. Open `http://localhost:3008/token/eth:0xc18360217d8f7ab5e7c516566761ea12ce7f9d72?resolver=defillama` and confirm the ENS logo comes back through DefiLlama alone
5. Once deployed, open `/s:dao.spaceid.eth/treasury` on the Snapshot UI and check the BNB Chain treasury shows real logos for BNB, USDT, USDC and ID within an hour


src/helpers/chains.ts
```diff
@@ -10,14 +10,17 @@ export function chainIdToShortName(chainId: string): string | null {
...
+  if (chainId === '56') return 'smartchain';
   if (chainId === '250') return 'fantom';
   if (chainId === '137') return 'polygon';
   if (chainId === '42161') return 'arbitrum';
   return null;
 }
 
 export const getBaseAssetIconUrl = (chainId: string) => {
+  // BNB
+  if (chainId === '56')
```
wa0x6e: The "no cache clearing is needed after deploy" claim in the PR body holds for BEP-20 blockie misses, but not for the two native-asset sentinels this branch fixes.

Before this change, `bnb:0x0000000000000000000000000000000000000000` and `bnb:0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` went through trustwallet's sentinel check into `getBaseAssetIconUrl('56')`, fell through to the generic Ethereum SVG, and fetched it *successfully*. That is a normal resolver hit, so `src/api.ts` wrote it to the S3 base cache. The cache key carries no resolver name or version and `src/aws.ts` sets no expiry, and the GET route reads the base cache before running any resolver, so on those two ids the new `56` branch here never executes.

Measured against production today: `/token/bnb:0x0…0?s=128` and `/token/eth:0x0…0?s=128` return byte-identical images (sha256 `8d6ddebe…27d6d7c6`, 2410 bytes, same for the `0xEeee…` form). That is the request the test plan's step 5 relies on, so the stale entry is already there.

No code change needed. The repo already has the endpoint; run once after deploy:

```
GET /clear/token/bnb:0x0000000000000000000000000000000000000000
GET /clear/token/bnb:0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE
```

`clear()` deletes by key prefix, so each call drops the base entry and every resized variant (query-param variants such as a non-default `fb`/`fit` are separate prefixes). Responses already served keep their 12h `Cache-Control`, so allow for that before judging step 5. Worth qualifying the PR body accordingly.

src/constants.json
```diff
@@ -6,7 +6,7 @@
   "resolvers": {
     "avatar": ["snapshot", "ens", "basename", "lens", "farcaster", "starknet"],
     "user-cover": ["user-cover"],
-    "token": ["trustwallet", "coingecko"],
+    "token": ["trustwallet", "defillama"],
```
wa0x6e: ApeChain (33139) belongs in the "still get a blockie" sentence of the PR body, next to ANDY on BNB Chain.

The deleted CoinGecko map carried `33139: 'apechain'`, and CoinGecko's public API does serve real artwork for tokens there (WAPE `0x48b62137EdfA95a428D35C09E44256a739F6B557`, CURTIS `0xFC2744A6Db0f97c606Df786b97255DFf6F27E320`, both 250×250). Neither source on this branch carries them: `token-icons.llamao.fi/icons/tokens/33139/<addr>` answers 404 for WAPE, ApeUSD and CURTIS, and TrustWallet's `assets` repo has no `blockchains/apechain` directory, so a `chainIdToName` entry could not help either.

It is not a user-visible regression: production returns a fresh blockie for all three today (`cf-cache-status: DYNAMIC`, `max-age=3600`, decoded as 8×8 identicons), so whatever CoinGecko key is configured there is not resolving ApeChain either. ApeChain also does not appear in the token population #517 measured. Documenting the gap seems enough; keeping CoinGecko as a keyed third fallback for DefiLlama misses would work but goes against the keyless goal of this change.

src/resolvers/image/defillama.ts
```diff
@@ -0,0 +1,12 @@
+import { getAddress } from '@ethersproject/address';
+import { fetchHttpImage } from '../../helpers/http';
+
+// DefiLlama keys token icons by numeric chain id and contract address, with no
+// API key, and answers 404 for a token it does not carry.
+export default async function resolve(address: string, chainId: string) {
+  if (!/^\d+$/.test(chainId)) return false;
```
wa0x6e: `getAddress` throws `INVALID_ARGUMENT` for an id that is not an address (`GET /token/hello` reaches the resolver unvalidated, `parseQuery` only lowercases), and only the `isRoutineMiss` mute in the failure contract turns that into "no data". Every other image resolver answers `false` itself. #638 proposes this exact guard, but against `coingecko.ts`, which this branch deletes, so it has to move here anyway.

```suggestion
import { getAddress } from '@ethersproject/address';
import { isEvmAddress } from '../../helpers/address';
import { fetchHttpImage } from '../../helpers/http';

// DefiLlama keys token icons by numeric chain id and contract address, with no
// API key, and answers 404 for a token it does not carry.
export default async function resolve(address: string, chainId: string) {
  if (!/^\d+$/.test(chainId)) return false;
  if (!isEvmAddress(address)) return false;
```

wa0x6e (approved):
tAck

PR #644, opened 2026-09-07
fix(image): stop Neynar 402 captures when NEYNAR_API_KEY is unset

### LINKED ISSUE #643: fix(image): every avatar cache miss reports a Neynar 402 since the API key was removed (STAMP-7J)
Sentry: [STAMP-7J](https://snapshot-labs.sentry.io/issues/STAMP-7J) — **33213 occurrences**, one isolated event 2026-08-27 15:52 UTC, continuous since 2026-09-05 13:11 UTC, last seen 2026-09-07 13:00 UTC, `environment:production`, `handled:yes`, `provider:farcaster`. Loudest open error in the project (next one has 9674 and stopped eight days ago).

Every avatar request that misses the base cache sends a Neynar lookup with an empty API key. Neynar refuses it, and stamp reports the refusal to Sentry, about 700 times an hour, steadily, since `NEYNAR_API_KEY` was removed from the production environment. Nothing is broken for the user: the resolver answers `false` and the other avatar resolvers fill in. The cost is Sentry noise that buries everything else in the project.

## What is reported

```
Error: [farcaster] 
```

The message is blank after the prefix: only the HTTP reason phrase goes into it, and production's fetch returns none. The status lives on `error.status`, which Sentry does not render.

Stack, first-party frames only:

```
src/resolvers/image/farcaster.ts:39   throw httpError('farcaster', response.status, response.statusText)
src/helpers/deadline.ts:13            withDeadline
src/resolvers/image/farcaster.ts:50   resolve → fetchAddressImageUrl
src/resolvers/image/index.ts:57       withFailureContract → capture
src/api.ts:134                        Promise.all(currentResolvers.map(...))  (index 4)
```

Sample event, 2026-09-07 13:00:54 UTC: `GET https://cdn.stamp.fyi/avatar/0xFA51d273fDAf5B23024F38A8Eb8B04Ec2beE53af`, resolver args `["0xfa51d273fdaf5b23024f38a8eb8b04ec2bee53af", "1", null]`, pod `stamp-5bc975cfb6-86756`.

| pod | first | last | events |
|---|---|---|---|
| stamp-7b6c78c7b-n7n7n | 2026-08-27 15:52 | 2026-08-27 15:52 | 1 |
| stamp-cc6c7b878-zkqck | 2026-09-05 13:11 | 2026-09-05 13:22 | 142 |
| stamp-6f6cdc845f-74899 | 2026-09-05 13:23 | 2026-09-06 09:13 | 9178 |
| stamp-5bc975cfb6-86756 | 2026-09-06 09:14 | ongoing | 23892 |

## Reproduction

1. With `NEYNAR_API_KEY` unset, request an avatar that is not in the base cache, e.g. `curl -s -o /dev/null http://localhost:3000/avatar/0xFA51d273fDAf5B23024F38A8Eb8B04Ec2beE53af`.
2. Expected: no Sentry event, the other resolvers answer. Actual: one `[farcaster]` capture per request, `error.status` 402.

The upstream answer, probed 2026-09-07:

```
$ curl -si 'https://api.neynar.com/v2/farcaster/user/bulk-by-address?addresses=0xFA51d273fDAf5B23024F38A8Eb8B04Ec2beE53af' -H 'api_key:'
HTTP/1.1 402 Payment Required
{"x402Version":1,"error":"X-PAYMENT header or API key required", ...}
```

## Cause

- `src/resolvers/image/farcaster.ts:9` reads the key as `process.env.NEYNAR_API_KEY ?? ''`, so a missing key still sends the request at `:32` with an empty `api_key` header.
- Neynar answers 402 and `:39` throws `httpError('farcaster', 402, '')`.
- `src/resolvers/image/index.ts:38-44`: `isRoutineMiss` treats 4xx as a routine miss except `AUTH_STATUS_CODES = [401, 402, 403]`, so the 402 reaches `capture` at `:60`. That exclusion is deliberate and correct: a dead credential must stay visible. The defect is that the resolver runs at all when there is no credential to be dead.
- `src/constants.json:7` keeps `farcaster` in the avatar chain, and `src/api.ts:134` fans out to it on every base-cache miss.

Not a regression. The resolver has always sent the request with whatever key it had, and 401/402/403 have been kept out of the routine band since [f1db8e9](https://github.com/snapshot-labs/stamp/commit/f1db8e9ee60d0016374e0a5fd1006b9312e11009) (2026-08-24). The trigger is the removal of `NEYNAR_API_KEY` from the production environment, effective with the pod rollout at 2026-09-05 13:11 UTC. The single 2026-08-27 event predates that and is unrelated.

## Fix

Keep the resolver in the tree but dormant, off the chain:

1. `src/resolvers/image/farcaster.ts`, top of `resolve`: `if (!API_KEY) return false;` so a missing key is a quiet no-avatar instead of a doomed request.
2. Remove the `farcaster` entry from `RESOLVERS` (`src/resolvers/image/index.ts:105`) and from `constants.resolvers.avatar` (`src/constants.json:7`). Both: `src/api.ts:134` indexes the registry by the constants list, and a name present in one but not the other throws `TypeError` on every avatar request.
3. `test/integration/resolvers/image/farcaster.test.ts` reaches the resolver through the registry (`helper.ts:35`), so it either imports `farcaster.ts` directly or goes with its snapshot png. The README lists Farcaster under avatar resolvers; drop that line.

Related, not covered here: STAMP-77 `[storage.googleapis.com] ` shows the same blank `httpError` message. Putting the status in the message is a separate change.

wa0x6e:
Every avatar request that misses the base cache asks Neynar with an empty API key since `NEYNAR_API_KEY` left the production environment. Neynar answers 402, which is deliberately not a routine miss, so each miss lands in Sentry: STAMP-7J, about 700 captures an hour.

A missing key now answers no-avatar without asking, and Farcaster is off the avatar chain until a key is configured again. The resolver stays registered for when it is.

Closes #643


ChaituVR (approved):
utAck

tony8713 (commented):
Confirmed the 402 captures do stop, but not quite for the reason the PR body describes, plus one small regression worth a follow-up.

**The `if (!API_KEY) return false` guard in `farcaster.ts:48` is dead code today.** `constants.json` already drops `"farcaster"` from `resolvers.avatar` in this same PR, so `api.ts` never calls `resolvers.farcaster` at all: not via the default avatar chain (`api.ts:116,134`), and not via the explicit `?resolver=farcaster` query either, since that path checks membership in the same array first (`api.ts:126`). Confirmed by mutation: deleting the guard line leaves all 16 related test suites (172 tests) green. The fix for #643 is entirely the `constants.json` change; the guard is forward-defense for if/when farcaster gets re-added to a chain (and it's real protection there, since `deadline.test.ts` and `noData.test.ts` call `farcaster()` directly, bypassing `api.ts`), but it isn't load-bearing for today's Sentry fix. Worth a one-line correction to the PR body language ("stays registered" undersells that it's also currently unreachable from any HTTP path).

**`GET /avatar/:id?resolver=farcaster` now 500s instead of degrading gracefully.** Before this PR it passed the `.includes()` check, ran the resolver, got nothing back, and served the normal 200 fallback avatar. Now the same request fails that check and returns `{status: 'error', error: 'invalid resolvers'}` at `api.ts:127`. No test exercises this path either before or after. Probably fine since it's arguably more correct (farcaster genuinely isn't a valid avatar resolver right now), but flagging since it's an observable behavior change with zero coverage.

**Nit:** README still lists Farcaster as an avatar resolver; issue #643's own fix plan mentioned dropping that line.

Nothing here blocks anything since this already merged, just context for a possible follow-up.

PR #646, opened 2026-09-11
fix(helpers): stop validation.ts importing limits from resolvers

### LINKED ISSUE #549: refactor: helpers/validation.ts should not import from resolvers
## Problem

`src/helpers/validation.ts` imports `MAX_LOOKUP_ADDRESSES` / `MAX_RESOLVE_NAMES` from `src/resolvers/address/index.ts`. Helpers sit below resolvers everywhere else in the tree; this is the one upward edge, and it pulls the whole address-resolver graph (redis client, metrics, every resolver module) into anything that imports the request schemas.

The same limits are also enforced a second time inside `resolvers/address/index.ts:_call` (`input.length > maxInputLength` → 400), so the check exists at two layers with two error formats.

## Proposal

- Move `MAX_LOOKUP_ADDRESSES` / `MAX_RESOLVE_NAMES` into `constants.json` (already holds `max`, `ttl`, `shortTtl`) or define them in `helpers/validation.ts` and import them from there in `resolvers/address`.
- Drop the duplicate length check in `_call` once zod is the single gate at the API boundary (or keep it and say why in a comment — `lookupAddresses`/`resolveNames` are also called internally from `image/ens.ts`, so decide whether internal callers need the guard).

Related: #546 (helpers regroup).

tony8713:
\`src/helpers/validation.ts\` imported \`MAX_LOOKUP_ADDRESSES\`/\`MAX_RESOLVE_NAMES\` from \`src/resolvers/address\`, the one upward edge from helpers into resolvers in the tree, pulling the whole address-resolver graph (redis client, metrics, every resolver module) into anything that imports the request schemas.

Both limits move into \`constants.json\`, already the home for \`max\`/\`maxCover\`/\`ttl\`/\`shortTtl\`, and \`validation.ts\` and \`resolvers/address/index.ts\` both read them from there instead.

The duplicate length check inside \`_call\` stays: \`lookupAddresses\`/\`resolveNames\` are also called directly from \`resolvers/image/ens.ts\` and from tests, bypassing the zod schema at the API boundary, so it remains the only guard for those internal callers.

Closes #549


wa0x6e (approved):
utAck

PR #648, opened 2026-09-12
fix(address): drop expired ENS names in resolve_names

### LINKED ISSUE #147: fix(ens): resolve_names returns addresses for expired ENS names

tony8713:
## Problem

`lookupDomains` already filters out expired ENS names (`src/resolvers/lookupDomains/ens.ts`), but `resolveNames` in `src/resolvers/address/ens.ts` queries the ENS subgraph by `name_in` with no expiry filter, so an expired name still resolves to its last known address.

Fix #147

## Fix

Query `expiryDate` alongside `resolvedAddress` and drop a domain whose expiry is in the past, the same numeric comparison `lookupDomains` uses. An expired name is excluded outright rather than falling through to the `provider.resolveName` fallback, since the on-chain resolver record for an expired name isn't cleared until the grace period ends and it's re-registered, so that fallback could silently return the same stale address.

## Test fixture

The integration test's fixture domain, `sdntestens.eth`, turned out to have itself expired (2024-12-20), so `resolveNames()` now correctly drops it, which broke the "happy path" assertions in `test/integration/resolvers/address/ens.test.ts`. Its reverse record still points to the name, so the `lookupAddresses()` half kept passing. Swapped the fixture to `vitalik.eth`, verified live against the subgraph and provider (forward, reverse, and an expiry far enough out not to repeat this).


wa0x6e (approved):
utAck

PR #649, opened 2026-09-12
fix(http): reject an oversized data: URL before fetch() decodes it

### LINKED ISSUE #627: fix(http): fetchHttpImage blocks the event loop decoding a large data: URL, and the request deadline cannot preempt it

tony8713:
## Summary

Fixes #627.

`fetch()` decodes a `data:` URL's payload synchronously inside the call itself, before `withDeadline`'s `setTimeout` abort ever gets a tick to run. A large `data:` URL handed to `fetchHttpImage`/`fetchWithDeadline` therefore stalls the whole process for the full decode, and the 5s deadline provides no protection at all. `readBoundedImage`'s existing size cap can't help either, since it only inspects a `Response` that doesn't exist yet while that decode is running.

`fetchWithDeadline` rejects on the URL's byte length before `fetch()` is ever called, for every URL, not only `data:` ones, since a scheme-prefix check alone still lets a case or whitespace variant through to the decode. This is the single point every current and future caller routes through (`fetchHttpImage` plus `starknet.ts`'s direct use); today `ens` is the only resolver that actually hands a `data:` URL into this path. `lens` and `farcaster` already reject one at their `isHttpUrl` gate (since #631), and `coingecko` no longer exists in this repo (replaced by `defillama` in #641). The central placement still covers `basename`, `snapshot`, `space-sx` and `starknet` uniformly, rather than needing a per-resolver check.

The cap is measured with `Buffer.byteLength(url)`, the URL's UTF-8 byte size, not its JS `.length` (UTF-16 code units), since those two diverge on a multi-byte-character payload and only the byte count bounds the decode.

It is measured against a dedicated `MAX_URL_BYTES`, not the existing `MAX_IMAGE_BYTES`. `MAX_IMAGE_BYTES` is a decoded-image-size budget, how large a downloaded image we're willing to accept, and it's sized accordingly (10MB). This guard needs a decode-time budget instead, how long we're willing to block the event loop synchronously inside `fetch()`, which is a much smaller number. The two budgets coincide only by accident, and sharing one between them let a `data:` URL up to 10MB clear the guard while still blocking the process for the length of that decode.

## Test plan

- [x] `npx tsc --noEmit` and `npx eslint src/ test/` clean
- [x] `jest test/integration/helpers/http.test.ts` covers a data: URL under/over each cap, a case/whitespace variant, a non-ASCII payload whose UTF-16 length hides its true byte size, a percent-encoded payload, and both cap boundaries
- [x] CI


tony8713:
Addressed a finding from a review that landed in our Discord rather than as a GitHub review: `fetchWithDeadline`'s guard matched `url.startsWith('data:')` against the raw string, but `fetch()` itself accepts `DATA:`, other-case variants, and leading whitespace as data URLs too, so those bypassed the guard and still hit the synchronous decode it exists to prevent (confirmed in Node: `DATA:...`, `  data:...`, etc. all decode).

Fixed in 6ffaa1d by capping the raw string length for every URL through this path, not only ones matching a literal `data:` prefix. No legitimate http(s) URL is anywhere near 10MB, so this closes the whole bypass class rather than trying to replicate the URL parser's scheme normalization. Added regression tests for the uppercase-scheme and leading-whitespace variants; mutation-proved by reverting to the old prefix check, which turns the new test red (TypeError, fetch actually gets called) while the other 8 cases in the file stay green.

src/helpers/http.ts
```diff
@@ -17,10 +17,16 @@ export function spaceIds(id: string): string[] | null {
   }
 }
 
+export const MAX_IMAGE_BYTES = 10 * 1024 * 1024;
+
 export function fetchWithDeadline<T>(
   url: string,
   read: (response: Response) => Promise<T>
 ): Promise<T> {
+  if (url.length > MAX_IMAGE_BYTES) {
```
wa0x6e:
```suggestion
  if (Buffer.byteLength(url) > MAX_IMAGE_BYTES) {
```

(Updated for `6ffaa1d` — the earlier version of this comment still carried the `startsWith('data:')` test that commit removed.)

`url.length` counts UTF-16 code units, but `fetch` decodes the URL's UTF-8 bytes, so a raw (non-base64) `data:` URL made of multi-byte characters passes this guard and still decodes ~3x the cap:

```
data:image/png, + 10,485,735 x '中'
url.length             = 10,485,750   (passes; cap is 10,485,760)
Buffer.byteLength(url) = 31,457,220   -> 377-441 ms of synchronous block on Node 24.20.0
```

Reachable on the same unauthenticated route this PR closes: `src/resolvers/image/ens.ts:29` admits any body matching `/^data:[^,]*,/`, so an owner-controlled ENS avatar record can be exactly this shape. Base64 and percent-encoded payloads are pure ASCII, so `Buffer.byteLength === .length` for them and their effective caps are unchanged.

Worth a regression case too — the `data:` block still only exercises ASCII payloads, which are the one shape where the two measures agree:

```ts
it('rejects a non-ASCII body whose UTF-16 length hides its true byte size', async () => {
  // Each '中' is 1 UTF-16 code unit but 3 UTF-8 bytes, so url.length stays
  // well under the cap while the decoded byte size blows past it.
  const oversized = `data:image/png,${'中'.repeat(3_500_000)}`;
  const fetchSpy = jest.spyOn(global, 'fetch');

  expect(oversized.length).toBeLessThan(MAX_IMAGE_BYTES);

  await expect(fetchHttpImage(oversized)).rejects.toMatchObject({
    status: 404,
    message: expect.stringContaining('too large')
  });
  expect(fetchSpy).not.toHaveBeenCalled();

  fetchSpy.mockRestore();
});
```

wa0x6e: Open PR #598 deletes `fetchWithDeadline` (this guard's home) and rewrites both `fetchHttpImage` and `starknet.ts`'s `fetchImageOrMetadata` to call `fetch(url, { signal })` inline with no size check — while also adding `if (url.startsWith('data:')) return url;` to `getUrl`, so a `data:` URL starts reaching that raw `fetch()` on the Starknet metadata path.

Verified with a merge simulation of both orders: whichever of #649/#598 merges second hits a real conflict in `src/helpers/http.ts` (not a silent auto-merge), but resolving it correctly means manually re-porting this guard into both of #598's new call sites, and nothing enforces that. #598's thread discusses conflicts with #457 and #610 but has never mentioned #627.

Worth flagging on #598 now, before that conflict actually has to be resolved.

wa0x6e: This PR's size guard sits in `fetchWithDeadline` right next to the scheme allowlist open PR #632 adds there (`assertAllowedUrl`, unreviewed as of writing). That allowlist admits `http:`/`https:` only and rejects `data:` with a 400, which `src/resolvers/image/index.ts`'s `isRoutineMiss` classifies as a silent, unreported miss (`AUTH_STATUS_CODES` only excludes 401/402/403).

So regardless of merge order, #632 would make `data:` handling here unreachable and regress the `data:` avatar support merged in #628/#631 — `src/resolvers/image/ens.ts` still accepts `data:` records — with no Sentry signal: the resolver just returns `false` and the avatar falls back to a generic identicon.

Recommend #632 exempt `data:` from `assertAllowedUrl` before it merges. A `data:` URL never reaches the network dispatcher, so that PR's SSRF rationale doesn't apply to it.

wa0x6e: **PR description's exposure list and `data:`-only framing are stale**

The rationale in the description no longer matches this repo or this branch's head:

- `coingecko` doesn't exist anymore — it was replaced by `defillama` in #641. There is no `src/resolvers/image/coingecko.ts` to protect.
- `lens` (`src/resolvers/image/lens.ts:60`) and `farcaster` (`src/resolvers/image/farcaster.ts:54`) both gate on `isHttpUrl` before calling `fetchHttpImage`, and `isHttpUrl` rejects `data:` outright (protocol must be `http:`/`https:`) — since #631. Neither can hand a `data:` URL to this path today.
- The only resolver that currently feeds a `data:` URL into `fetchWithDeadline`/`fetchHttpImage` is `src/resolvers/image/ens.ts:29`, which explicitly allows one past its own `data:` regex.
- Separately, `6ffaa1d` changed the guard from `url.startsWith('data:') && url.length > MAX_IMAGE_BYTES` to `url.length > MAX_IMAGE_BYTES` unconditionally — so "for any `data:` URL" is inaccurate too; the guard now caps every URL string.

None of this changes the fix itself, or the call to keep the check centralized in `fetchWithDeadline`. It is the paragraph justifying the central placement that should describe what is true today rather than resting on #627's exposure list, which predates #631's `isHttpUrl` gate and #641's coingecko removal.

Suggested replacement for that paragraph:

> `fetchWithDeadline` now rejects on the URL string length before `fetch()` is ever called, for every URL — not only `data:` ones, since a scheme-prefix check alone still let a case or whitespace variant through to the decode. This is the single point every current and future caller routes through (`fetchHttpImage` plus `starknet.ts`'s direct use); today `ens` is the only resolver that actually hands a `data:` URL into this path — `lens` and `farcaster` already reject one at their `isHttpUrl` gate (since #631), and `coingecko` no longer exists in this repo (replaced by `defillama` in #641). The central placement still covers `basename`, `snapshot`, `space-sx` and `starknet` uniformly, rather than needing a per-resolver check.

test/integration/helpers/http.test.ts
```diff
@@ -131,4 +131,43 @@ describe('fetchHttpImage', () => {
...
+      ]) {
+        await expect(fetchHttpImage(oversized)).rejects.toMatchObject({
+          status: 404,
+          message: expect.stringContaining('too large')
+        });
+      }
+      expect(fetchSpy).not.toHaveBeenCalled();
+
+      fetchSpy.mockRestore();
+    });
```
wa0x6e, line 173: The `data:` cases here are all ASCII base64, which is the one encoding where `url.length` and the decoded byte size roughly agree. Two shapes are still uncovered: a percent-encoded (non-base64) payload, and a URL sized to exactly `MAX_IMAGE_BYTES` — the latter pins the `>` vs `>=` edge of the new comparison, which is currently only tested at `MAX + 1`.

```suggestion
    });

    it('still decodes a non-base64 (percent-encoded) one under the cap', async () => {
      const percentEncoded = [...BODY]
        .map(byte => `%${byte.toString(16).padStart(2, '0')}`)
        .join('');
      const dataUrl = `data:image/png,${percentEncoded}`;

      await expect(fetchHttpImage(dataUrl)).resolves.toEqual(BODY);
    });

    it('still decodes one exactly at the cap', async () => {
      const prefix = 'data:image/png;base64,';
      const dataUrl = `${prefix}${'A'.repeat(MAX_IMAGE_BYTES - prefix.length)}`;

      expect(dataUrl.length).toBe(MAX_IMAGE_BYTES);
      await expect(fetchHttpImage(dataUrl)).resolves.toBeInstanceOf(Buffer);
    });
```

src/helpers/http.ts
```diff
@@ -30,8 +36,6 @@ export function fetchWithDeadline<T>(
   }, 5e3);
 }
 
-export const MAX_IMAGE_BYTES = 10 * 1024 * 1024;
-
 export async function readBoundedImage(url: string, response: Response): Promise<Buffer> {
   const host = new URL(url).host;
```
wa0x6e: [pre-existing, but on the path this PR is making cheap] `new URL(url).host` runs on every successful fetch, including the `data:` path. On a ~9.3 MB `data:` URL (just under the cap, same shape as the new test) that measured ~12 ms per call — 20 iterations, avg 11.88 ms — and `host` is only read in the two throw branches below, so the success path pays it for nothing.

It also buys nothing when it *is* used: `new URL('data:image/png;base64,AAAA').host === ''`, which is why an oversize `data:` URL raises `[] image too large: ...`.

Computing it lazily in the two error branches keeps the hot path free:

```diff
-  const host = new URL(url).host;
   const declared = Number(response.headers.get('content-length'));
   if (declared > MAX_IMAGE_BYTES) {
     await response.body?.cancel();
-    throw httpError(host, 404, `image too large: ${declared} bytes`);
+    throw httpError(new URL(url).host, 404, `image too large: ${declared} bytes`);
   }
...
     if (total > MAX_IMAGE_BYTES) {
-      throw httpError(host, 404, `image too large: over ${MAX_IMAGE_BYTES} bytes`);
+      throw httpError(new URL(url).host, 404, `image too large: over ${MAX_IMAGE_BYTES} bytes`);
     }
```

wa0x6e (changes requested):

src/helpers/http.ts
tony8713: Fixed in 9b191d9, swapped to `Buffer.byteLength(url) > MAX_IMAGE_BYTES` exactly as suggested. Added your `中`-repeat regression case; mutation-proved by reverting to `url.length`, which turns only that test red (11/12 stay green).

tony8713: Confirmed against #598's current diff: it does delete `fetchWithDeadline` and adds `if (url.startsWith('data:')) return url;` to `getUrl`, so a `data:` URL reaches a bare `fetch()` on the Starknet path with no size check. Whichever of #649/#598 merges second hits a real conflict in `http.ts`, and resolving it correctly means re-porting this guard into both of #598's new call sites manually. Leaving the coordination call to you and #598's author rather than editing #598 myself.

tony8713: Confirmed — #632's `assertAllowedUrl` (`ALLOWED_PROTOCOLS = new Set(['http:', 'https:'])`) would 400 every `data:` URL, and since `isRoutineMiss` only excludes 401/402/403 that comes back as a silent miss, regressing `ens`'s `data:` avatar support with no signal. Agree #632 should exempt `data:` from that check before merging; not touching #632 myself, flagging it for you/that PR's author to decide.

tony8713: Updated the description with your replacement paragraph, adjusted 'URL string length' to 'byte length' since 9b191d9 also fixes the UTF-16-vs-UTF-8 measure you flagged above. Verified ens.ts:29, lens.ts:60, farcaster.ts:54, and that there's no coingecko.ts left in this tree.

test/integration/helpers/http.test.ts
tony8713: Added both cases from your suggestion (percent-encoded under the cap, exact-cap boundary) plus the non-ASCII case from your first comment, 12/12 pass. Mutation-proved the boundary one specifically: flipping `>` to `>=` on line 26 turns only 'still decodes one exactly at the cap' red, the other 11 stay green.

src/helpers/http.ts
tony8713: Applied your diff, `host` is no longer computed up front in `readBoundedImage`, only inline in each of the two throw branches now.

src/helpers/http.ts
```diff
@@ -17,10 +17,16 @@ export function spaceIds(id: string): string[] | null {
...
 }
 
+export const MAX_IMAGE_BYTES = 10 * 1024 * 1024;
+
 export function fetchWithDeadline<T>(
   url: string,
   read: (response: Response) => Promise<T>
 ): Promise<T> {
+  if (Buffer.byteLength(url) > MAX_IMAGE_BYTES) {
+    return Promise.reject(httpError('url', 404, `url too large: over ${MAX_IMAGE_BYTES} bytes`));
```
wa0x6e: **Major — the cap this guard reuses is a decoded-image budget, not a decode-time budget, so ~100-130 ms of synchronous event-loop block survives it.**

`fetch()` decodes a `data:` URL inside the call expression, before it returns a promise — which is the property #627 is about, and the reason `withDeadline` cannot help. This guard bounds that decode, but it bounds it with `MAX_IMAGE_BYTES`, a constant that exists to cap how large a *decoded image* we accept over HTTP. How big an image we want and how long we are willing to freeze the process are different budgets; they coincide here only by accident.

Measured on Node v24.20.0, timing only the synchronous portion (`const p = fetch(url)`, before any `await`), median of 5 per size:

```
    64 KiB encoded ->   1.5 ms
   256 KiB encoded ->   3.8 ms
   512 KiB encoded ->   7.8 ms
  1024 KiB encoded ->   9.6 ms
  4096 KiB encoded ->  40.4 ms
 10240 KiB encoded -> 102.3 ms   <- what this guard still admits
```

Two runs of the at-cap case disagreed — 102.3 ms interleaved in the sweep above, 125.4 ms standalone after a settle — so read the residual as ~100-130 ms rather than one figure. Cross-checked against a 1 ms `setInterval` starvation counter, read after yielding to the macrotask queue: max inter-tick gap 127.8 ms at the cap versus 3.6 ms for a 26-byte `data:` URL control. It is a real loop block, not wall time.

Failure scenario: an ENS name whose `avatar` text record is a 10 MiB base64 `data:` URL. `src/resolvers/image/ens.ts:29` admits it (`/^data:[^,]*,/`), this guard admits it (10,485,760 bytes, exactly at the cap), and every request to the image route for that name freezes the process ~115 ms. Unauthenticated, one attacker-owned name, no concurrency limit: roughly 8 req/s saturates the event loop.

No legitimate `http(s)` URL is anywhere near either number, so a separate, much smaller URL-length cap costs nothing on the HTTP side and keeps the guard scheme-agnostic — the new constant belongs here at the choke point, not as a `data:`-only branch, since a scheme test is exactly what 6ffaa1d removed for good reason:

```suggestion
export const MAX_IMAGE_BYTES = 10 * 1024 * 1024;

// Bounds the synchronous decode fetch() performs on a data: URL inside the call
// itself, before withDeadline's abort can ever run. Sized for decode time, not
// image size: ~10ms of blocked event loop at 1 MiB, ~115ms at 10 MiB.
const MAX_URL_BYTES = 1024 * 1024;

export function fetchWithDeadline<T>(
  url: string,
  read: (response: Response) => Promise<T>
): Promise<T> {
  if (Buffer.byteLength(url) > MAX_URL_BYTES) {
    return Promise.reject(httpError('url', 404, `url too large: over ${MAX_URL_BYTES} bytes`));
  }
```

The number is the part worth deciding, not the shape. 1 MiB caps the block at ~10 ms but rejects `data:` avatars over 1 MiB that work today, a regression against the `data:` support added in #628/#631 — though an SVG avatar that size is already pathological. 4 MiB keeps ~40 ms and leaves more headroom. Either way the existing `data:` tests should keep passing; only the oversized fixtures need resizing to the new constant.

tony8713: Fixed in 3d871ee: split into a dedicated MAX_URL_BYTES (1MB, your smaller option) used only for this front-door check, leaving MAX_IMAGE_BYTES for the decoded-size cap in readBoundedImage where it belongs. Added a regression test for a data: URL between the two caps (over MAX_URL_BYTES, still well under MAX_IMAGE_BYTES); mutation-proved by reverting to MAX_IMAGE_BYTES here, which turns exactly the tests sized in that gap red and leaves the rest green.

tony8713:
Addressed the finding from your latest review (the "Major" one on `src/helpers/http.ts:27`, posted after 9b191d9): the guard was capping decode time with `MAX_IMAGE_BYTES`, a decoded-image-size budget, so a `data:` URL up to 10MB still cleared it and still blocked the event loop for the full synchronous decode inside `fetch()`.

Fixed in 3d871ee by splitting that into a dedicated `MAX_URL_BYTES` (1MB) used only for this front-door check, leaving `MAX_IMAGE_BYTES` for what it always meant, the decoded-size cap in `readBoundedImage`. Went with your suggested shape and the smaller of your two proposed numbers.

Regression test added: a `data:` URL sized between the two caps (over 1MB, still well under 10MB) now gets rejected before `fetch()` is ever called. Mutation-proved by reverting the guard back to `MAX_IMAGE_BYTES`: that turns exactly the three tests sized in that gap red (the new one plus the case/whitespace-variant and non-ASCII cases, which needed resizing down to the new cap to still demonstrate anything), the other 9 in the file stay green.

Also resized the existing over/under/exact-cap fixtures in `test/integration/helpers/http.test.ts` down to `MAX_URL_BYTES` so they pin the guard that's actually in effect.

wa0x6e (approved):
utAck

PR #652, opened 2026-09-14
fix(image): treat a refused avatar download as a missing image

### LINKED ISSUE #650: fix(image): avatar hosts answering 401/402/403 are reported as resolver errors (STAMP-73, STAMP-77)
Sentry: [STAMP-73](https://snapshot-labs.sentry.io/issues/STAMP-73) — **128 occurrences**, first seen 2026-08-20 08:49 UTC, last seen 2026-09-13 13:44 UTC, `environment:production`, `handled:yes`, `provider:ens`.
Sentry: [STAMP-77](https://snapshot-labs.sentry.io/issues/STAMP-77) — **135 occurrences**, first seen 2026-08-20 16:47 UTC, last seen 2026-09-12 13:38 UTC, `environment:production`, `handled:yes`, `provider:space`.

An ENS or space avatar that points at a host refusing anonymous downloads (a private Google Cloud Storage object, a Pinata gateway, a Cloudinary asset) reports a Sentry error on every cache miss for that name, forever. Nothing is broken for the user: the resolver answers `false` and the route falls back to the next resolver or a blockie, exactly as it does for a 404. Nothing on our side can make those hosts serve the image.

Both groups are grab-bags keyed on the `httpError` stack frame, not on this defect. Over the last 14 days (all inputs probed 2026-09-13):

- STAMP-77: **21 of 34** events are a 402/403 — `app.helena.financial` 403 (10), `storage.googleapis.com` 403 (8, spaces `thecypherverse.eth` and `dsla.eth`), `uploads-ssl.webflow.com` 403, `braze-images.com` 403, `blockblend.io` 402. The other 12 are `gasdao.eth`, whose origin answered 520 at the time and 404 today.
- STAMP-73: **14 of 49** events are a 401/403 — `catblox.mypinata.cloud` 403 (8, `bttb.eth`), `img.rarible.com` 401 (5, `gmdao.eth`), `ik.imagekit.io` 403 (1). Not this defect: third-party 5xx (`luisc.xyz` 522, `neotokyo.codes` 500; 10 events), 20 names that fall through to `metadata.ens.domains` (14 of them in one second at 2026-09-01 05:44 UTC) and answer 404 today, and 5 events whose host no longer answers at all.

For scale, stamp served ~699k requests over the same 14 days (Prometheus, `http_request_duration_seconds_count{instance=~"stamp.fyi:.*"}`).

## What is reported

```
Error: [img.rarible.com] 
```

The message is blank after the host because the upstream speaks HTTP/2, which has no reason phrase, so `statusText` is `''`.

```
src/api.ts:133                          Promise.all(currentResolvers.map(r => resolvers[r](address, network, networkId)))
src/resolvers/image/index.ts:57         return await resolve(...args);            (withFailureContract)
src/resolvers/image/index.ts:69         const input = await resolve(...args);     (withResize)
src/resolvers/image/ens.ts:33           return await fetchHttpImage(url);
src/helpers/deadline.ts:14              return await fn(controller.signal);
src/helpers/http.ts:40                  throw httpError(host, response.status, response.statusText);
```

STAMP-73 sample: `GET https://cdn.stamp.fyi/avatar/0xD18e205b41eEe3D208D3B10445DB95Ff02ba4acA`, input `["0xd18e205b41eee3d208d3b10445db95ff02ba4aca", "1", null]`. The address reverse-resolves to `gmdao.eth`, whose avatar record is `https://img.rarible.com/prod/image/upload/prod-itemImages/0xd07dc4262bcdbf85190c01c996b4c06a461d2430:706480/788e5263`.

STAMP-77 sample: `GET https://cdn.stamp.fyi/space/dsla.eth`, input `["dsla.eth", "1", null]`, same stack through `src/resolvers/image/snapshot.ts:144`. The space avatar is `https://storage.googleapis.com/stacktical-public/dsla.png`.

## Reproduction

1. `curl -s -o /dev/null -w '%{http_code}\n' 'https://img.rarible.com/prod/image/upload/prod-itemImages/0xd07dc4262bcdbf85190c01c996b4c06a461d2430:706480/788e5263'` → `401`
2. `curl -s -o /dev/null -w '%{http_code}\n' https://storage.googleapis.com/stacktical-public/dsla.png` → `403`
3. With `SENTRY_DSN` set, request `GET /avatar/0xD18e205b41eEe3D208D3B10445DB95Ff02ba4acA` and `GET /space/dsla.eth` on a cold cache.

Expected: both answer the fallback image with no Sentry event, as a 404 from the same hosts would.
Actual: both answer the fallback image and each reports an `Error: [<host>] ` event.

## Cause

`readHttpImage` turns a non-ok response into `httpError(host, response.status, response.statusText)` (`src/helpers/http.ts:38-41`), keeping the upstream status. `withFailureContract` captures unless `isSilencedError` or `isRoutineMiss` says otherwise (`src/resolvers/image/index.ts:59`). `isRoutineMiss` treats 400–499 as a routine miss **except** 401/402/403 (`src/resolvers/image/index.ts:38-47`), because `withFailureContract` also wraps a resolver's own authenticated API calls, where those statuses mean a dead key. `isSilencedError` only silences 429 and 504 among statuses (`src/helpers/errors.ts:55`). So a 401/402/403 from an avatar host is captured.

The carve-out's reason does not hold for `fetchHttpImage`: it calls `fetch(url, { signal })` with no headers (`src/helpers/http.ts:32`). The only credentialed calls on the image path are the Snapshot Hub query (`src/resolvers/image/snapshot.ts:64`) and the Neynar lookup (`src/resolvers/image/farcaster.ts:32`), and both raise their own errors before `fetchHttpImage` runs. A 401/402/403 out of `fetchHttpImage` can only be a host refusing an anonymous download.

Not a regression. The 401/402/403 carve-out arrived in [f1db8e9](https://github.com/snapshot-labs/stamp/commit/f1db8e9ee60d0016374e0a5fd1006b9312e11009), the same change that started muting 4xx at all; avatar downloads have been caught by it since.

## Fix

Classify the status where the request is known to be anonymous, alongside the other "no usable image here" answers `readHttpImage` already maps to 404:

```ts
if (!response.ok) {
  // No credentials are sent, so 401/402/403 is the host refusing this image,
  // not the dead API key isRoutineMiss keeps those statuses loud for.
  const status = [401, 402, 403].includes(response.status) ? 404 : response.status;
  throw httpError(host, status, response.statusText);
}
```

A Neynar or Hub 401 still reports: neither goes through `readHttpImage`. Upstream 5xx are unchanged.

wa0x6e:
Closes #650

An ENS, space or Farcaster avatar hosted somewhere that refuses anonymous downloads (a private Google Cloud Storage object, a Pinata gateway, a Rarible or Cloudinary asset) raised a Sentry error on every cache miss for that name. Users saw nothing wrong: the resolver already fell back exactly as it does for a 404.

That download never carries credentials, so a 401/402/403 there can only be the host refusing the image. It is now a routine miss like any other 4xx. A 401/402/403 from our own authenticated calls (Snapshot Hub, Neynar) still reports, and upstream 5xx are unchanged.

Measured from the upstream status recorded on each event over the last 14 days, this clears 16 of the STAMP-73 events and 22 of the STAMP-77 events. None of the 401/402/403 in 30 days came from a host we choose ourselves (IPFS gateway, ENS metadata service, token icon sources), so no infrastructure block is being muted.


tony8713 (approved):
Verified independently: mutating the status-flatten line fails exactly the 3 new 401/402/403 tests and nothing else, and none of the resolvers' avatar-download URLs actually carry credentials (only the separate metadata/API calls do). One nit: the AUTH_STATUS_CODES comment in resolvers/image/index.ts still says the exclusion covers 'third-party avatar download', which is no longer true now that readHttpImage flattens those to 404 before isRoutineMiss ever sees them.

PR #653, opened 2026-09-14
test(resolvers): stall the deadline cases behind an image head

wa0x6e:
Master's Test job fails on "when it sends headers and then nothing more > defillama raises the abort": it gets `not an image: application/json` instead of the abort.

The stalled upstream in that test answers with a JSON head, a leftover from when both resolvers under test read JSON first. DefiLlama fetches the image directly, and since #598 the image reader rejects a non-image type before touching the body, so the case failed fast and stopped proving the body read is bounded by the deadline.

The stall now sends an image head. DefiLlama reads into the stalled body and aborts at its deadline again; Farcaster's JSON read ignores the type and still aborts at its own. Checked against a build whose deadline covers only the request: the headers case still passes, the body case times out.


tony8713 (approved):
Verified independently: reverting the content-type back to application/json reproduces the exact master failure this fixes (not an image: application/json instead of AbortError), while farcaster's JSON-read path stays unaffected either way. CI confirms it too: PR 652's branch (pre-fix) still shows the deadline test failing, this branch does not.

PR #657, opened 2026-09-14
fix(errors): treat a TLS unrecognized_name alert as a transport failure

### LINKED ISSUE #656: fix(errors): avatar hosts rejecting the TLS server name are reported as resolver errors (STAMP-7A)
Sentry: [STAMP-7A](https://snapshot-labs.sentry.io/issues/STAMP-7A) — **5 occurrences** across 2 spaces, first seen 2026-08-21 23:12 UTC, last seen 2026-08-27 10:33 UTC, `environment:production`, `handled:yes`, `provider:space`.

Spaces whose avatar sits on a host with a broken HTTPS setup show the default space avatar, and every such request is also reported as an error. Nothing is broken for the user: no client can download those images, and nothing on our side can make the hosts reachable. The report is noise, and the two hosts below still fail today, so it recurs whenever those spaces are requested.

## What is reported

```
TypeError: fetch failed
  caused by: Error: 0073631CF47E0000:error:0A000458:SSL routines:ssl3_read_bytes:tlsv1 unrecognized name:../deps/openssl/openssl/ssl/record/rec_layer_s3.c:918:SSL alert number 112
```

Neither exception carries a stack. The path is `src/resolvers/image/snapshot.ts:144` → `src/helpers/http.ts:86`.

Sample event `42bffe0d23ca4f5ebb987ed6d58c40be`, 2026-08-27 10:33:08 UTC, pod `stamp-5896d6d565-9lg9v`, node v26.7.0:

- Request: `GET https://cdn.stamp.fyi/space/s:popworld.eth`
- Resolver input `args`: `["popworld.eth", "1", "s"]`

All 5 events:

| space | events | avatar |
|---|---|---|
| `s:popworld.eth` | 3 | `https://popman.popop.world/icon.png` |
| `mongolnft.eth` | 2 | `https://media.mongolnft.xyz/assets/MNFT_dark.png` |

## Reproduction

1. On Node 26.8.2 (production's version), `await fetch('https://popman.popop.world/icon.png')`.
2. Pass the rejection to `isTransportFailure` from `src/helpers/errors.ts`.

Expected: `true`, as for an unresolvable host (`ENOTFOUND` returns `true`).
Actual: `false`. The rejection's cause carries `code: 'ERR_SSL_TLSV1_UNRECOGNIZED_NAME'`, `library: 'SSL routines'`, `reason: 'tlsv1 unrecognized name'`.

## Cause

TLS alert 112 is `unrecognized_name`: the server rejects the hostname sent in SNI and presents no certificate. Both hosts do this for every client, checked 2026-09-14 13:41 UTC, three attempts each:

```
LibreSSL/3.3.6: error:1404B458:SSL routines:ST_CONNECT:tlsv1 unrecognized name
HTTP 000 — server presented no certificate for this SNI name
```

`withFailureContract` (`src/resolvers/image/index.ts:60`) reports unless `isSilencedError` or `isRoutineMiss` matches. `isRoutineMiss` forgives unreachable hosts through `isTransportFailure` (`src/resolvers/image/index.ts:45`), which compares `cause.code` against `TRANSPORT_FAILURE_CODES` (`src/helpers/errors.ts:65`): `ENOTFOUND`, `EAI_AGAIN`, `ECONNREFUSED`, `ERR_TLS_CERT_ALTNAME_INVALID`, `CERT_HAS_EXPIRED`, `UNABLE_TO_VERIFY_LEAF_SIGNATURE`, `DEPTH_ZERO_SELF_SIGNED_CERT`. `ERR_SSL_TLSV1_UNRECOGNIZED_NAME` is not in it, and `isSilencedError` matches neither the message nor the code.

Sentry displays only the OpenSSL string, not `cause.code`, so this shape was previously taken to have no code. A broader arm matching `cause.library === 'SSL routines'` was added for it in [9649446](https://github.com/snapshot-labs/stamp/commit/964944648668ef2add0592587cd0a2627f182a30) and removed again in [f1db8e9](https://github.com/snapshot-labs/stamp/commit/f1db8e9ee60d0016374e0a5fd1006b9312e11009) within the same PR, so it never shipped. The code is set on Node 22.11.0 and 26.8.2 (checked 2026-09-14), so the existing code list covers it without that arm.

Not a regression: this alert has never been classified.

## Fix

```diff
 const TRANSPORT_FAILURE_CODES = [
   'ENOTFOUND',
   'EAI_AGAIN',
   'ECONNREFUSED',
   'ERR_TLS_CERT_ALTNAME_INVALID',
+  'ERR_SSL_TLSV1_UNRECOGNIZED_NAME',
   'CERT_HAS_EXPIRED',
   'UNABLE_TO_VERIFY_LEAF_SIGNATURE',
   'DEPTH_ZERO_SELF_SIGNED_CERT'
 ];
```

`isTransportFailure` is shared with the address and lookupDomains resolvers, which already treat the certificate failures in this list as routine, so the entry extends that class rather than widening it.

wa0x6e:
## Summary

Closes #656.

Spaces whose avatar sits on a host that rejects the TLS server name (STAMP-7A) were reported to Sentry on every request. The host aborts the handshake with alert 112 and presents no certificate, so no client can download the image; the space already falls back to the default avatar. Node exposes this as `ERR_SSL_TLSV1_UNRECOGNIZED_NAME`, which was not among the codes treated as an unreachable host.

That alert is now treated like the certificate failures already there: silenced as a routine miss for avatar downloads, and likewise in the address and lookupDomains resolvers that share the check. Other TLS failures stay reported.

## Test plan

- [x] Both hosts from the Sentry events, fetched live on Node 24 and on 26.8.2 (production): the rejection carries the code and is now classified as a transport failure
- [x] `tsc` and eslint clean
- [ ] CI


tony8713 (approved):
Verified: ERR_SSL_TLSV1_UNRECOGNIZED_NAME is the real Node code for OpenSSL alert 112 (checked on Node's TLS error mapping), and it's a single-point fix since isTransportFailure is shared across all the resolvers that need it. Mutation-tested the new test cases by reverting the array entry: both go red (the errors.test.ts case and the failureContract.test.ts case), rest of the 55 related tests stay green. CI's 5 Starknet failures are unrelated (upstream RPC returning HTML instead of JSON, also flaky on master right now, unconnected to this diff).

PR #658, opened 2026-09-14
fix(image): decode Starknet NFT profile pictures ourselves

### LINKED ISSUE #655: fix(image): Starknet profile pictures from Cairo 1 NFTs never render (STAMP-7R)
Sentry: [STAMP-7R](https://snapshot-labs.sentry.io/issues/STAMP-7R) — **8 occurrences** across 4 Starknet addresses, first seen 2026-09-13 22:58 UTC, last seen 2026-09-13 23:52 UTC, `environment:production`, `handled:yes`, `provider:starknet`.

A Starknet user whose starknet.id profile picture is a Blobert NFT gets the fallback avatar instead of their NFT, and every such request is reported as an error. The NFT's metadata is valid on-chain; it is corrupted by one trailing character on its way to us, and the download then fails.

## What is reported

```
TypeError: fetch failed
  caused by: failed to fetch the data URL
```

The rest of the stack is undici's. Our last frames:

```
src/helpers/http.ts:86   const response = await fetchBounded(url, signal);
src/helpers/http.ts:32   return fetch(url, { signal });
```

Sample event `dabc1fb6c1764c7abf6df133b5fc73c8`, 2026-09-13 23:52:29 UTC, pod `stamp-5455cc578d-9f4sh`, node v26.8.2:

- Request: `GET https://cdn.stamp.fyi/avatar/0x022b3840acceb9698a3e22bd602d63c8f7666c0aaba488a92253931ba4539c94`
- Resolver input `args`: `["0x022b3840acceb9698a3e22bd602d63c8f7666c0aaba488a92253931ba4539c94", "1", null]`

The 8 events are two bursts over the same 4 addresses, all Blobert holders:

```
0x04afd95ca9ce0cd53ac5ab05606d7200cd2922714b744fc3730b07e771a59c9e
0x022b3840acceb9698a3e22bd602d63c8f7666c0aaba488a92253931ba4539c94
0x058e058cc515404046b9a125531b92dfb9542b02829447381483520df7e7e39e
0x008fe793143ae0c3f00bc50e9572a2c8a1d812d97cb700f9a4e7cf7e432b1f25
```

## Reproduction

1. On mainnet, call `provider.getStarkProfile('0x022b3840acceb9698a3e22bd602d63c8f7666c0aaba488a92253931ba4539c94')`.
2. `fetch()` the returned `profilePicture`, a `data:application/json;base64,eyJuYW1lIjoiQmxvYmVydCAjMjg3NCIs…` URL whose payload ends in `…ifV198`.

Expected: a JSON metadata document whose `image` is `data:image/svg+xml;base64,…`.
Actual: `TypeError: fetch failed`, cause `failed to fetch the data URL`. Same for all four addresses (2026-09-14).

## Cause

`src/resolvers/image/starknet.ts:106` takes the picture from `provider.getStarkProfile(address).profilePicture`. starknet.js decodes the NFT's `token_uri` reply there as a Cairo 0 `Array<felt252>`:

```js
// starknet@6.21.1 (our lockfile) dist/index.js:6254 — identical in 10.8.0, the latest release
finalArray[8].slice(1).map((val) => decodeShortString(val)).join("")
```

Blobert (`0x539f522b29ae9251dbf7443c7a950cf260372e69efab3710a11bf17a9599f1`) is a Cairo 1 contract and replies with a `ByteArray`: `[num_full_words, ...full_words, pending_word, pending_word_len]`. `slice(1)` drops only the word count, so `pending_word_len` is decoded as text as well. For `0x022b…c94` the raw reply is 138 felts ending in `['0x6232776966563139', '0x8']`, and the URL gains a trailing `8`.

That extra character makes the base64 payload 4165 characters long, 1 mod 4, which is never valid base64. Node's `Buffer.from(…, 'base64')` tolerates it; `fetch()` applies the WHATWG forgiving-base64 decode and rejects the whole URL. All four payloads are 1 mod 4, with trailing `3`, `8`, `8` and `4`.

The resulting error matches neither classifier in `withFailureContract` (`src/resolvers/image/index.ts:60`): its cause has no `code` for `isTransportFailure` and there is no status for `isRoutineMiss`, so it is captured.

The resolver already has a decoder for both reply shapes, `decodeTokenUri` (`src/resolvers/image/starknet.ts:70`), reached through `getNftProfilePicture` only when `getStarkProfile` throws `ENTRYPOINT_NOT_FOUND` (`src/resolvers/image/starknet.ts:108`). Decoding the same `token_uri` reply with it gives a 4164-character payload, and `fetchHttpImage(url, followMetadata)` on it returns the NFT's SVG for all four addresses (2079 bytes for `0x022b…c94`).

Not a regression: these pictures have never rendered. Before [37bd0d6](https://github.com/snapshot-labs/stamp/commit/37bd0d6b7daec4238e497c79281f8594842fae7f), `getUrl` rewrote every `data:` URL onto the IPFS gateway, which answered 400; that commit let `data:` URLs through, which moved the failure into `fetch()`.

## Fix

Keep `getStarkProfile` for telling an identicon from an NFT picture, but read an NFT picture's `token_uri` through `getNftProfilePicture`, which decodes both shapes:

```ts
const picture = (await provider.getStarkProfile(address))?.profilePicture;
if (!picture || picture.startsWith('https://starknet.id/api/identicons/')) return picture ?? null;

return withDeadline(signal => untilAborted(signal, getNftProfilePicture(address)));
```

This costs five sequential RPC calls instead of one multicall, for profiles with an NFT picture only. The decoding bug itself is upstream in starknet.js `getStarkProfile` and is unreported there.

wa0x6e:
Closes #655

Starknet profile pictures from NFTs whose `tokenURI` replies with a Cairo 1 `ByteArray` come out of starknet.js's profile lookup with one extra trailing character. When that character is a control code the URL parser strips it and the picture renders by luck; when it is a digit the URL is corrupt. For Blobert holders, whose pictures are `data:` URLs, `fetch()` rejects the URL outright: the fallback avatar is served and every request is reported to Sentry (STAMP-7R).

The resolver now sends the same starknet.id multicall itself, without the social fields it never used, and decodes the token URI for both reply shapes. It still reads through `tokenURI` first, since several whitelisted profile-picture contracts are Cairo 0 and expose only that, and retries with `token_uri` for the contracts that only have that one.

A profile costs a single RPC request, or two when its NFT contract only has `token_uri`.


tony8713 (approved):
Verified against starknet.js's actual multicall decode: it slices off the ByteArray's length word but still short-string-decodes `pending_word_len` along with the data, which is exactly the stray trailing character described. Re-reading via `tokenURI` and decoding with the existing Cairo1-aware `decodeTokenUri` is the right fix, and the entrypoint choice is sound since a successful multicall proves `tokenURI` exists on that contract. Reverted the core change and the new stray-character test goes red while the identicon test stays green, so the coverage discriminates correctly.

tony8713 (commented):
Reviewed the second commit (2209958f), pushed after the earlier approval which only covered a4af1475. It replaces the getStarkProfile-then-reread-tokenURI approach with a hand-rolled starknet.id `aggregate` multicall, and drops the `withDeadline`/`untilAborted` wrapping.

Traced the multicall construction against starknet.js's own reference decode and the provider's per-call timeout: the removal of the deadline wrapper is safe (each RPC call is still individually bounded, and there's now only one or two round trips instead of up to five), and the skip-indexing for a profile with no NFT (`results[4]` undefined, falls through to the identicon) checks out.

One gap worth knowing about: `test/unit/resolvers/image/starknet.test.ts`'s mock returns a fixed payload regardless of the calldata sent, so the multicall's cross-call reference wiring (`starknet.ts:43-66`, the `[i,j]` indices into prior call results) isn't actually exercised. Mutating a reference index (wrong target contract, swapped token-id felt order) left all 27 tests green; a control mutation to the decode path correctly reddened 10. Not a bug in the current code as far as I can tell, just an untested edge that a future edit to that block could break silently.

No blocking findings, LGTM.

PR #659, opened 2026-09-15
refactor(lookupDomains): derive the default chain set from CHAIN_IDS

### LINKED ISSUE #621: refactor(lookupDomains): derive the default chain set from CHAIN_IDS
## Problem

`src/resolvers/lookupDomains/index.ts:21` builds the dispatcher's default chain set out of each provider's own fallback argument:

```ts
const DEFAULT_CHAIN_IDS = PROVIDERS.map(provider => provider.DEFAULT_CHAIN_ID);
```

So `DEFAULT_CHAIN_ID` does two unrelated jobs — the provider's default when called bare (`chainId = DEFAULT_CHAIN_ID`), and a vote in the **global** default set. That means any provider can widen the default for every *other* provider.

Measured on #614, counting outbound requests for a single default `lookup_domains` call (no `network`), with `global.fetch` spied:

| `ensV2.DEFAULT_CHAIN_ID` | outbound requests |
|---|---|
| `'1'` (as shipped) | 1 — mainnet v1 subgraph |
| `'11155111'` | 3 — mainnet v1 subgraph, **sepolia v1 subgraph**, ENSv2 staging |

The sepolia v1 subgraph in the second row is the point. ENSv2 does not serve chain 1, but pointing its default at sepolia drags ENS **v1** onto a chain nothing asked for, and testnet names into a default production lookup.

That is the only reason `ensV2.DEFAULT_CHAIN_ID` is `'1'` today — a chain that provider cannot serve. It is a workaround for the dispatcher rather than a statement about ENSv2, and it breaks the invariant every sibling provider holds, `DEFAULT_CHAIN_ID ∈ CHAIN_IDS`.

## Proposal

Derive the set from what providers actually serve, minus testnets.

```ts
// src/helpers/chains.ts
import networks from '@snapshot-labs/snapshot.js/src/networks.json';

export function isTestnet(chainId: string): boolean {
  return !!(networks as Record<string, { testnet?: boolean }>)[chainId]?.testnet;
}
```

```ts
// src/resolvers/lookupDomains/index.ts
const DEFAULT_CHAIN_IDS = [...new Set(PROVIDERS.flatMap(provider => provider.CHAIN_IDS))].filter(
  chainId => !isTestnet(chainId)
);
```

A provider can then only contribute chains it serves, so a testnet-only provider contributes nothing and cannot pull a sibling onto a new chain. A provider adding a *mainnet* chain does enter the default set and other providers serving it get called there — that is the intended auto-enable, not accidental widening.

## Verification

```
union of all CHAIN_IDS : ["1","11155111","109","157","146"]
after testnet filter   : ["1","109","146"]
today                  : ["1","109","146"]   → same set
```

Per-provider routing is identical, so this is a zero-behavior-change refactor:

| provider | today | proposed |
|---|---|---|
| Ens | `[1]` | `[1]` |
| Ens V2 | `[]` | `[]` |
| Shibarium | `[109]` | `[109]` |
| Unstoppable Domains | `[146]` | `[146]` |

Existing unit tests pass unchanged — the mocks in `test/unit/resolvers/lookupDomains.test.ts` hardcode `CHAIN_IDS`, and their union filters to the same three chains.

## Knock-on changes

- `DEFAULT_CHAIN_ID` stops feeding the dispatcher and keeps only its real job, each provider's bare-call default. It can also come out of the `Provider` type at `src/resolvers/lookupDomains/index.ts:13`, since nothing in that file would read it. (Touches the contract discussed in #550.)
- `src/resolvers/lookupDomains/ensV2.ts:8` can become `'11155111'`, which is the honest value, and the explanatory comment above it at lines 6-7 goes away.

## One branch to name deliberately

`!!networks[chainId]?.testnet` treats a chain **absent** from snapshot.js as mainnet, so it would land in the defaults. Unreachable today — every chain any provider serves is listed, with the right flag:

| chain | name | testnet |
|---|---|---|
| 1 | Ethereum | false |
| 109 | Shibarium | false |
| 146 | Sonic | false |
| 157 | Shibarium Puppynet Testnet | **true** |
| 11155111 | Sepolia testnet | **true** |

Suggest keeping the one-liner with a comment rather than building a fail-closed path for a case that cannot happen yet, but it should be a stated choice: if it ever fires, an unlisted testnet leaks into production defaults, which is the class this change exists to prevent.

## Note on the dependency

`networks` is not on snapshot.js's public export (`{ Client, Client712, schemas, utils }`) and is not in `dist/`, so this is a deep import. It is published on purpose — `package.json` declares `files: ["dist", "src"]` — there is no `exports` map to block it, and `resolveJsonModule` is already on, so it typechecks clean (`tsc --noEmit` exit 0 with the helper in `src/`). Isolating it in `src/helpers/chains.ts` keeps it to one line to fix if a future major relocates it.

Stamp's own `src/chains.json` cannot do this job — it is chainId↔shortName only, with no testnet field.

## Also in these files

Two test cleanups from the #614 review, worth folding in since this change touches the same files:

- `test/unit/resolvers/lookupDomains.test.ts:59` — `does not widen the default chain set to a chain no provider defaults to` reads `DEFAULT_CHAIN_ID` from the module mock, not from `ensV2.ts`, so it cannot fail if the value it names changes. Verified: setting `ensV2.ts` to `'11155111'` leaves it green while `test/unit/resolvers/lookupDomains/ensV2.test.ts:66` goes red. What remains duplicates the test above it plus one assertion, which is better merged into it.
- `test/unit/resolvers/lookupDomains/ensV2.test.ts:66` — `does not query anything on the default chain` is the test that actually guards the value, and it will go red by design once an `ensV2Graphql` mainnet entry lands. Worth a note so it is not deleted as stale. Dissolves entirely if `DEFAULT_CHAIN_ID` becomes `'11155111'` as above.

wa0x6e:
Closes #621

`DEFAULT_CHAIN_ID` did two unrelated jobs: a provider's own fallback when called bare, and a vote in the dispatcher's **global** default chain set. So any provider could widen the default for every *other* provider.

Concretely, ENSv2 only serves sepolia. Naming sepolia as its default would have put sepolia in the global default set, and ENS **v1** — which does serve sepolia — would then be queried there on every default lookup, returning testnet names from a production call. Avoiding that is the only reason ENSv2 declared chain `1`, a chain it cannot serve, breaking the invariant every sibling holds: `DEFAULT_CHAIN_ID ∈ CHAIN_IDS`.

The default set is now the union of the chains providers actually serve, minus testnets. A provider can only contribute a mainnet chain it serves, and a testnet-only provider contributes nothing. A provider that adds a *mainnet* chain does enter the default set and siblings serving it get called there — that is intended auto-enable, not accidental widening.

No behaviour change: the resolved set is `["1","109","146"]` before and after, and per-provider routing is identical. ENSv2 now names its real chain, and `DEFAULT_CHAIN_ID` keeps only its real job.

One branch named deliberately: a chain absent from snapshot.js counts as mainnet and would land in the defaults. Unreachable today — every chain a provider serves is listed with the right flag — but if it ever fires, an unlisted testnet leaks into production defaults, which is the class this change prevents. Covered by a test so the choice is visible rather than implicit.


tony8713 (approved):
Verified the arithmetic: union of all providers' CHAIN_IDS minus testnets resolves to ["1","109","146"], same as before. Mutated the aggregation filter, isTestnet, and ensV2's DEFAULT_CHAIN_ID one at a time; each mutation reddened the right tests and only those. Looks correct.

PR #660, opened 2026-09-15
fix(image): guard trustwallet's address parsing

### LINKED ISSUE #638: fix(image): guard the token resolvers' address parsing and drop the INVALID_ARGUMENT mute

wa0x6e:
Closes #638

`GET /token/<anything>` handed the raw path segment to an address parser that throws on anything it cannot read, so any non-address id raised `INVALID_ARGUMENT` out of the trustwallet resolver. Both token resolvers key their icon URL on an EVM contract address, so such an id can never resolve; they now answer no-data for it, like every other image resolver.

That throw was the only reason the image failure contract muted `INVALID_ARGUMENT`, and the mute was indiscriminate: an address a *contract or upstream API* returned and ethers rejected was silenced alongside it. Both token resolvers now check the id against exactly what the parser accepts, so no request can raise the code, the mute goes, and that case is reported again.

A contract address sent without its `0x` prefix is an address to the parser, so it keeps resolving on trustwallet and resolves on DefiLlama again — DefiLlama answered no-data for it since it replaced CoinGecko in #641, which accepted it.

Two premises in the issue did not survive checking, and neither changes the fix:

- It names CoinGecko as a second unguarded resolver. DefiLlama replaced it in #641 and already guarded; trustwallet was the last one.
- It names the basename, gwei and ENS address resolvers as what still raises the code. Those parse inside the address-resolution path, which catches and reports on its own, so they never reached the image path's classifier and were never muted by it. After this change the clause has no known source at all — anything that does raise it there is an unexplained defect, which is the point of reporting it.


src/resolvers/image/trustwallet.ts
```diff
@@ -8,6 +9,8 @@ const ETH = [
 ];
 
 export default async function resolve(address, chainId) {
+  if (!isEvmAddress(address)) return false;
```
tony8713: isEvmAddress requires the `0x` prefix; `getAddress` doesn't, it adds it. A bare 40-hex token id (no `0x`) used to checksum and resolve here and now fails this guard, falling through to the blockie fallback instead of the real icon — confirmed against master, e.g. `GET /token/a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48` (USDC without `0x`) served the real logo before this PR and serves blockie after. Might be fine if no one relies on the unprefixed form, but it's a silent behavior change the PR doesn't mention.

tony8713: Separately: isEvmAddress's regex accepts any-case 40-hex, so a mixed-case address with a bad EIP-55 checksum (e.g. `0xa0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48`) passes this guard and still throws INVALID_ARGUMENT from `getAddress` below — which is now captured to Sentry instead of muted. It's not reachable through the route today only because `parseQuery` lowercases the id twice before any resolver sees it (src/helpers/api.ts:13,28), not because this guard catches it. defillama.ts has the identical gap. Not asking for a fix, just flagging that the guard's safety is coupled to a lowercasing step in a different file rather than being self-contained.

test/unit/resolvers/image/failureContract.test.ts
```diff
@@ -170,6 +164,17 @@ describe('resolvers - failure contract', () => {
     }
   );
 
+  // Every resolver checks the id it was handed before parsing it, so an
+  // INVALID_ARGUMENT can no longer come from the request -- it means a resolver
+  // parsed something an upstream returned.
```
tony8713: This comment overclaims a bit given the checksum gap noted above: it's not that "every resolver checks the id" that makes INVALID_ARGUMENT unreachable from a request, it's parseQuery's lowercasing upstream. Not a blocker, just flagging since it reads as the mechanism guaranteeing the new test's premise.

tony8713 (commented):
Traced the load-bearing claim (trustwallet's new guard removes the last caller-input source of INVALID_ARGUMENT, so dropping the mute from isRoutineMiss is safe) against every other getAddress call site in the image resolvers: lens/snapshot/farcaster/basename/ens already guard with isAddress or try/catch, defillama already guards with isEvmAddress. Mutating the guard and the removed clause separately confirms both new tests discriminate (the trustwallet case goes red without the guard, the classifier case goes red if the clause comes back). The claim holds for what the route can send it today. Two things beside it are worth a look, both inline below, plus a note on the two comments the diff adds.

test/unit/resolvers/image/noData.test.ts
```diff
@@ -175,6 +176,21 @@ describe('resolvers answer false rather than throwing when there is no data', ()
     });
   });
 
+  // Both token resolvers key their icon URL on an EVM contract address, so a
+  // `token` id that is not one can never resolve.
```
tony8713: comment-gate flags this one too, and per repo convention it mostly restates the test titles below it ("which is not an address"). Not blocking, just noting alongside the other one.

src/resolvers/image/trustwallet.ts
wa0x6e: Confirmed live: on master `a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48` fetched the real USDC logo from trustwallet, on 8003f16 it fell back to the blockie.

Fixed in [e008fb6](https://github.com/snapshot-labs/stamp/commit/e008fb69871d97743a6bb96e1b736ba1ac1ccfbb): the guard is now ethers' `isAddress`, which is `getAddress` in a try/catch, so it accepts exactly what the next line parses. Unprefixed ids resolve again on trustwallet, and on defillama too, which had rejected them since #641 replaced coingecko (unguarded, so it accepted them). Covered by a test on both token resolvers; PR body now mentions it.

wa0x6e: Closed by the same change in [e008fb6](https://github.com/snapshot-labs/stamp/commit/e008fb69871d97743a6bb96e1b736ba1ac1ccfbb): `isAddress` rejects a bad checksum itself, so the guard no longer leans on the route's lowercasing. Applied to defillama as well, with `0xa0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48` added to the no-data cases for both resolvers.

test/unit/resolvers/image/failureContract.test.ts
wa0x6e: Removed in [e008fb6](https://github.com/snapshot-labs/stamp/commit/e008fb69871d97743a6bb96e1b736ba1ac1ccfbb). With the guard change both token resolvers now reject anything `getAddress` would throw on, but the test title carries the case and the reasoning lives in the PR body.

test/unit/resolvers/image/noData.test.ts
wa0x6e: Removed in [e008fb6](https://github.com/snapshot-labs/stamp/commit/e008fb69871d97743a6bb96e1b736ba1ac1ccfbb).

tony8713 (approved):
Re-reviewed after e008fb6. Both risk findings from the first pass are fixed: the guard is now ethers' `isAddress` (which is `getAddress` wrapped in try/catch), so it accepts a bare 40-hex id without `0x` and rejects a bad-checksum mixed-case address itself, no longer leaning on `parseQuery`'s lowercasing. New adversarial pass mutation-confirmed the two new tests actually discriminate on that boundary, checked the ICAP edge case (accepted by `isAddress` but dead on the real call path), and found nothing else. Simplify pass made no changes. Resolved the four threads from the first review.

PR #661, opened 2026-09-15
fix(image): skip onchain spaces with an empty metadata link

wa0x6e:
Space avatar requests for an onchain space whose metadata link is empty fail instead of answering "no image": `api.snapshot.box` returns a partial response with an error on that space's metadata, and the GraphQL helper rejects any response carrying errors. Seen on `/space/eth:0x11DbAFcB0b8A966c6730Af5Cda5a7D7E5FF2Bb58` (18 events since 2026-08-22).

Such spaces are now left out of the API query, so they resolve to no image like any space without an avatar. The filter is a workaround for snapshot-labs/sx-monorepo#2261 and can be dropped once that ships.

Fixes STAMP-7B


tony8713 (approved):
Approving. The `metadata_not: ""` fix is correct and well verified: `metadata_not` is a real, schema-typed `String` filter on `Space_filter` on both `api.snapshot.box` and `testnet-api.snapshot.box` (confirmed via live GraphQL introspection, not just "CI passed"), and both the reported broken space and a second known-broken space from sx-monorepo#2261 get cleanly excluded, while a real space with a valid avatar still resolves correctly with the filter applied.

A few things worth a look, none blocking:

1. `src/resolvers/image/snapshot.ts:71-99` (`getOnchainProperty`, used by `space`/`space-cover`): no test exercises this function's onchain query against a real avatar-bearing space. The existing "space avatar" happy-path test (`snapshot.test.ts:32`) resolves through the offchain Hub path, since the default `networkId` is `'s'`, and never reaches the modified query. Coverage here is only indirect, via `space-sx.ts`'s byte-identical query shape hitting the same endpoint. If the two duplicated implementations ever diverge, this file's own suite wouldn't catch a break in `getOnchainProperty` specifically.

2. `src/resolvers/image/space-sx.ts:6-26`: this half of the fix looks like a no-op for the reported STAMP-7B symptom. `getSpaceProperty` already queries both `api.snapshot.box` and `testnet-api.snapshot.box` via `Promise.allSettled` and only throws if both reject; for the reported mainnet-only broken space, the testnet leg fulfills cleanly with no matching space, so the resolver already returned `false` before this fix. Reverting the filter here locally leaves the whole related test set green. Still reasonable as defense-in-depth, but it isn't covering the bug as described, and unlike `snapshot.ts` it has no dedicated regression test (`space-sx.test.ts` is untouched by this PR).

3. The added 3-line comments in `snapshot.ts:84-86` and `space-sx.ts:11-13` reference `sx-monorepo#2261` directly in code; no other comment in `src/` links out to a GitHub issue like this. For `snapshot.ts` the first two lines are also now redundant with the new test at `snapshot.test.ts:42-48`, which already goes red if the filter is reverted. Easy trim if you want it, not asking for anything in its place.

No bugs found.

PR #663, opened 2026-09-16
fix(image): treat any non-2xx image download as a missing image

### LINKED ISSUE #662: fix(image): avatar hosts answering 5xx are reported as resolver errors (STAMP-7S)
Sentry: [STAMP-7S](https://snapshot-labs.sentry.io/issues/STAMP-7S) — **4 occurrences**, first seen 2026-09-14 08:18 UTC, last seen 2026-09-15 14:12 UTC, `environment:production`, `handled:yes`, `provider:ens` / `provider:space`. Stamp served ~66k requests over the same window (Prometheus, `http_request_duration_seconds_count{instance=~"stamp.fyi:.*"}`, 34h).

An ENS or space avatar whose host answers a 5xx reports a Sentry error on every cache miss for that name. A third-party origin that is down for good (a Cloudflare 522 in front of a dead server) reports forever. Nothing is broken for the user: the resolver answers `false` and the route serves the next resolver's image or the fallback, exactly as it does for a 404. The fallback is sent with a short cache header and never stored (`src/api.ts:138-144`), so the failing host is asked again on every expiry.

Events in the group, inputs probed 2026-09-15 17:42 UTC:

| time (UTC) | host | request | today |
|---|---|---|---|
| 2026-09-14 08:18 | `storage.googleapis.com` | `GET /space/dsla.eth` | 403 (3 of 3). Before 401/402/403 were muted, not this defect |
| 2026-09-14 22:27 | `luisc.xyz` | `GET /avatar/eth:0x327b0c1ecd110651dc9994b6764b695c9cdc29ba` | **522 (3 of 3)** |
| 2026-09-15 06:56 | `metadata.ens.domains` | `GET /avatar/0x8e68a7F5f57A330879cDE69979A7BC8dA03C3435` | 404 |
| 2026-09-15 14:12 | `metadata.ens.domains` | `GET /avatar/eth:0x666aEaA4C42F99E450062E8719B0cAF64cD0f233` | 404 |

The two `metadata.ens.domains` events carry no status (see below). Both came from a pod started 2026-09-14 22:58 UTC, after 4xx from this path stopped reporting. `isSilencedError` already mutes 504 and 429, and `fetch` follows redirects, so these were 5xx. Today both names (`zerocoke.eth`, `kinness.eth`) have no avatar record and the service answers `404 {"message":"There is no avatar set under given address"}`. The 5xx was transient.

## What is reported

```
Error: [metadata.ens.domains] 
```

The message is blank after the host because on Node 26 (prod runs v26.8.2; undici 8), `fetch` leaves `statusText` as `''`. Measured on Node 26.2.0 against `metadata.ens.domains`, `storage.googleapis.com` and `httpbin.org/status/503`. Node 22 fills it in. The status is set on the error object, but the Sentry event does not show it.

```
src/api.ts:133                       Promise.all(currentResolvers.map(r => resolvers[r](address, network, networkId)))
src/resolvers/image/index.ts:57      return await resolve(...args);            (withFailureContract)
src/resolvers/image/index.ts:69      const input = await resolve(...args);     (withResize)
src/resolvers/image/ens.ts:33        return await fetchHttpImage(url);
src/helpers/deadline.ts:14           return await fn(controller.signal);
src/helpers/http.ts:43               throw httpError(host, status, response.statusText);
```

Sample: `GET https://cdn.stamp.fyi/avatar/eth:0x666aEaA4C42F99E450062E8719B0cAF64cD0f233`, input `["0x666aeaa4c42f99e450062e8719b0caf64cd0f233", "1", "eth"]`. The address reverse-resolves to `kinness.eth`, which has an empty `avatar` record, so `src/resolvers/image/ens.ts:27-31` falls back to `https://metadata.ens.domains/mainnet/avatar/kinness.eth`.

## Reproduction

1. `curl -s -o /dev/null -w '%{http_code}\n' 'https://luisc.xyz/_next/image?url=%2F_next%2Fstatic%2Fmedia%2Flogo.fb595efc.jpg&w=256&q=75'` → `522`. This is the `avatar` record of `luisc.eth`.
2. With `SENTRY_DSN` set, request `GET /avatar/0x327b0c1ecd110651dc9994b6764b695c9cdc29ba` (reverse record `luisc.eth`) on a cold cache.

Expected: the fallback image is served with no Sentry event, as for a 4xx from the same host.
Actual: the fallback image is served and an `Error: [luisc.xyz] ` event is reported.

## Cause

`readHttpImage` maps every 4xx to 404 and keeps any other status (`src/helpers/http.ts:38-44`). `withFailureContract` captures unless `isSilencedError` or `isRoutineMiss` matches (`src/resolvers/image/index.ts:57`). `isRoutineMiss` only covers 400–499 (minus 401/402/403), transport failures and `INVALID_ARGUMENT` (`src/resolvers/image/index.ts:40-48`). Among statuses, `isSilencedError` only mutes 504 and 429 (`src/helpers/errors.ts:59`). A 500/502/503 or a Cloudflare 520–530 from an avatar host is therefore captured.

Not a regression. 5xx were left visible on purpose when 4xx from image downloads were first muted: a 5xx can also mean a real dependency outage. The plan was to revisit if a 5xx group got noisy. This issue settles that: an image download that fails with a 5xx is a missing image.

## Fix

Treat a 5xx the same as a 4xx where the request is known to be an anonymous image download:

```ts
if (!response.ok) {
  const status = response.status >= 400 ? 404 : response.status;
  throw httpError(host, status, response.statusText);
}
```

What still reports: our authenticated API calls do not go through `readHttpImage` and raise their own errors first. That covers the Snapshot Hub query (`src/helpers/graphql.ts:52`, e.g. a `[hub.snapshot.org] status code 521` stays visible) and the Neynar lookup (`src/resolvers/image/farcaster.ts:39`). The Lens API's 503 stays muted by its own list, as it is today (`src/resolvers/image/lens.ts:15`).

What goes silent along with third-party hosts: a 5xx from a host we pick ourselves also goes through `readHttpImage`. That means the IPFS gateway (`IPFS_GATEWAY`, `src/helpers/http.ts:107`), `metadata.ens.domains` (`src/resolvers/image/ens.ts:31`), `token-icons.llamao.fi` (`src/resolvers/image/defillama.ts:13`), `raw.githubusercontent.com` (`src/resolvers/image/trustwallet.ts:14`) and `wrpcd.net` (`src/resolvers/image/farcaster.ts:17`). Users see the same fallback either way; only the Sentry signal for an outage of one of those goes away.

`test/integration/helpers/http.test.ts:244` (504) and `test/integration/helpers/socketRelease.test.ts:41` (500) pin the upstream status coming out of `fetchHttpImage` and change with this. `test/unit/resolvers/image/failureContract.test.ts:173` ("still reports an upstream 500") tests the resolver-level contract, which still holds for errors that do not come from `readHttpImage`.

wa0x6e:
Fixes #662

An avatar host answering a 5xx reported a Sentry error on every cache miss for that name. A third-party origin that is down for good (a Cloudflare 522 in front of a dead server) reported forever, while the user got the same fallback image a 404 has always produced.

Audited against the issue before taking its patch:

- The production runtime is Node 26.8.2, which leaves `statusText` empty, so these events carry a blank message — confirmed on the Sentry event's `runtime` tag. Node 24 fills it in, so the blank is version-specific and not a second defect.
- A blank-message capture out of the image download proves only that the status was outside the routine band, not that it was a 5xx: a 3xx with no `Location` comes back unfollowed and is reported the same way. Measured against a local server. The issue's patch (`status >= 400 ? 404`) leaves that case loud; dropping the status check entirely covers both and is smaller.
- The issue lists `INVALID_ARGUMENT` as already silenced. It is not, and stays reported.

Authenticated API calls (Snapshot Hub, Neynar) do not use this path and keep reporting. What goes quiet along with third-party hosts is a 5xx from a host we pick ourselves — the IPFS gateway, `metadata.ens.domains`, `token-icons.llamao.fi`, `raw.githubusercontent.com`, `wrpcd.net`. Users get the same fallback either way; only the Sentry signal for an outage of one of those is lost.


tony8713 (approved):
LGTM on the behavior change. The targeted tests pass, and reverting the status mapping makes all changed cases fail; failed downloads still take the same fallback path while authenticated API failures remain reported.

One non-blocking cleanup in `src/helpers/http.ts:40-43`: please remove the added four-line explanation. The parameterized test already pins the policy, and `comment-gate` flags the prose.

The CI failure is in the unchanged Sepolia ENSv2 lookup fixture: the live indexer no longer returns `testspace.eth`, independently of this image change.

PR #664, opened 2026-09-16
test(lookupDomains): replace the missing ENSv2 fixture

tony8713:
The ENSv2 lookup test depends on a registration no longer returned by the staging API. Replace it with an active ENSv2 registration whose owner has no ENSv1 Sepolia domains, preserving the ENSv2 coverage.


wa0x6e (approved):
utAck

PR #667, opened 2026-09-22
fix(errors): treat an unreachable route as a transport failure

### LINKED ISSUE #666: fix(errors): EHOSTUNREACH from a happy-eyeballs connect is reported as a resolver error (STAMP-80)
Sentry: [STAMP-80](https://snapshot-labs.sentry.io/issues/STAMP-80) — **51 occurrences**, first seen 2026-09-21 02:49 UTC, last seen 2026-09-21 06:58 UTC, `environment:production`, `handled:yes`, `provider:lens` (43, avatar GET) and `provider:Lens` (8, `lookup_addresses` POST), all from one pod `stamp-69d7f47fdc-bhgzf`. Second-loudest stamp error of the week after STAMP-7Z (1594, Starknet RPC).

## What is reported

```
TypeError: fetch failed
  cause: AggregateError                      (code: 'EHOSTUNREACH', copied from errors[0])
    errors[0]: Error: connect EHOSTUNREACH 2606:4700:20::ac43:45a4:443 - Local (:::0)
    errors[1]: Error: connect EHOSTUNREACH 2606:4700:20::681a:c1c:443 - Local (:::0)
    errors[2]: Error: connect EHOSTUNREACH 2606:4700:20::681a:d1c:443 - Local (:::0)
```

Stack, all Node internals (no app frames in any event):

```
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:134:8)
    at GetAddrInfoReqWrap.emitLookup [as callback] (node:net:1925:7)
    at defaultTriggerAsyncIdScope (node:internal/async_hooks:474:12)
    at internalConnectMultiple (node:net:1506:5)
    at internalConnectMultiple (node:net:1430:18)
```

Sample events:

- GET `https://cdn.stamp.fyi/avatar/0xB7BE3cde061ecB3a55a50A9854DcCfA8564663a9`, 06:58:34 UTC, `provider:lens`, input `args: ["0xb7be3cde061ecb3a55a50a9854dccfa8564663a9", "1", null]`.
- POST `https://stamp.fyi/`, 06:55:40 UTC, `provider:Lens`, input `lookupAddresses: ["0x8f8c3b3d99A2cD1d4Ca8aA1A386B743510436b16"]`.

Both paths call `https://api.lens.xyz/graphql`; the three addresses are that host's AAAA records (Cloudflare). Two bursts: 40 events 02:49–03:07 while the pod served ~405 requests, then 11 events 06:55–06:58 against ~57 requests (Prometheus `http_request_duration_seconds_count` for that instance). The pod was healthy throughout: `up`=1, event-loop lag max under 0.2 s, no restart (process up since 09-17), and no other pod, project or Betterstack incident in either window. Stamp's Betterstack source holds only "Stored cache" lines for those hours.

What the chain establishes: the lookup returned only AAAA records and the node had no IPv6 route to that prefix. Node's happy-eyeballs connect records every attempt it makes in `errors[]`, timed-out ones included (`lib/net.js` `internalConnectMultipleTimeout` pushes `ETIMEDOUT`), and interleaves IPv6/IPv4, so any A record would appear as a fourth entry. The Sentry SDK does not drop `errors[]` siblings either (`@sentry/core` 10.62.0, `aggregateExceptionsFromError`). Three entries is the whole list. That is a cluster network condition that cleared on its own, not a stamp defect.

## Root cause

All three capture sites run the same gate:

- `src/resolvers/image/index.ts:58`, `withFailureContract`: `!isSilencedError(err, mutedErrors) && !isRoutineMiss(err)`, where `isRoutineMiss` (`:41`) is the 4xx band or `isTransportFailure(error)`.
- `src/resolvers/address/index.ts:71` and `src/resolvers/lookupDomains/index.ts:50`: `!isSilencedError(err) && !isTransportFailure(err)`.

`isTransportFailure` (`src/helpers/errors.ts:76`) is `TRANSPORT_FAILURE_CODES.includes(error?.cause?.code ?? error?.code)`. Here `error.cause` is Node's `NodeAggregateError`, whose constructor sets `this.code = errors[0]?.code` (`lib/internal/errors.js`), so the value checked is `'EHOSTUNREACH'`. The list at `src/helpers/errors.ts:65` holds:

```
ENOTFOUND, EAI_AGAIN, ECONNREFUSED,
ERR_TLS_CERT_ALTNAME_INVALID, ERR_SSL_TLSV1_UNRECOGNIZED_NAME,
CERT_HAS_EXPIRED, UNABLE_TO_VERIFY_LEAF_SIGNATURE, DEPTH_ZERO_SELF_SIGNED_CERT
```

No unreachable-route errno is in it. `isSilencedError` (`src/helpers/errors.ts:22`) does not catch it either: its code list is `TIMEOUT, ETIMEDOUT, ECONNRESET, UND_ERR_SOCKET, 504, 429`, and its `wrapped()` walker follows `cause`/`error`/`serverError`/`response` but never `errors[]`, so only the aggregate's own `code` is visible. Both gates return false and `capture()` runs.

Measured shapes, `net.connect` with a lookup returning only AAAA and `autoSelectFamily` on:

- macOS, Node v24.14.0, IPv6 configured but unrouted: `cause.code = 'EHOSTUNREACH'`, entries `{ code: 'EHOSTUNREACH', errno: -65, syscall: 'connect' }`. Matches production.
- `node:26-alpine` (v26.8.2, the production runtime) with IPv6 disabled: `cause.code = 'EADDRNOTAVAIL'`, entries `{ errno: -99, syscall: 'connect' }`.

So "no IPv6 path" surfaces as `EHOSTUNREACH`, `ENETUNREACH` (the same condition at network scope, not measured here) or `EADDRNOTAVAIL` depending on the host's network config. Same class as the `ENOTFOUND`/`ECONNREFUSED` already muted.

**User impact.** Avatar GET: `withFailureContract` returns `false`, the next resolver answers or the fallback (blockie/jazzicon) is served with `SHORT_CACHE` and not stored (`src/api.ts:133`), so nothing wrong is cached. `lookup_addresses` POST: the failed provider contributes nothing and the merged result is written to Redis with `cacheEmpty: true` for `constants.ttl` = 12 h (`src/resolvers/address/cache.ts:7`), so an address whose only name is a `.lens` handle reads as nameless until then. That is existing behaviour for every silenced transport failure, not part of this issue. Otherwise this is Sentry noise only.

**Provenance.** Not a regression. `TRANSPORT_FAILURE_CODES` was introduced by #639 (2026-08-30) from the cases seen then (STAMP-75 `ENOTFOUND`, STAMP-78 altname mismatch) and widened by #657 (`ERR_SSL_TLSV1_UNRECOGNIZED_NAME`, STAMP-7A). No unreachable-route errno has ever been in it; 2026-09-21 is the first time one was observed.

Sweep: the other unresolved stamp issues (STAMP-71, STAMP-7Z, STAMP-81, STAMP-6X) and the archived `fetch failed` groups (STAMP-7H HTTP/2 GOAWAY, STAMP-6Y undici ConnectTimeout, STAMP-7R data-URL) do not share this cause. Open PRs: #594 adds `ECONNABORTED` to `isSilencedError`'s code list and #568 rewrites its walker; neither touches `TRANSPORT_FAILURE_CODES`. #632 changes only the avatar-download dispatcher, not the GraphQL fetch.

## Fix

```ts
// src/helpers/errors.ts
const TRANSPORT_FAILURE_CODES = [
  'ENOTFOUND',
  'EAI_AGAIN',
  'ECONNREFUSED',
  'EHOSTUNREACH',
  'ENETUNREACH',
  'ERR_TLS_CERT_ALTNAME_INVALID',
  ...
```

plus the two errnos in the `it.each` at `test/unit/helpers/errors.test.ts:49`. Walked against the event: `error.cause.code === 'EHOSTUNREACH'` makes `isTransportFailure` true, so `isRoutineMiss` is true and none of the three sites capture. `EADDRNOTAVAIL` is left out on purpose: it is also what ephemeral-port exhaustion on the pod returns, which is worth seeing.

This is the third arm added to this list in three weeks (#639, #657, now this). The structural alternative: a `cause` (or `cause.errors[0]`) carrying libuv's signature, a numeric `errno` plus `syscall`, is a transport failure by construction whatever the errno, which covers every socket and DNS errno at once and leaves the explicit list for the TLS codes, which carry no `errno`. The list is the smaller diff; the signature check ends the series.

wa0x6e:
A lookup that only returns AAAA records on a pod with no IPv6 route fails every happy-eyeballs attempt with `EHOSTUNREACH` (or `ENETUNREACH`, depending on the host's network config). Node's `AggregateError` carries `errors[0].code`, so that errno is what the transport-failure check sees, and it was not in the list: 51 events from one pod on 2026-09-21 were reported as Lens resolver errors for a cluster network condition that cleared on its own.

Both errnos are now classified as transport failures, so the three capture sites stay quiet the same way they do for `ENOTFOUND` and `ECONNREFUSED`. `EADDRNOTAVAIL` is left out on purpose: it is also what ephemeral-port exhaustion on the pod returns, which is worth seeing.

Fixes STAMP-80
Fixes #666


tony8713 (approved):
Approving. The two errnos join the list `isTransportFailure` already reads through `cause.code`, and since Node's `aggregateErrors` copies `errors[0].code` onto the AggregateError undici hands back as `cause`, an all-AAAA lookup on a v4-only pod does arrive there as `EHOSTUNREACH` or `ENETUNREACH`. Dropping the two entries turns exactly the two new `it.each` cases red and leaves the other 21 in that file green, so the test does guard the change. Keeping `EADDRNOTAVAIL` out is the right call given it is also the port-exhaustion signal. The red Test job on this branch is the four `image/starknet` snapshot cases that are already failing on master, not this change.

PR #668, opened 2026-09-22
fix(image): treat bytes sharp cannot decode as a missing image

### LINKED ISSUE #665: fix(image): image/* bodies sharp cannot decode are still reported (STAMP-71)
Sentry: [STAMP-71](https://snapshot-labs.sentry.io/issues/STAMP-71) — **105 occurrences**, first seen 2026-08-16 12:59 UTC, last seen 2026-09-22 08:00 UTC, `environment:production`, `handled:yes`, providers `space` and `ens`. Substatus `regressed`: #587 was closed by #598 on 2026-09-13 16:10 UTC, and events resumed on 2026-09-22 07:36 UTC — 2 so far, both `space/s:chinadefi.eth`. The 103 earlier events are the `text/html` class #598 fixed; the total is not this bug's size.

## What is reported

```
Error: Input buffer contains unsupported image format
```

```
src/api.ts                       Promise.all(currentResolvers.map(r => resolvers[r](address, network, networkId)))
src/resolvers/image/index.ts:56  withFailureContract → resolve(...args)
src/resolvers/image/index.ts:72  withResize → resize(input, max, max)
src/helpers/image.ts:56          resize → toWebp
src/helpers/image.ts:51          sharp(input).resize(w, h, options).webp().toBuffer()
```

Sample event, 2026-09-22 08:00:18 UTC — `GET https://cdn.stamp.fyi/space/s:chinadefi.eth`, `provider:space`, input `["chinadefi.eth", "1", "s"]`. Its HTTP breadcrumbs:

```
POST https://hub.snapshot.org/graphql                                                                      → 200
GET  https://nimg.ws.126.net/?url=https%3A%2F%2Fcrawl.ws.126.net%2F…jpeg&thumbnail=650x2147483647&quality=80&type=jpg → 302
GET  https://so.v.163.com/AEDm2GYACxMBAJqc.png                                                             → 200
```

## Root cause

The hub avatar of `chinadefi.eth` is a NetEase image-proxy URL. From production's egress the proxy answers 302 to a placeholder, and the placeholder is not a PNG (fetched 2026-09-22 08:05 UTC):

```
$ curl -sD - https://so.v.163.com/AEDm2GYACxMBAJqc.png
HTTP/1.1 200 OK
Content-Type: image/png
Content-Length: 194

data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAA…
```

194 bytes of ASCII — a `data:` URL saved as a file — served as `image/png`. sharp 0.34.5 on exactly those bytes throws the reported message. (From a residential IP and from Cloudflare the proxy serves the real 38 KB JPEG, which sharp decodes fine; the failure only shows from production, hence the breadcrumbs above.)

`readHttpImage` (`src/helpers/http.ts:46-50`, from #598) accepts any body whose `content-type` starts with `image/`, so these bytes reach `withResize` (`src/resolvers/image/index.ts:71-77`), where sharp throws and the catch runs `capture(err)` before `return false`. #598 checks the server's claim about the bytes; nothing checks the bytes. #587 named this gap as its optional second half ("covers a server that lies about content type"), and #598 shipped without it.

Same gap, reachable today though not yet seen since #598: a format sharp has no loader for. `beltdao.eth`'s avatar is `https://belt.fi/images/icon/favicon.ico` → 200 `image/vnd.microsoft.icon`; the production sharp build has no ImageMagick (`sharp.format.magick.input.buffer === false`), and sharp throws the identical message on it.

Not a regression: `withResize` has captured every sharp error since [95c4fe7](https://github.com/snapshot-labs/stamp/commit/95c4fe79560983cfcb625ccfd4f3892a327d98f1) (2026-08-16; the first event is from the same hour). sharp's format sniff is the only check that runs on the bytes themselves, and its verdict is currently reported instead of used.

**Nothing is broken for the user.** `withResize` returns `false` after capturing, the chain falls through, and `cdn.stamp.fyi/space/s:chinadefi.eth` serves the 64×64 blockie fallback with `SHORT_CACHE`. Only the Sentry report is wrong, so there is no service degradation to measure.

## Fix

Treat "sharp found no loader for these bytes" as the routine miss a 404 already is. The sharp message stays in the file that owns sharp:

`src/helpers/image.ts`
```ts
// What sharp says of bytes it finds no loader for: a mislabeled body, or a
// format libvips lacks (ICO, BMP). The host served no image — a miss, not a fault.
export function isUnsupportedImageError(err: unknown): boolean {
  return err instanceof Error && err.message === 'Input buffer contains unsupported image format';
}
```

`src/resolvers/image/index.ts`, `withResize`
```ts
    } catch (err) {
      if (isUnsupportedImageError(err)) return false;
      capture(err, { tags: { provider: name }, contexts: { input: { args } } });
      return false;
    }
```

Every resolver with `resize: true` goes through `withResize`, `starknet` included (its `fetchImageOrMetadata` now fetches via `fetchHttpImage`), so this is the single choke point. Other sharp errors — pixel limit, truncated JPEG, the SVG entity path — carry different messages and keep reporting.

## Acceptance

- A resolver answering `Buffer.from('data:image/png;base64,iVBORw0KGgo')` resolves `false` with no `capture`.
- Same for an ICO body.
- A decodable image still resizes; a sharp error with any other message is still captured.
- One unit case per branch, in `test/unit/resolvers/image/failureContract.test.ts`.

wa0x6e:
Sentry STAMP-71 reports `Input buffer contains unsupported image format` from the image resolvers, 105 times since August, most recently for `space/s:chinadefi.eth`. The avatar host answers `200 image/png` with a `data:` URL saved as text, so the content-type check passes and sharp, which finds no loader for the bytes, throws. The resolver already answers `false` and the blockie fallback serves, so the user sees nothing wrong; only the report is.

A host serving image/* bytes that decode as no image is the same routine miss a 404 is. That one sharp message is now swallowed at the resize wrapper every resized resolver goes through; every other sharp error (corrupt header, pixel limit, the SVG entity path) keeps reporting with the resolver attributed.

Unit cases cover a data: URL body, an ICO body, a decodable image, and a corrupt PNG that still reports.

Closes #665


tony8713 (approved):
Approving. I checked the boundary: `isUnsupportedImageError` matches sharp's exact "Input buffer contains unsupported image format" string, which is the no-loader case, and a corrupt PNG header still throws a different message and still reaches `capture` with the provider tag. Breaking either side (widening the match, or dropping the early return) turns the two new failureContract cases red while the four existing ones stay green, so the tests do discriminate.

One thing worth knowing rather than changing: the ICO case in the tests is one of a class, and BMP, JXL and a UTF-16 encoded SVG all hit the same no-loader message, so a host that genuinely serves one of those is now a silent blockie fallback instead of a report. That seems fine for a 105-count noise source, just noting the trade.

The red CI on this branch is the same four starknet image snapshot failures master has, nothing from this diff. comment-gate flags the one added comment at src/resolvers/image/index.ts:74; the guard is already pinned by the new tests, so it can go if you prefer the file bare.

PR #669, opened 2026-09-22
Update snapshot.js to 0.17.4

tony8713:
stamp was pinned to `@snapshot-labs/snapshot.js@^0.17.2`. This bumps it to `^0.17.4`, the latest published version.

The change touches the dependency range in `package.json` and the matching resolution in `yarn.lock`, nothing else.


wa0x6e (approved):
utAck

